PR #1, opened 2019-12-11
add a separate package for lsp

c-cube:


rgrinberg:
Could you please run `$ dune build @check` and commit the generated opam file?

rgrinberg:
I believe you need to make the lsp library public as well (by adding a public_name field)

PR #6, opened 2019-12-11
An attempt to fix e2e tests

andreypopp:
When dune runs rules it tries to copy deps into its own `_build` dir. It tries to copy `node_modules` dir as well.

Generally `node_modules` cannot be relocated and this is why it failed before this PR. Now instead we try to install `node_modules` within dune's build dir with a separate `e2e/node_modules` rule. This rule realizes `node_modules` dir and only depends on `package-lock.json` (which is I believe is the only correct way).

I'm not sure though it is optimal to run jest via dune:

- Not clear to me how it's going to well with jest's snapshot mechanism (a sort of expect tests which update test source files with recorded expectations), can we make it work with dune's promote mechanism?
- TS tooling wouldn't work if we realize `node_modules` inside dune's `_build` dir only — that would make editing test files harder. Though we can ask developers to do `npm install` manually if they want to edit test files... still not optimal from my point of view.

Would it be possible to instruct dune not to copy things into build dir when running some rules? That would address those points above.


rgrinberg:
> Would it be possible to instruct dune not to copy things into build dir when running some rules? That would address those points above.

It's possible to consider doing this, but the rules need to make sure they don't write anything to the source dir.

I suppose if there's a fundamental limitation in npm that prevents folder from being relocatable, then we should just drop this idea and run the tests from the source dir.

By the way, I plan on switching from npm to yarn. Does it also have this limitation?

andreypopp:
Ok, so the reason it was failed before not because of `node_modules` is not relocatable but because dune copies symlinks into build dir as files. This happens to `node_modules/.bin/jest` which then cannot find `../build/cli` (which was prev resolved against symnlink target).

> By the way, I plan on switching from npm to yarn. Does it also have this limitation?

Same.

I think if we cannot use dune as just a task runner for this case we better off using npm/yarn directly for test workflow.

rgrinberg:
> Ok, so the reason it was failed before not because of node_modules is not relocatable but because dune copies symlinks into build dir as files. This happens to node_modules/.bin/jest which then cannot find ../build/cli (which was prev resolved against symnlink target).
Sounds like a bug in dune.

> I think if we cannot use dune as just a task runner for this case we better off using npm/yarn directly for test workflow.

For now let’s just keep it simple and run it separately as you say. Would you mind updating this PR and adding a make test target?

rgrinberg:
Looks good. To find the executable for the server, what do you think about running the tests inside `$ dune exec`? This should make the lsp server binary available in PATH. I pushed a commit with what I have in mind.

rgrinberg:
Okay, this is good to go. For now let's just run the e2e test on 4.08 and 4.09

PR #10, opened 2019-12-12
Add travis ci

rgrinberg:
@avsm could you set up travis for this repo? I don't think I have the
credentials

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: c
+install: wget https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/.travis-docker.sh
+script: bash -ex .travis-docker.sh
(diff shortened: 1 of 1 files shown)


andreypopp:
Can we go with azure instead so we can test on Windows, macOS and Linux?

rgrinberg:
That makes sense. Do you think the extra testing from travis isn't useful? Or is it just redundant?

Do you want to setup the azure stuff? I assume you have some experience with it in esy already.

andreypopp:
Not redundant, at least for now while we don't have azure configured. Let's go with travis and we can add/replace it with azure later.

rgrinberg:
Alright, let's get some other CI integrations in. I'll make some tickets.

PR #13, opened 2019-12-13
Bump minimum version to 4.06

rgrinberg:
As discussed before, we can relax things retroactively as the need
arises. We will not be introducing any dependencies to lsp so it should
always be easy.


c-cube:
I still don't understand why we'd restrict artificially the versions on this package, which should ideally be as easy as possible to consume? It's pretty easy to support at least `>= 4.03` just for `lsp` (not the full `ocaml-lsp-server`). Just my 2¢.

PR #18, opened 2019-12-14
Setup OCamlFormat

### LINKED ISSUE #17: Setup OCamlFormat for this repository
Some code styles are a bit broken, so setting the OCamlFormat doesn't sound bad.

imbsky:
Closes #17


rgrinberg:
Good idea. Could you update the PR to use the following style (taken from dune)

```
version=0.12
break-separators=before
dock-collection-brackets=false
break-sequences=true
doc-comments=before
field-space=loose
let-and=sparse
sequence-style=terminator
type-decl=sparse
wrap-comments=true
if-then-else=k-r
let-and=sparse
space-around-records
space-around-lists
space-around-arrays
cases-exp-indent=2
break-cases=all
indicate-nested-or-patterns=unsafe-no
```

imbsky:
Updated!

PR #20, opened 2019-12-17
Directing both lsp and ocaml-lsp-server to same logger

mrkmndz:
Because lsp/src/ had its own copy of logger.ml, the log file was not being set by the logger call in ocaml-lsp-server.ml.  This meant that only the ocamlmerlin-lsp messages were ever actually being logged.

I fixed this by removing the copy of merlin/utils/logger.ml, and having both lsp and ocaml-lsp-server log to the original copy in merlin.

I'm not sure what your testing infra looks like now, but I was able to test this by setting MERLIN_LOG and then interacting with the lsp server, and now both classes of logs are coming through.


rgrinberg:
Thanks for your contribution.

> I fixed this by removing the copy of merlin/utils/logger.ml, and having both lsp and ocaml-lsp-server log to the original copy in merlin.

First, let's clarify the end goal. I think this following situation would be best:

* LSP server's logging is completely decoupled from merlin's.
* We minimize the dependencies in our LSP library.

To achieve these goals, I think we need our own logging implementation. Do you agree with these goals?

c-cube:
could we just use `logs` for that? It's pretty nice. It could also be vendored, I imagine.

rgrinberg:
We could, but I think that vendoring logs kind of defeats half its purpose. Now it would be impossible for the user to redirect library logs with one command.

So really, we'd be vendoring it just for logs interface. And while the interface is nice, we can easily recreate in a few dozen lines of code. I'm not sure what's the right way to go here yet...

c-cube:
A possibility for `lsp` would be to reproduce the API of logs (really, just the callback based thing, I also have many similar functions in personal projects) but with a global reference that changes where logs go (in effect, a reporter-like API):

```ocaml
(* in Lsp.Logger *)
val _logger  : (level -> (<continuation type for logs> -> unit) -> unit) ref
```

so that ocaml-lsp can just override this reference with its own logger backend.

rgrinberg:
What you suggest sounds reasonable to me. However I'm actually a bit skeptical of the need to log in LSP itself. Currently, we are logging requests and responses in the `RPC` module which seems to implement a "generic" lsp server. IMO, it would be better if the state machine for the protocol was factored out and did not include any logging. The logging should be pushed to the concrete lsp server and there we are free to use the vendored logs library.

mrkmndz:
IMO just having logging of the raw requests and responses has been really useful, which is why I ended up trying to figure out what was going wrong with this in the first place.  I don't have any strong opinions on whether to use merlin's logger, an external library, or a custom logger, I'd just like to ensure that we maintain the ability to enable that kind of low-level logging, since it really helps when debugging downstream clients of the lsp server.

rgrinberg:
@mrkmndz I agree. It's just that I'd like to move it to the LSP server itself. Adding merlin dependencies to the LSP library makes it less useful as a general purpose library.

rgrinberg:
As discussed above, depending on more merlin than is necessary isn't the right way to go forward.

PR #28, opened 2019-12-19
opam pin commands are needed

UnixJunkie:
so that instructions in the README actually work...

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ We recommend to install the server via a project such as
 To install the lsp server in a particular opam switch:
 
 ```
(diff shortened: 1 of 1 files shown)


imbsky:
I wonder if we should write this in the README because this is still a working in progress project.

imbsky:
People who want to use this at this point probably don't need this, and beginners just get confused.

UnixJunkie:
Advertising non working install instructions is just annoying. You can still update the README later if this becomes non necessary.

imbsky:
Well, I'll leave it to @rgrinberg. I understand your opinion.

rgrinberg:
Thanks

PR #30, opened 2019-12-19
Run e2e tests on CI

andreypopp:
Run e2e tests on CI


rgrinberg:
It seems like we aren't running the tests in the sandbox in this PR.

andreypopp:
Tests fail though now.

andreypopp:
Fixed completion tests for 408 and 409, disabled on 406 and 407.

The reason is that completion results differ for different ocaml versions.

PR #31, opened 2019-12-19
Bump dependencies version to latest

imbsky:


imbsky:
I updated the lock file and the packages version as well, because the lock file was outdated.

PR #33, opened 2019-12-21
More completion related fields in LSP

rgrinberg:
* Add triggerCharacter field to completionContext
* commitCharacters field to result

@andreypopp I see that we're using `[@default None]` for these optional fields.
What's the difference between this and `[@yojson.option]`?

### DIFF
--- a/lsp/src/protocol.ml
+++ b/lsp/src/protocol.ml
@@ -2543,7 +2543,10 @@ module Completion = struct
     }
   [@@yojson.allow_extra_fields]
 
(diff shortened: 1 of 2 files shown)


andreypopp:
`@optional` is expressed by omitting the key entirely during serialization while `@default None` will be serialized as null. I think this `@optional` is the correct way to specify fields marked with `name?: type` in LSP spec.

rgrinberg:
Okay, thanks for clarifying. I think we have quite a few places to change in the code then.

rgrinberg:
I've updated this PR accordingly.

PR #34, opened 2019-12-21
Add signature help lsp definitions

rgrinberg:
* SignatureHelpOptions
* ParameterInformation
* SignatureInformation

### DIFF
--- a/lsp/src/protocol.ml
+++ b/lsp/src/protocol.ml
@@ -3486,8 +3486,23 @@ module SignatureHelpOptions = struct
 end
 
 module ParameterInformation = struct
+  module Label = struct
+    type t =
+      | Substring of string
+      | Range of int * int
+
+    let t_of_yojson = function
+      | `String b -> Substring b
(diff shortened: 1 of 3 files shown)


rgrinberg:
@andreypopp this is now ready

lsp/src/protocol.ml
```diff
@@ -3391,6 +3406,482 @@ module Hover = struct
...
+      : t -> Ppx_yojson_conv_lib.Yojson.Safe.t )
+
+  let _ = yojson_of_t
+
+  [@@@end]
+end
+
+module ParameterInformation = struct
+  type t =
+    { label : (int * int) Or_string.t
```
andreypopp: nit: I'd have another datatype here for label, protocol says:

> Either a string or an inclusive start and exclusive end offsets within its containing signature label.

I think having something like:
```
type paramInfoLabel = 
  | BySubstring of string
  | ByRange of int * int` would
```
would be self-documenting, otherwise I'd expect code which uses those types to have comments describing what means what.

PR #38, opened 2019-12-22
Logging improvements

pinotree:
This patch series:
- improves the way command line arguments are handled (using `Arg`)
- add an option to specify the file (better than the environment variable)
- make sure the logging output is flushed

### DIFF
--- a/lsp/src/logger.ml
+++ b/lsp/src/logger.ml
@@ -59,7 +59,8 @@ let log ~section ~title fmt =
         if str <> "" then (
           output_string oc str;
           if str.[String.length str - 1] <> '\n' then output_char oc '\n'
-        ))
+        );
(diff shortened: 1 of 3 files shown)


rgrinberg:
Looks good. Let's use this opportunity not to reuse the merlin environment variable by the way. Can we rename this to `OCAML_LSP_SERVER_LOG` instead?

cc @andreypopp

pinotree:
> Can we rename this to OCAML_LSP_SERVER_LOG instead?

For me it's the same -- do I add a new commit to this PR, or open a new PR once this is merged?

andreypopp:
Yes, OCAML_LSP_SERVER_LOG sounds good to me. Just another commit to the PR would be ok.

PR #42, opened 2019-12-24
Remove dependency on result package

andreypopp:
As we depend on OCaml >= 4.06 we don't need this package.

PR #43, opened 2019-12-24
Allow to use ocaml-lsp-server with opam pin / esy link workflows

andreypopp:
This is a workaround for ocaml/dune#2982 which allows to use ocaml-lsp with opam
pin and esy link workflows.

Summary:

- Remove a dep on lsp package as we treat lsp as a part of the vendored
  set of deps.

- Fix build command first to consider the entire workspace while
  building ocaml-lsp-server package. This is needed so we see vendored
  merlin and other packages.

This also removes `with-test` and `with-doc` which can be restored later once we
address that in dune.


andreypopp:
This should land after #42.

dune-project
```diff
@@ -15,30 +15,4 @@
 
 (source (github ocaml/ocaml-lsp))
 
-(generate_opam_files true)
```
andreypopp: Had to remove generation of opam files for now to be able to provide custom build commands.

ocaml-lsp-server.opam
```diff
@@ -14,10 +13,9 @@ license: "ISC"
 homepage: "https://github.com/ocaml/ocaml-lsp"
 bug-reports: "https://github.com/ocaml/ocaml-lsp/issues"
 depends: [
-  "result"
-  "lsp"
   "yojson"
   "ppx_yojson_conv_lib"
+  "ocamlfind" {>= "1.5.2"}
```
andreypopp: This is needed as merlin depends on `findlib` lib.

ocaml-lsp-server.opam
```diff
@@ -26,13 +24,20 @@ build: [
...
+    "--ignore-promoted-rules"
+    "--no-config"
+    "--profile"
+    "release"
+    "%{name}%.install"
+  ]
+  [
+    "cp"
+    "%{_:build}%/_build/default/%{name}%.install"
+    "%{name}%.install"
```
andreypopp: This is a sad hack to make it work both with dune 1.0 an dune 2.0 (in dune 2.0 `--ignore-promoted-rules` doesn't promote install files but `--promote-install-files` is only available in dune 2.0).

I think it's worth it.

ocaml-lsp-server.opam
```diff
@@ -26,13 +24,20 @@ build: [
...
-    "@runtest" {with-test}
-    "@doc" {with-doc}
+    "--root"
+    "."
+    "--ignore-promoted-rules"
+    "--no-config"
+    "--profile"
+    "release"
+    "%{name}%.install"
+  ]
```
andreypopp: My intention with this is that this invocation is the same as `-p %{name}%` but allows to see other packages in the workspace.

ocaml-lsp-server.opam
rgrinberg: What is `%{_:build}%`?

dune-project
rgrinberg: There's no need to remove it. It's possible to override the build step only in the generated .opam file. See how we do it in dune for example https://github.com/ocaml/dune/blob/master/dune.opam.template

ocaml-lsp-server.opam
andreypopp: Build dir of the package opam is building currently.

dune-project
andreypopp: Whoa, didn't know dune supports it. Fixed.

PR #46, opened 2019-12-25
Rename ocaml-lsp-server to ocamllsp

andreypopp:
To be consistent with other ocaml tools.


andreypopp:
Should we rename the package with `ocamllsp` executable to `ocamllsp` as well?

PR #48, opened 2019-12-26
Add destruct code action

rgrinberg:
- [x] Add destruct code action
- [x] Add tests

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -10,7 +10,7 @@ end
 
 let initializeInfo : Lsp.Protocol.Initialize.result =
   let codeActionProvider : Lsp.Protocol.CodeActionOptions.t =
-    { codeActionsKinds = [ Other Action.destruct] }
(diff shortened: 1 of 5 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -102,6 +105,21 @@ let send_diagnostics rpc doc =
...
 let on_initialize _rpc state _params = Ok (state, initializeInfo)
 
+let code_action_of_case_analysis uri (loc, newText) =
+  let edit : Lsp.Protocol.WorkspaceEdit.t =
+    let textedit : Lsp.Protocol.TextEdit.t =
+      { range = range_of_loc loc; newText }
+    in
+    { changes = Some [ (uri, [ textedit ]) ]; documentChanges = None }
+  in
+  let title = "destruct" in
```
andreypopp: nit: not sure this is the best name for this action, at least we could capitalize it

rgrinberg:
Okay, I think this is ready. One thing that's a bit of a shame is that we can't insert a snippet using a code action (at least I couldn't see how after reading the specification). In any case, this should be useful.

PR #52, opened 2019-12-26
Do not terminate server after exceptions

rgrinberg:
Return an error instead.


rgrinberg:
I'm merging this because the server is hardly usable without this change.

PR #53, opened 2019-12-27
CompletionItemResolve support

rgrinberg:
Currently this does nothing and just returns the same item


rgrinberg:
> The intention is to move fetching doc into resolve phase, right?

Correct. But this requires a bit more support on the merlin side and we're still discussing the details here. See my question on slack.

PR #54, opened 2019-12-30
Remove definition location

rgrinberg:
This type is not present in the spec


rgrinberg:
I've also tweaked definition to use the correction location type. I think we should stick to the protocol as much as possible in our library.

PR #55, opened 2019-12-30
More faithful protocol implementation

rgrinberg:
* Remove nuclide related fields
* Reflect optional fields more closely

### DIFF
--- a/ocaml-lsp-server/test/e2e/__tests__/textDocument-completion.test.ts
+++ b/ocaml-lsp-server/test/e2e/__tests__/textDocument-completion.test.ts
@@ -129,17 +129,17 @@ describe("textDocument/completion", () => {
       let somestring = "hello"
 
       let plus_42 (x:int) (y:int) =
-        somenum + 
+        somenum +
     `);
 
     let items = await queryCompletion(Types.Position.create(4, 12));
     let items_top5 = items.slice(0, 5);
     expect(items_top5).toMatchObject([
-      { label: "somenum", textEdit: null },
-      { label: "x", textEdit: null },
-      { label: "y", textEdit: null },
-      { label: "max_int", textEdit: null },
-      { label: "min_int", textEdit: null },
+      { label: "somenum" },
+      { label: "x" },
+      { label: "y" },
+      { label: "max_int" },
+      { label: "min_int" },
     ]);
   });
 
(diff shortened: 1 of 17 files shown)


rgrinberg:
@andreypopp on second thought, I'm wondering if if we should automatically coerce nulls to empty lists. As far as I understand, this distinction shouldn't matter although it's not specified explicitly anywhere.

ocaml-lsp-server/test/e2e/__tests__/textDocument-completion.test.ts
```diff
@@ -285,17 +285,17 @@ describe("textDocument/completion", () => {
 
   it("completes labels", async () => {
     openDocument(outdent`
-      let f = ListLabels.map 
+      let f = ListLabels.map
```
andreypopp: Tests now fail because of this

rgrinberg:
> Yeah, shouldn't matter in most cases but there might be some where "absence" is different than "empty". Don't know if there are any in LSP though .

@c-cube what do you think? will this make a difference for you?

I would like it be possible to implement generic LSP servers with this library, so  hopefully this will not matter in practice

rgrinberg:
I'll disable the completion tests for now. Let's write them in a portable manner first, and then we can enable them.

ocaml-lsp-server/test/e2e/__tests__/textDocument-completion.test.ts
rgrinberg: Currently, the completion tests aren't in a good shape. Let's fix them first before wasting too many cycles on this.

rgrinberg:
I coerced all nulls to empty lists

rgrinberg:
And I've found an instance where `[]` is not the same as absence. When handling `only` for code actions, `[]` means to present no actions while absence of the field means to present all of them.

c-cube:
> And I've found an instance where `[]` is not the same as absence. `only` for code actions:
> 
>     * `[]` means to present no actions
>     * absence of the field means to present all of them.

I think it's reasonable to have a custom sum type there (basically a list option with `All` or `Only of _ list`) and still adopt the convention that `null` is empty list otherwise. I don't think it'd be problematic for me, after all the exact serialization doesn't matter to users, only the semantics of messages.

Issue #57, opened 2020-01-01
function disappeared from API


c-cube:
`empty_server_capabilities`, introduced in #40, disappeared in 86aefc096e6d546ecdd044bdaa68de0bf08a0016 . Probably a merge gone wrong :)

rgrinberg:
It should be back now. However, don’t expect these empty capabilities to stick around. We’re more likely to rename them to defaults.

PR #58, opened 2020-01-01
Clean up more types and bring back empties

rgrinberg:
I'm looking into fixing #57 but encountering some weirdness in our API.

First of all, it seems like we are defaulting missing capabilities to absence. Is that ok?

Our default for hover looks even weirder:

```ocaml
    let empty = { contentFormat = [ Plaintext ] }
```

Should we a bit more conservative here perhaps?


rgrinberg:
@c-cube I've added an Only type as you asked.

rgrinberg:
> What would be more conservative than plaintext? Again, the spec doesn't say but I'd think that the absence of contentFormat would mean that hover returns plaintext.

Well, the list of supported content types depends on the declared client capabilities. I agree that in practice everyone supports plain text, but a valid client could explicitly reject it.

Also an "empty" capability implies no support for anything in the other cases, while here we are breaking with our convention. It seems like we can pick defaults consistently or sensibly, but we cannot do both.

rgrinberg:
I've added complete support for the various editing types.

@c-cube @andreypopp what are your thoughts on naming the modules/types? Currently, I'm trying to stick to names that make it easy to cross reference the LSP specification. But I can imagine a better way to organize some things. E.g:

```ocaml
module Completion : sig
  module Item : sig
    type t
  end
  module Kind : sig
    type t
  end

  type t = Item.t list
end
```

andreypopp:
> Well, the list of supported content types depends on the declared client capabilities. I agree that in practice everyone supports plain text, but a valid client could explicitly reject it.

It seems unlikely that the absence of `contentFormat` implies no support for displaying anything — it makes no sense to me that client declares support for hover but cannot display its result. More likely that `contentFormat` wasn't present in the previous editions of the spec and then it was added as an optional field for backward compat reasons.

> Also an "empty" capability implies no support for anything in the other cases

If by "empty" you mean 
```
hover: {}
```
then I don't think it implies no support. There are capabilities which have only `dynamicRegistration` key allowed — how would you specify support w/o dynamic registration then? My thinking is taht you do that by specifying an "empty" capability `{}`.

andreypopp:
>  Currently, I'm trying to stick to names that make it easy to cross reference the LSP specification.

This sounds useful to me.

rgrinberg:
I've added support for workspace folders as well

```
/**
	 * The workspace folders configured in the client when the server starts.
	 * This property is only available if the client supports workspace folders.
	 * It can be `null` if the client supports workspace folders but none are
	 * configured.
	 *
	 * Since 3.6.0
	 */
```

rgrinberg:
> then I don't think it implies no support. There are capabilities which have only dynamicRegistration key allowed — how would you specify support w/o dynamic registration then? My thinking is taht you do that by specifying an "empty" capability {}.

It seems like `default` is the better name then. `empty` implies certain absence (and sometimes even some laws), while in our case it doesn't really.

andreypopp:
Agree, default is a better naming.

PR #60, opened 2020-01-03
Add interface for OO server

rgrinberg:
@c-cube @andreypopp, as discussed before, this the idea that I have in mind.

With this approach, we make it easy for the user to write complete servers with
minimal boilerplate:

* We add sane default implementation that verify client capabilities
* We give some common exception types and a way to convert exceptions to valid
  jsonrpc errors
* We reserve the ability to add new server features without the user changing
  their code.
* Exhaustive pattern matching on the requests leads to fragile code and we
  should not encourage it
* We allow the user to add custom logging by adding a general function to
  transform requests/responses (more on this later)

Of course, this also doesn't prevent us from supporting Lwt later either. We
simply need to wrap the class type in a functor.


c-cube:
I'd say it's better to have only one exception with a sum type as argument (+ a printer for the sum type). Otherwise, I do like the idea a lot, it's a textbook case for OO.

Rather than having a class type, shouldn't we have a class that directly implements the whole thing, but with no-op for each message? Or is that separate?

rgrinberg:
> I'd say it's better to have only one exception with a sum type as argument (+ a printer for the sum type). Otherwise, I do like the idea a lot, it's a textbook case for OO.
How would a user define his own errors? Would this sum type be open?
> Rather than having a class type, shouldn't we have a class that directly implements the whole thing, but with no-op for each message? Or is that separate?
Yes that’s the idea. I just wanted to write the signature down first. Rather than no-op we’ll have a standard exception for unimplemented stuff

c-cube:
I was more thinking of a closed sum type, maybe with a `… | Custom of {msg: string}` or something like that, so that jsonrpc error codes are always well defined. It's hard to catch open-ended exceptions :)

rgrinberg:
> I was more thinking of a closed sum type, maybe with a … | Custom of {msg: string} or something like that, so that jsonrpc error codes are always well defined. It's hard to catch open-ended exceptions :)

Ah yes, that would be good. However, I think that it would also be good for the user to automatically register converters for his own custom error types.

andreypopp:
We've already discussed using OO offline with @rgrinberg and I've mentioned that while I'm not strongly against OO I can't say I like it.

As @rgrinberg explained to me — the main advantage of using OO is late binding — but it's not clear to me why this is going to be useful when implementing an LSP server. There's a single example of usefulness of late binding in #64 (client.ml, showMessage can use the implementation of logMessage which could be overridden) but that's just a single example which is not (I think) representative.

Probably the main reason is that I'm too inexperienced OCaml programmer (which is true) — I just haven't seen the need to use OO in OCaml yet. But that could be a valid reason not to use OO here though! There could be more people like me who don't have enough experience with OCaml's OO — that could be a barrier (albeit small as I'm sure/hope it's not hard to grok) for contribution/adoption. Not sure if you see this as a valid point but I thought I'll mention it at the very least.

Finally, I think the approach with encoding LSP handler as a record would work well too while being familiar to all OCaml programmers.

So if we look here:
https://github.com/ocaml/ocaml-lsp/blob/821edf2b9dc56157b64b00d0f3215de77295aa96/lsp/src/rpc.mli

We have a `handler` type defined which exposes `on_initialize`, `on_request` and `on_notification` fields. We can extend it further to accommodate all LSP methods (there are not that many) and then make the `handler` type abstract, exposing only constructor with optional arguments (useful for providing a default implementation):
```
type ctx
type server

val start :
  ?on_initialize:(ctx -> Initialize.Params.t -> Initialize.Response.t)
  -> ?on_hover:(ctx -> Hover.Params.t -> Hover.Response.t)
  ...
  -> server

val notify : ctx -> Server_notification.t -> unit

val stop : server -> unit
```

As you can see it solves the problem of implementing only a subset of LSP methods. Also, if I were presenting an example of `client.ml` implemented this way, it would allow implementing `showMessage` in terms of `logMessage` and have `logMessage` being overridable.

I know, I know, this is essentially like re-implementing OO via explicitly passing `ctx` but for me this approach seems more familiar and thus clear.

Again, I'm not against OO, just my 2 cents.

rgrinberg:
Do you think that the object oriented code poses such a block for OCaml programmers? Most programmers are already familiar with OO from other languages and it really isn't all that uncommon in OCaml libraries. Just off the top of my head, I can think of encountering it in Lwt, Cryptokit, ocamlnet, webmachine, lambda-term, and various ppx libs. Avoiding features like GADTs and first class modules makes just as much sense if we want to keep the API simple. From my experience, OCaml programmers struggle with them just as much as, if not more.

I can give you a couple of other examples where the OO style pays off:

* Adding a mixin class that will keep track of workspace folders
* Add a mixin that will keep track of client configuration changes
* A mixin that keeps track of the states of opened documents.

With that being said, how about the following compromise:

We add a simple records based API like with your `start` function and leave out the manual self passing. For those who want more control, the OO api will be available.

andreypopp:
> Do you think that the object oriented code poses such a block for OCaml programmers?

Definitely not a **block**.

> I can give you a couple of other examples where the OO style pays off:

Those are good examples.

> With that being said, how about the following compromise:

Let's go with OO only api, let's not complicate API with multiple choices.

rgrinberg:
Note that this branch needs the code generation stuff to be merged before it’s revived. We will be able to generate the class definition as well

giltho:
I think that the OO interface is quite nice and will be familiar to people who wrote LSP servers using the standard VSCode API in TS.
I had a question about the signature of notifications like `shutdown` that has type `unit` while I would expect `unit -> unit`, but I guess code generation will settle that anyway.

PR #63, opened 2020-01-05
Add jsonrpc to requests

### LINKED ISSUE #62: Missing RPC version field (jsonrpc) causes lsp-mode to crash
lsp-mode fails if `jsonrpc` field is missing

https://github.com/emacs-lsp/lsp-mode/blob/44bb6a31a892887f3d05a3333e49d816daf92a2f/lsp-mode.el#L4992

I'm on commit: `e5e6ebf9dcf157`

rgrinberg:
Fix #62


rgrinberg:
Merging as it's a trivial fix that's affecting users.

PR #64, opened 2020-01-06
Refactor before OO Interface

rgrinberg:
This refactor adds more complete support for LSP and separates:

* Client notifications
* Server notifications
* Client requests
* Server requests

I've added some hierarchical names this time. I'm still not sure which approach is better.

After this PR the OO server will come.


rgrinberg:
For fun, I've added a pseudo LSP client in OO style. @c-cube could you have a quick look over `Client`? I'm doing some awkward things to make sure that `request` & `notify` can be overriden.

lsp/src/header.ml
```diff
@@ -0,0 +1,91 @@
+open Import
+
+(* TODO rewrite all this using faraday/angstrom *)
```
c-cube: you can take code from https://github.com/c-cube/jsonrpc2/blob/master/src/jsonrpc2_core.ml#L515 if you want. Parsing these  is pretty simple.

c-cube:
@rgrinberg right now `request` and `notify` are not implemented, right? What kind of overriding were you thinking about, I'm not seeing anything related in `Client`.

lsp/src/header.ml
rgrinberg: Would you mind sending a PR?

rgrinberg:
> right now request and notify are not implemented, right?

Yes they aren't implemented

> What kind of overriding were you thinking about, I'm not seeing anything related in Client.

Oh, I guess removed it already. Suppose I want to call `request` or `notify` from within one of my default implementations. To achieve this, I need a bit of boilerplate to pass `t` to the created object itself.

c-cube:
> Oh, I guess removed it already. Suppose I want to call `request` or `notify` from within one of my default implementations. To achieve this, I need a bit of boilerplate to pass `t` to the created object itself.

could you just provide methods for that? so you'd `self#notify` or `self#request`.

rgrinberg:
> could you just provide methods for that? so you'd self#notify or self#request.

I would prefer if it was not possible to override these methods.

c-cube:
A bit of a hack, but you might be able to use `val` for the functions.

```ocaml
class foo = object(self)
  val f = (fun x -> x+1)
  method g = f 40
end;;
class bar = object(self)
  inherit foo
  method g = f 50
end;;
let b = new bar;;
b#g;;
```

rgrinberg:
I got rid of the experimental client API for now. The rest of the refactoring is straight forward I believe.

PR #67, opened 2020-01-24
Fuller support for LSP

rgrinberg:
This is some odd work I had accumulated a couple of weeks ago. Mostly it's just
more implementation of the protocol.


rgrinberg:
@andreypopp Following this PR, I would like to introduce stdune as the standard library for LSP. Do you have any objections?

PR #68, opened 2020-01-27
Import stdune

rgrinberg:
This is much better than our home made stdlib


rgrinberg:
For now, I've dropped support for 4.06. I'll restore it once we're ready for the 1.0 release.

PR #81, opened 2020-02-03
Remove dummy command from code lens

rgrinberg:
@andreypopp any reason why it was attached in the first place?

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -277,17 +277,12 @@ let on_request :
         in
         match item.Query_protocol.outline_type with
         | None -> children
-        | Some typ ->
+        | Some _ ->
           let loc = item.Query_protocol.location in
           let info =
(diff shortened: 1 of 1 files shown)


rgrinberg:
Actually, I've looked into the documentation of code lenses and it seems like we should really just not output a code lens at all if we don't have useful commands to attach to the elements.

andreypopp:
The way codelens are implemented right now is they provide info about types for top level bindings. I felt like it's a nice gimmick and probably can help beginners to learn about how type inference is working





> Actually, I've looked into the documentation of code lenses and it seems like we should really just not output a code lens at all if we don't have useful commands to attach to the elements.

I'm not convinced that there should be a command attached for code lens to be useful. I see codelens as a way to overlay code with some important info — types, show git blame info and so on.

That said I'm not attached to it (don't even use it now as my new lsp vim client doesn't have that feature yet), if you feel it's useless — let's remove, less code to maintain.

rgrinberg:
Okay, I'm closing this PR. I suppose it's more of an issue of code lenses in LSP that they don't allow you to put pure display information.

PR #87, opened 2020-02-11
Generate LSP sources from Spec

rgrinberg:
This started off as an experiment, but it has gone far enough that I think we
should consider finishing this work off and including it.

The spec is constantly evolving, and it's hard to keep up with it. Additionally,
the naming convention is completely inconsistent and it's too much of a pain to
maintain it manually.

In this PR, I parse the spec directly from the markdown and (will) use it to:

- [x] Generate type definitions
- [x] Generate constructor functions with labelled/optional arguments
- [ ] Possibly: generate JS encoders/decoders. This will be a lot more work, so
      I'm not so sure about it.
- [ ] Extract rpc requests/response types. Currently, I just read the json
      types.
- [ ] Attach locations/`@since` annotations to the AST.

The parser was quite easy to write with @bobot's help :)


giltho:
This is absolutely fantastic ! It would take a huge load off our shoulders.

giltho:
Coming in a tiny bit late with the idea, but have you thought about using tree-sitter ?
[Bindings](https://github.com/onivim/reason-tree-sitter) were written for oni2, and there is a very up-to-date [syntax for TypeScript](https://github.com/tree-sitter/tree-sitter-typescript).
It's a bit heavy to install tree-sitter, but it is only a dev-dependency so it will never be installed on the target user machine.

I've never used tree-sitter, so I don't know how easy that would be to use it, but it seems like one less thing to maintain.

rgrinberg:
It’s plausible but I seriously doubt I would write it faster with tree sitter. The parser itself was written in one day (with help from an experienced dev). The challenges with this PR stem from other parts of the code generation process. The parsing is already done and I’m working on the other(more difficult) parts now.

giltho:
Ah cool, didn't see the parser was already done. Perfect then :)

rgrinberg:
Some progress to report:

* I'm now able to generate all enums and interfaces
* For now, we'll still rely on ppx_deriving_yojson for conversion. It will be too hard to implement our own json serialization/deserialization in a single PR.
* `[@key ..]` and `[yojson.option ..]` annotations are available.
* Generation of primitive types is complete.

Next steps:

* Generate modules in topological order. Using recursive modules is quite suboptimal from a readability point of view.
* Improve the pipeline of code generation -> deriving inline -> ocamlformat. It's getting quite hairy to run things manually.
* Support for sum types. These are somewhat harder to support as there's no clear recipe.

@Gbury at this point, we can re-think how we handle default values. Perhaps it makes more sense not to handle them at all at this level and let implementors of servers to choose their own implementations?

Gbury:
I agree that having options might be simpler than default values: seems easier to generate, and server implementations can then choose what to consider as their default/fallback options.

rgrinberg:
Progress updates: (cc @giltho @Gbury)

The generation of types is almost complete. I just to finish the generating type definitions to match typescript's untagged unions. My convention is as follows:

* Any `number | string` will be treated as a `Jsonrpc.t`

* Anything with a `| null` will be an `option`.

* Avoid introducing any names that aren't present in the spec. This can be done by generating polymorphic variants. We'll choose constructor names according to the following scheme:

If it's a primitive type (string, bool, number), we'll use the standard constructor from yojson `String, Assoc, Bool`, etc.

If it's type defined elsewhere, we'll use the name of the type. E.g. `Location | ..` will be ```[`Location of Location.t```

* `ppx_deriving_yojson` makes it quite awkward to parse things such as `kind` fields on records. E.g:

```
/**
 * Create file operation
 */
export interface CreateFile {
	/**
	 * A create
	 */
	kind: 'create';
	/**
	 * The resource to create.
	 */
	uri: DocumentUri;
	/**
	 * Additional options
	 */
	options?: CreateFileOptions;
}
```

We'll need to generate some silly looking code that will inject this extra field on encoding and validate its existence on decoding.

In the future, we should just get rid of ppx_deriving completely and generate our own conversion. I'm getting a little sick of writing more code generators at this point.

giltho:
This is fantastic by the way!
But I can understand you're getting sick of code generation 😅 
I do agree with dropping ppx_deriving in the long term, but this is already pretty good for now

rgrinberg:
And now all the type signatures are generated! I've switched to cinaps from a plain code generator so that we can simply write the type definitions for cases when it's easier (See `MarkedString`).

giltho:
This is great. I didn't real through the 2700 of the gprotocol.ml. I read the beginning, looked at the main types and then read random types in the file. I couldn't see any problem with the generated types.

One "potential improvement" would be to replace every `bool Json.Nullable_option.t` with `bool` and `null` defaulting to `false`, but that doesn't change the correctness of what is generated.
The other thing is the yojson serialization and deserialization for things like `kind: 'create'`, but you already pointed it out earlier

giltho:
I'm thinking, you could make a library out of it. This could potentially be an amazing tool for writing bindings to TS libraries in general, not just in LSP.

rgrinberg:
> One "potential improvement" would be to replace every bool Json.Nullable_option.t with bool and null defaulting to false, but that doesn't change the correctness of what is generated.
I’m not entirely sure this simplification is correct however. In some cases, the absence might not necessarily imply something falsey. For example, what if LSP introduces a mechanism to disable some existing setting? The correct default would be true.
> The other thing is the yojson serialization and deserialization for things like kind: 'create', but you already pointed it out earlier
Yup, I will try to do this this week. It’s the main thing that’s blocking this PR from being merged.

> I'm thinking, you could make a library out of it. This could potentially be an amazing tool for writing bindings to TS libraries in general, not just in LSP.

That’s an interesting thought. For now, this is all a bit experimental so let’s see how it evolves in our repository. If there’s more demand for this, I’m happy to make this available on its own. Although I have to say that I find specialized code generators far more useful than general ones. Nevertheless, there are definitely some components that could be useful to people.

giltho:
> For example, what if LSP introduces a mechanism to disable some existing setting? The correct default would be true.

My intuition is that optional booleans are false by default and will always be because in JS/TS you can just say if(opt) and wether it's false or null the test will return the same thing. Although you're right there is no guarantee that it will stay true.

That being said, this looks great and it looks quite close to being mergeable :)

rgrinberg:
Okay, the generated code was good enough to replace some the initialization process. I will generate some `create` functions to make it easier to replace existing code.

rgrinberg:
I've replace initialization and code actions with the generated API and things seem to be working just fine.

rgrinberg:
Okay, I've ported everything to use the generated types. Right now, updating the types is a bit of a pain due to a bug in the dune rules for cinaps. I can guide people through the process if anyone is curious.

For now, the PR is good enough to be merged.

PR #89, opened 2020-02-12
Implement textDocument/formatting

cannorin:
This PR implements `textDocument/formatting` with a wrapper module to `ocamlformat`.

It finds `ocamlformat` from the current PATH and runs it at the current directory `ocamllsp` is running. Note that I didn't add any facility to find the nearest `.ocamlformat`.

`textDocument/{rangeFormatting, onTypeFormatting}` is not yet implemented. Once https://github.com/ocaml-ppx/ocamlformat/pull/1188 comes into reality, we can easily extend this implementation to support them.

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -649,7 +649,7 @@ let on_request :
     let fileName = Document.uri doc |> Lsp.Uri.to_path in
     let result =
       Ocamlformat.format_file
-        (Ocamlformat.Input.Stdin (src, Ocamlformat.FileType.Name fileName))
+        (Ocamlformat.Input.Stdin (src, Ocamlformat.File_type.Name fileName))
         Ocamlformat.Output.Stdout Ocamlformat.Options.default
     in
     match result with
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -731,7 +731,6 @@ let start () =
   let on_request rpc state caps req =
     prepare_and_run @@ fun () -> on_request rpc state caps req
   in
-  log ~title:"info" "starting";
   Lsp.Rpc.start docs { on_initialize; on_request; on_notification } stdin stdout;
   log ~title:"info" "exiting"
 
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -644,12 +644,15 @@ let on_request :
   | Lsp.Client_request.CompletionItemResolve compl -> Ok (store, compl)
   | Lsp.Client_request.TextDocumentFormatting
       { textDocument = { uri }; options = _ } ->
+    Printf.eprintf "start format\n";
     Document_store.get store uri >>= fun doc ->
     let src = Document.source doc |> Msource.text in
+    let fileName = Document.uri doc |> Lsp.Uri.to_path in
     let edits =
-      begin match Ocamlformat.exec src [] with
+      begin match Ocamlformat.exec ~fileName ~content:src [] with
       | Result.Error e -> begin
-        Lsp.Logger.log ~title:"error" ~section:"textDocument_formatting" "%s" e;
+        Printf.eprintf "format error: \n%s\n" e;
+        log ~title:"error" "%s" e;
         []
         end
       | Result.Ok result -> begin
@@ -736,6 +739,7 @@ let start () =
   let on_request rpc state caps req =
     prepare_and_run @@ fun () -> on_request rpc state caps req
   in
+  log ~title:"info" "starting";
   Lsp.Rpc.start docs { on_initialize; on_request; on_notification } stdin stdout;
   log ~title:"info" "exiting"
 
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -643,36 +643,28 @@ let on_request :
   | Lsp.Client_request.CodeAction params -> code_action store params
   | Lsp.Client_request.CompletionItemResolve compl -> Ok (store, compl)
   | Lsp.Client_request.TextDocumentFormatting
-      { textDocument = { uri }; options = _ } ->
-    Printf.eprintf "start format\n";
+      { textDocument = { uri }; options = _ } -> (
     Document_store.get store uri >>= fun doc ->
     let src = Document.source doc |> Msource.text in
     let fileName = Document.uri doc |> Lsp.Uri.to_path in
-    let edits =
-      begin match Ocamlformat.exec ~fileName ~content:src [] with
-      | Result.Error e -> begin
-        Printf.eprintf "format error: \n%s\n" e;
-        log ~title:"error" "%s" e;
-        []
-        end
-      | Result.Ok result -> begin
-        let pos line col =
-          { Lsp.Protocol.Position.character = col; line }
-        in
-        let range =
-          let startPos = pos 0 0 in
-          begin match Msource.get_logical (Document.source doc) `End with
-          | `Logical (l, c) ->
-            let endPos = pos l c in
-            { Lsp.Protocol.Range.start_ = startPos; end_ = endPos }
-          end
-        in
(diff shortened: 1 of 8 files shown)


ocaml-lsp-server/src/ocamlformat.ml
```diff
@@ -0,0 +1,239 @@
+open Import
+
+module FileType = struct
```
rgrinberg: The convention in OCaml is to use underscores. It should be `File_type` in this case.

The reason why some identifiers are in fact in `PascalCase` is because they directly correspond to identifiers in the LSP spec.

ocaml-lsp-server/src/ocamlformat.ml
```diff
@@ -0,0 +1,239 @@
...
+      | Ocamlformat
+      | Janestreet
+
+    let to_string = function
+      | Conventional -> "conventional"
+      | Default -> "default"
+      | Compact -> "compact"
+      | Sparse -> "sparse"
+      | Ocamlformat -> "ocamlformat"
+      | Janestreet -> "janestreet"
```
rgrinberg: I don't think we should include all these options to be honest. In ocamlformat, the plan is to eliminate them anyway.

ocaml-lsp-server/src/ocamlformat.ml
cannorin: I didn't know about the convention, thanks! I'll fix it tomorrow.

cannorin:
OK, I'll remove those config stuffs.

cannorin:
We will have to set `--lines` option to support `textDocument/{rangeFormatting, onTypeFormatting}`, so I'll leave minimal support of passing options instead of completely removing it.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -640,7 +640,29 @@ let on_request :
...
+      Ocamlformat.format_file
+        (Ocamlformat.Input.Stdin (src, Ocamlformat.File_type.Name fileName))
+        Ocamlformat.Output.Stdout Ocamlformat.Options.default
+    in
+    match result with
+    | Result.Error e -> Error (Printf.sprintf "format failed: %s" e)
+    | Result.Ok result ->
+      let pos line col = { Lsp.Protocol.Position.character = col; line } in
+      let range =
+        let startPos = pos 0 0 in
```
rgrinberg: naming convention is wrong.

ocaml-lsp-server/src/ocamlformat.ml
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+let run_command command ?stdin_value args : command_result =
+  let command =
+    match args with
+    | [] -> command
+    | _ -> Printf.sprintf "%s %s" command (String.concat ~sep:" " args)
+  in
+  let env = Unix.environment () in
+  let inChan, outChan, errChan = Unix.open_process_full command env in
```
rgrinberg: `in_chan`, etc.

ocaml-lsp-server/src/ocamlformat.mli
```diff
@@ -0,0 +1,40 @@
...
+
+module Options : sig
+  type t = string list
+
+  val default : t
+end
+
+type 'result command =
+  | FormatFile : Input.t * 'result Output.t -> 'result command
+  | FormatFilesInPlace : string list -> unit command
```
rgrinberg: `Format_files_in_place`.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -640,7 +640,29 @@ let on_request :
...
   | Lsp.Client_request.CompletionItemResolve compl -> Ok (store, compl)
-  | Lsp.Client_request.TextDocumentFormatting _ -> Ok (store, [])
+  | Lsp.Client_request.TextDocumentFormatting
+      { textDocument = { uri }; options = _ } -> (
+    Document_store.get store uri >>= fun doc ->
+    let src = Document.source doc |> Msource.text in
+    let fileName = Document.uri doc |> Lsp.Uri.to_path in
+    let result =
+      Ocamlformat.format_file
+        (Ocamlformat.Input.Stdin (src, Ocamlformat.File_type.Name fileName))
```
rgrinberg: `file_name`.

ocaml-lsp-server/src/ocamlformat.ml
```diff
@@ -0,0 +1,119 @@
...
+  | Check : Input.t -> bool command
+
+let read_to_end ?(hint = 0) (inChan : in_channel) : string =
+  let buf = Buffer.create hint in
+  let rec go () =
+    try
+      let line = input_line inChan in
+      Buffer.add_string buf line;
+      Buffer.add_char buf '\n';
+      go ()
```
rgrinberg: Note that this implementation isn't tail recursive.

You should use `Io.read_all` in any case however.

rgrinberg:
Looking good. In this PR, we still need to add:

- [ ] Tests (we may install ocamlformat for this)
- [ ] Change the capabilities of the server to signal to the client that it can format documents.

cannorin:
Ok, but where the convention is documented?

ocaml-lsp-server/src/ocamlformat.ml
cannorin: If I understand correctly that function exists in ExtLib and we don't use it as of now. Is it OK to add it to the dependencies?

rgrinberg: No, it exists in stdune in fact. So it should already be available to you.

rgrinberg:
> Ok, but where the convention is documented?

Nowhere 😭 Once I switch to generated code, I will document the convention more thoroughly. For now, the simple rule is: any identifier from the LSP spec should maintain its casing. Otherwise, we should stick to the OCaml convention of `[Ff]oo_bar`

ocaml-lsp-server/src/ocamlformat.ml
cannorin: Nice! It would be great to have something like Hoogle, though.

cannorin:
rebased.

cannorin:
@rgrinberg I revert https://github.com/ocaml/ocaml-lsp/pull/89/commits/efe26727c7ee6f59962d44fc80d02a4948811c95. Stdune's `read_all` uses `in_channel_length` which is only meaningful for channels connected to regular files. In our case the channel is connected to the `stdout` of `ocamlformat` and it raises `Sys_error("Illegal seek")`.

cannorin:
I guess we should add `ocamlformat` to the CI environment and I'm not sure how.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -709,6 +731,7 @@ let start () =
   let on_request rpc state caps req =
     prepare_and_run @@ fun () -> on_request rpc state caps req
   in
+  log ~title:"info" "starting";
```
rgrinberg: This logging doesn't seem very helpful

ocaml-lsp-server/src/ocamlformat.ml
```diff
@@ -0,0 +1,118 @@
...
+
+  let default : t = []
+end
+
+type 'result command =
+  | Format_file : Input.t * 'result Output.t -> 'result command
+  | Format_files_in_place : string list -> unit command
+  | Check : Input.t -> bool command
+
+let read_to_end ?(hint = 0) (in_chan : in_channel) : string =
```
rgrinberg: Let's get rid of the optional arg if we aren't passing it anywhere.

ocaml-lsp-server/src/ocamlformat.ml
```diff
@@ -0,0 +1,118 @@
...
+  let buf = Buffer.create hint in
+  let rec go () =
+    let line = input_line in_chan in
+    Buffer.add_string buf line;
+    Buffer.add_char buf '\n';
```
rgrinberg: This seems to change the user's line endings to `\n` even if they're using `\r\n`. I think that is quite annoying for Windows users. How about using `input` instead of `input_line`?

ocaml-lsp-server/src/ocaml_lsp_server.ml
cannorin: Ah I added it for testing and forgot to remove

rgrinberg:
I pushed a commit that handles errors a little better notifies the user if the ocamlformat binary is missing.

PR #93, opened 2020-02-13
Fix TextDocumentSyncOptions

rgrinberg:
* s/didSave/save
* Add correct defaults

Fix #93 as well

@Gbury could you check if this works for you?

### DIFF
--- a/lsp/src/client_notification.ml
+++ b/lsp/src/client_notification.ml
@@ -29,10 +29,10 @@ let of_jsonrpc (r : Jsonrpc.Request.t) =
   | "workspace/didChangeConfiguration" ->
     Jsonrpc.Request.params r DidChangeConfiguration.Params.t_of_yojson
     >>| fun params -> ChangeConfiguration params
-  | "textdocument/willSave" ->
+  | "textDocument/willSave" ->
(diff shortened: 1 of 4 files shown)


Gbury:
Good news: the server now correctly receives didSave notification, so the option now works. However, upon receiving a didSave notifcation the server appears to process it as an `Unknown_notification`, as the following log suggests (in it, the server appears to receive a caml value with head constructor `Unknown_notification` instead of `didSave`):

```
# 0.00 lsp - debug
time elapsed processing message: 0.007168s
# 0.00 lsp - debug
recv: {
  "method": "textDocument/didSave",
  "jsonrpc": "2.0",
  "params": {
    "textDocument": {
      "uri": "file:///home/guigui/build/dolmen/tests/smtlib/test-002.smt2"
    }
  }
}
# 0.00 dolmen_lsp - on-notif
unknown notification
```

Gbury:
It appears there's a missing capital D for document at https://github.com/ocaml/ocaml-lsp/blob/master/lsp/src/client_notification.ml#L35

Gbury:
Thanks, this seems to have fixed it for me !

Issue #66, opened 2020-01-09
Strings that contain typable names get wrongly typed


giltho:
When hovering over a word in a string that is a correct name of a typable entity (variable name, function name or constructor), the pop up gives the type of the entity, even if completely unrelated.
Several examples :

![image](https://user-images.githubusercontent.com/8451179/72083663-1a23fb80-32fa-11ea-8e99-9f1d33fd924b.png)

![image](https://user-images.githubusercontent.com/8451179/72083807-50fa1180-32fa-11ea-91f3-f38a74e95428.png)

![image](https://user-images.githubusercontent.com/8451179/72083937-80108300-32fa-11ea-909c-4a929ad3f30b.png)

giltho:
This will probably be my first PR, I'm navigating through the code.
hover is executing a command with constructor `Type_enclosing`, which is in merlin. Merlin is usually able to understand that I'm expecting string (according to what the other extensions do), so I guess this is the wrong merlin command ? Maybe it should be calling `Type_expr` in this case ?

giltho:
Scratch that, `Type_enclosing` [should be the right request](https://github.com/ocaml-lsp/ocaml-language-server/blob/master/src/bin/server/command/getType.ts#L18). So some parameters are wrong. Is there any documentation on what the parameters mean ?
```ocaml
| Type_enclosing(* *)
    :  (string * int) option * Msource.position * int option
    -> (Location.t * [`String of string | `Index of int] * is_tail_position) list t
```
there is no documentation there :(

rgrinberg:
In general, the commands are documented in merlin's frontend. This is the `new_commands` module. For example:

```ocaml
  command "type-enclosing"
~doc:"Returns a list of type information for all expressions at given \
position, sorted by increasing size.
That is asking for type enlosing around `2` in `string_of_int 2` will return \
the types of `2 : int` and `string_of_int 2 : string`.

If `-expression` and `-cursor` are specified, the first result will be the type
relevant to the prefix ending at the `cursor` offset.

`-index` can be used to print only one type information. This is useful to
query the types lazily: normally, Merlin would return the signature of all
enclosing modules, which can be very expensive.

The result is returned as a list of:
```javascript
{
  'start': position,
  'end': position,
  'type': string,
  // is this expression not in tail position, in tail position, \
or even a tail call?
  'tail': ('no' | 'position' | 'call')
}
```"
    ~spec: [
      arg "-position" "<position> Position to complete"
[...]

PR #99, opened 2020-02-18
Add test to highlight broken hover on strings

Khady:
When a string contains a word which is a name of a type, hovering on this word will return the type name instead of string. This PR doesn't fix the problem, but it adds a test which highlights the problem.

### DIFF
--- a/ocaml-lsp-server/test/e2e/__tests__/textDocument-hover.test.ts
+++ b/ocaml-lsp-server/test/e2e/__tests__/textDocument-hover.test.ts
@@ -139,4 +139,42 @@ describe("textDocument/hover", () => {
       },
     });
   });
+
+  it("returns string type when string contains a type name", async () => {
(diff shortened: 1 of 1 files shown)


giltho:
Hey @Khady, as explained [here](https://github.com/ocaml/ocaml-lsp/issues/66), I think the issue is due to merlin itself and not us. Maybe this should be done in upstream merlin ?
Also, please tell me if I misunderstood something and the issue is actually from us and not from merlin, so that I can reopen the issue :)

Khady:
Ah I missed your issue, sorry. Yes it seems that merlin has the some problem with its original frontend.

giltho:
Given that in the current case ocaml-lsp does exactly what it's expected to do (follow merlin's output), and that [a PR has already been opened](https://github.com/ocaml/merlin/pull/1129) to fix this issue on the merlin side, I'm going to close this PR

PR #102, opened 2020-02-19
Try and recover on InternalError

### LINKED ISSUE #101: the main `lsp` entry point should catch exceptions
` Lsp.Rpc.start state handler ic oc; ` could have a catch all + logging, so that a stray `failwith` or `invalid_arg` doesn't kill the whole thing?
cc @Gbury

giltho:
Fixes #101 

However, I am not entirely sure that this is the right answer.
For example when handling client requests, should I only try and catch the exceptions in the handler or in the whole process of responding. Meaning, should the `try..with` be surrounding the whole block (like what I did) or just the 
```ocaml
handler.on_request rpc state client_capabilities req
```
What if the exception arises after a response has already been sent ? Right now this can't really happen, but if we ever add things to do after the `send_response`...

But then, if I only surround the handler, I don't know what I should return by default.

Conclusion : I think this is the right answer, but there may be slightly different choices to make

### DIFF
--- a/lsp/src/jsonrpc.ml
+++ b/lsp/src/jsonrpc.ml
@@ -271,6 +271,10 @@ module Response = struct
     [@@@end]
 
     let make ?data ~code ~message () = { data; code; message }
+
+    let of_exn exn =
+      let message = Printexc.to_string exn in
+      make ~code:InternalError ~message ()
   end
 
   type t =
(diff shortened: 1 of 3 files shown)


c-cube:
could
```ocaml
match hander.on_request … with
| res -> reply …
| exception e -> …
```
be a possible answer to limiting the scope of exceptions?

giltho:
After thinking about it, I think I'm more comfortable capturing errors only during the handling of error, and letting it die in case something happens outside of handling.

PR #104, opened 2020-02-25
Convert doc comment to markdown using omd and octavius

jorisgio:
This attempts to convert ocaml doc comment to Markdown for better display and syntax highlight in editors. 
### Vscode
Before 
2020-02-25-170824_3840x2160_scrot

After
2020-02-25-171128_3840x2160_scrot

### neovim / coc.nvim
2020-02-25-161135_3840x2160_scrot


jonludlam:
Octavius isn't being actively maintained - you should be using odoc instead (for now). See for example https://github.com/ocaml-ppx/ocamlformat/pull/721

c-cube:
Octavius seems pretty small (1.3kloc as far as I can tell) and is pretty stable, so it's a small dependency that can easily be vendored. I don't imagine the ocamldoc syntax is changing a lot?

jorisgio:
I tried to rewrite it with odoc, but it does not build because odoc is not wrapped and it clashes with compiler libs and merlin.

Kakadu:
How this change is related to details of language server protocol? I.e. how it will work when client doesn't understand markdown?

rgrinberg:
@jonludlam why doesn't odoc use wrapping?

jorisgio:
@Kakadu this PR does not change the behavior when client does not advertises markdown support. Comment is still sent as plain text ocaml markup

jonludlam:
I don't think the problem is odoc - the error in travis is:

```
Error: Files /home/opam/.opam/4.09/lib/ocaml/compiler-libs/ocamlcommon.cmxa
-        and vendor/merlin/src/ocaml/utils/utils.cmxa
-        both define a module named Config
```

It looks to me like the problem is that odoc is pulling in a dependency upon compiler-libs and that's what's conflicting with the vendored merlin's copy.

rgrinberg:
me @jonludlam talked about this offline and we've decided that the implementation in this PR that uses octavius is good enough for inclusion. Once odoc becomes more modular and introduces library components, we can revisit this implementation to use odoc.

To proceed with this PR we just need to add omd and octavius as vendored submodules & add some appropriate tests.

@jorisgio how does that sound?

jorisgio:
That sounds great. I'll have to backport some improvements i did for the odoc version and vendor the deps.

ulugbekna:
If @jorisgio doesn't mind, I could complete this work in a new PR.

jorisgio:
Hi sorry I totally forgot about that with lots of stuff going on at work. I am currently in the process of relocating half way across earth so I don't think I'll work on this in the next month at least. If you want to finish this feel free.

ulugbekna:
Could be closed now, I think.

PR #105, opened 2020-02-26
Implement textDocument/selectionRange

### LINKED ISSUE #100: Implement textDocument/selectionRange
The most recent version of the specification (3.15) introduced a `textDocument/selectionRange` request. This request is supposed to suggest available selections to the server.

This very closely corresponds to the `Shape` command in merlin (cf `query_protocol.ml`). There is just a bit of a mismatch because merlin returns a tree of shapes (every shape can have many sub shapes), so we'll need to do some intelligent flattening. This last bit we can tweak experimentally after trying out an initial implementation.

cannorin:
Closes #100.

This PR implements `textDocument/selectionRange`.

This is still WIP:
- [x] Find the desired way to flatten Merlin's `Shape` into LSP's `SelectionRange`
- [x] Change the server's capabilities
- [x] Write some tests

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -685,40 +685,49 @@ let on_request :
       Ok (store, [ change ]) )
   | Lsp.Client_request.TextDocumentOnTypeFormatting _ -> Ok (store, [])
   | Lsp.Client_request.SelectionRange { textDocument = { uri }; positions } ->
-    let selection_range_of_shapes (shapes : Query_protocol.shape list) :
+    let selection_range_of_shapes (cursor_position : Lsp.Protocol.Position.t)
+        (shapes : Query_protocol.shape list) :
         Lsp.Protocol.SelectionRange.t option =
-      match shapes with
-      | [] -> None
-      | h :: t ->
-        let merge_range r1 r2 =
-          let start_ =
-            min r1.Lsp.Protocol.Range.start_ r2.Lsp.Protocol.Range.start_
-          in
-          let end_ =
-            max r1.Lsp.Protocol.Range.end_ r2.Lsp.Protocol.Range.end_
-          in
-          { Lsp.Protocol.Range.start_; end_ }
-        in
-        let rec ranges_of_shape s =
-          range_of_loc s.Query_protocol.shape_loc
-          :: List.concat_map s.Query_protocol.shape_sub ~f:ranges_of_shape
-        in
-        let range =
-          List.fold_left
-            (List.append
-               (List.concat_map h.Query_protocol.shape_sub ~f:ranges_of_shape)
-               (List.concat_map t ~f:ranges_of_shape))
-            ~init:(range_of_loc h.Query_protocol.shape_loc)
-            ~f:merge_range
-        in
-        Some { Lsp.Protocol.SelectionRange.range; parent = None }
+      let rec ranges_of_shape parent s =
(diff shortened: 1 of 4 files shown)


ocaml-lsp-server/test/e2e/__tests__/textDocument-formatting.test.ts
```diff
@@ -108,13 +110,15 @@ describe("textDocument/formatting", () => {
...
-            "end", ""].join(newline)
+        newText: [
+          "module Test : sig",
+          "  type t =",
+          "    | Foo",
+          "    | Bar",
+          "    | Baz",
+          "end",
+          "",
+        ].join(newline),
```
rgrinberg: Why do we need these unrelated changes? Perhaps they can be in a separate PR?

cannorin: that was because of running `make fmt`. I will exclude this change.

rgrinberg: Yes it because of the recent PR that changed the ocamlformat tests? If so we can revert those changes.

rgrinberg: I reverterd those changes in master. So if you rebase, you shouldn't need these unrelated changes anymore.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -683,7 +684,52 @@ let on_request :
...
+            )
+      in
+      (* try to find the nearest range inside first, then outside *)
+      let nearest_range =
+        List.map ranges ~f:(fun r ->
+            ( r
+            , range_includes_pos r.Lsp.Protocol.SelectionRange.range
+                cursor_position ))
+        |> List.sort ~compare:(fun r1 r2 ->
+               compare (snd r1) (snd r2) |> Ordering.of_int)
```
rgrinberg: You're comparing a polymorphic variant using polymorphic comparison. This is quite suspicious to me and we should write an explicit comparison function here.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -696,28 +696,37 @@ let on_request :
...
+              Ordering.of_int (l1 - l2)
+        in
         List.map ranges ~f:(fun r ->
             ( r
             , range_includes_pos r.Lsp.Protocol.SelectionRange.range
                 cursor_position ))
-        |> List.sort ~compare:(fun r1 r2 ->
-               compare (snd r1) (snd r2) |> Ordering.of_int)
+        |> List.sort ~compare:(fun r1 r2 -> compare_inclusion (snd r1) (snd r2))
         |> List.map ~f:fst |> List.hd_opt
```
rgrinberg: This seems like you're just finding the minimum. How about we just do it in one pass without sorting/mapping.

We can start by writing `min_by : 'a list -> f:('a -> 'a -> Ordering.t) -> 'a option`

And then simply call it: `min_by ranges ~f:(fun x y -> let on r = range_includes_pos .. in compare_inclusion (on x) (on y))`.

What do you think?

cannorin: sounds good.

rgrinberg:
I've pushed a commit that moves the range math outside the main function. This will make it easier for rebase my generated code changes. Also, I've cleaned up the code as I was trying to understand it. Take a look to see if you agree with the changes.

cannorin:
Looks good!

Issue #107, opened 2020-02-26
Memory leak


jfeser:
ocaml-lsp's memory use grows when I modify a file that it's watching. I've seen it use up to 20Gb of memory. After bisecting, it seems that the bug was introduced by 8a27349f6bc5e1d2483f574ca416f726ed1c9b07. Unfortunately, it's a huge commit so I don't see an obvious culprit.

This is on macos 10.15.3, ocaml 4.08.1.

rgrinberg:
Interesting. Can you confirm that there is no leak prior to that commit?

jfeser:
Yep. Before that commit, the server allocates memory as the file is modified, but it periodically frees it. After, I don't see total memory use go down.

Kakadu:
These leaks are annoying. Please, do something!

rgrinberg:
Right away boss, we're working around the clock 😂 😂 😂

giltho:
@Kakadu if you want to contribute to fix these annoying leaks, I believe PRs to the project are most welcome :)

giltho:
I have been looking a tiny bit, and the only possible suspect I found is the use of `Stdune.Table` (and `Stdune.String.Table`)  instead of `hashtbl`s. However, given the implementation, I cannot see how that could be the cause

rgrinberg:
I've tried measuring what is leaking using spacetime before & after the commit but wasn't able to observe anything interesting either. @jfeser perhaps you can use spacetime to give us some snapshots of how the heap is evolving?

rgrinberg:
> I have been looking a tiny bit, and the only possible suspect I found is the use of Stdune.Table (and Stdune.String.Table) instead of hashtbls. However, given the implementation, I cannot see how that could be the cause

Indeed the commit seems innocuous.

ulugbekna:
I noticed memory leaks with ocamlmerlin-server a long time ago ocaml/merlin#878; don't know if related or helpful.

rgrinberg:
> I noticed memory leaks with ocamlmerlin-server a long time ago ocaml/merlin#878; don't know if related or helpful.

This is my suspicion as well. ocamllsp makes the problem worse because it introduces additional caching of the merlin state. However, this caching was present way before the commit @jfeser pointed to was introduced.

Kakadu:
> @Kakadu if you want to contribute to fix these annoying leaks, I believe PRs to the project are most welcome :)

[...]

PR #110, opened 2020-03-01
Implements didclose

giltho:
Hoping that this will help improve #107

### DIFF
--- a/ocaml-lsp-server/src/document_store.ml
+++ b/ocaml-lsp-server/src/document_store.ml
@@ -14,3 +14,8 @@ let get store uri =
   | Some doc -> Ok doc
   | None ->
     Lsp.Import.Result.errorf "no document found with uri: %a" Lsp.Uri.pp uri
+
+let remove_document store uri =
+  Table.filter_map_inplace ~f:(fun ~key ~data ->  if Lsp.Uri.equal uri key then None else Some data) store
+
+let get_size store = Table.length store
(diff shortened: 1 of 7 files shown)


ocaml-lsp-server/src/document_store.ml
```diff
@@ -14,3 +14,14 @@ let get store uri =
   | Some doc -> Ok doc
   | None ->
     Lsp.Import.Result.errorf "no document found with uri: %a" Lsp.Uri.pp uri
+
+let remove_document store uri =
```
rgrinberg: What about `Table.remove`?

giltho: Isn't `Table.remove` an alias for `Hashtbl.remove`  which 
> removes the current binding of x in tbl, restoring the previous binding if it exists.

?

rgrinberg: Somehow github swallowed my comment.

This table shouldn't have more than one binding per key, so `Table.remove` is the right thing to use. You can just push that fix directly if you want.

giltho: Are you sure ?
`Table.set` is an alias of `Table.add`, and uses `Lsp.Uri` (i.e. TextDocumentIdentifier) as a key instead of `VersionedTestDocumentIdentifier` (uri + version). It seems there is as many elements associated to an URI as there are modifications in the document since you opened the file.

I tested that by printing the size of the hashtable, and after closing a file it showed that the size of the table reduced by more than a hundred element after playing with a file for a few seconds. Maybe in that case we should use `Table.replace` instead

rgrinberg: I believe you may have found the memory leak that stdune introduced!

It's absolutely wrong to use `add` here and I thought that `set` was changed in `stdune` to be `replace`. Now you can fix 2 performance issues!

giltho: Created the PR, merging it when the check passes :)

PR #111, opened 2020-03-02
use Table.replace instead of Table.set

### LINKED ISSUE #107: Memory leak

giltho:
This supposedly fixes #107, comes after discussion in #110 with @rgrinberg 

I'm creating the PR, for history, so that it'll be linked for the issue, but merging it right away this should be very straightforward

### DIFF
--- a/lsp/src/logger.ml
+++ b/lsp/src/logger.ml
@@ -112,7 +112,7 @@ let with_sections sections f =
     | [] -> None
     | sections ->
       let table = String.Table.create (List.length sections) in
-      List.iter sections ~f:(fun section -> String.Table.set table section ());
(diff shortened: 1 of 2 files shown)


rgrinberg:
Looks good. After [this pr](https://github.com/ocaml/dune/pull/3228) I will also switch back to the `set` terminology as dune's hash tables should not have multiple bindings per key.

Issue #106, opened 2020-02-26
support refmt for Reason files


tsnobip:
Reason support was added recently, it works quite well included, what's missing now is to support `refmt` instead of `ocamlformat` for Reason files.

If it's not too complicated, I can try my luck on it, any pointer will be appreciated :)

rgrinberg:
This shouldn't be particularly difficult. Did you familiarize yourself with how the `ocamlformat` support works? https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/src/ocaml_lsp_server.ml#L645-L653

Making this work for `refmt` should be quite similar. There needs to be a way to detect if we're in reason mode, and I guess the easiest way to do this is via the document URI. Perhaps there's a better approach than this however.

You can start this by introducing a `Refmt` module in the same spirit as `Ocamlformat` and we can proceed from there.

tsnobip:
@rgrinberg I opened a PR to solve this issue, I'll update it accordingly to your remarks.

tsnobip:
Closing this issue since #112 was merged

PR #112, opened 2020-03-02
support refmt

tsnobip:
Here is a rough implementation of the support of refmt for Reason files (#106), I tested it with projects using dune and esy workflows. 

This is my first contribution to the ocaml-lsp so don't hesitate to guide me if things can be improved.

It would be useful to have a custom setting to define the formatting width (it defaults to 80 columns) since there's no `.refmt` configuration file for refmt. But I didn't find any existing use of LSP custom settings so I'm not sure how to implement that. It could be added later.


rgrinberg:
At a first glance, this looks good I'm happy that you took `ocamlformat` as a strong inspiration, but I think that we are duplicating a bit too much code between these two modules. What do you think of factoring out some common behavior?

tsnobip:
Yes sure, it felt redundant, I think it looks better like that. Tell me what you think.
If you prefer I can squash the commits into one.

tsnobip:
hmm it doesn't build because the tests failed when it cannot find refmt binary. I'm pretty new to opam workflow so I don't know how to add a dev dependency, adding reason as a normal dependency would be way too much here I think.

rgrinberg:
What about just creating our own refmt binary for the tests? It’s just a simple command line tool that process files and writes to stdout. I don’t think it’s entirely necessary to test against.

It doesn’t help that the reason is packaged on opam quite poorly, so it wouldn’t be something we’d want to impose on devs regardless

tsnobip:
what do you mean? Creating a fake refmt just for the tests? I agree with you we're actually testing refmt here (and ocamlformat) while what we'd probably want would be to test that the binary is indeed called when present and that the expected errors are raised when not.

rgrinberg:
Yes. A mock refmt (and even ocamlformat) should be enough for the tests.

tsnobip:
OK I see. Do you have any pointer on how to do that?

rgrinberg:
Here’s a simple approach:

define a binary for refmt:

(executable (name refmt))

Then some makefile hackery: (nothing is tested here, just sketching things out)

REFMT_DIR = _build/default/bla
REFMT = $(REFMT_DIR)/refmt.exe

FORCE:

$(REFMT): FORCE
  dune build $@

test: $(REFMT)
 PATH=$(REFMT_DIR):$(PATH) $(MAKE) actual_test

actual_test:
  ...

tsnobip:
as discussed in PM, I finally implemented refmt mock as a node script. If we're ok with this solution, ocamlformat can be easily mocked the same way.

rgrinberg:
> as discussed in PM, I finally implemented refmt mock as a node script. If we're ok with this solution, ocamlformat can be easily mocked the same way.

I reviewed the test commit and the approach looks quite good to me. Doing the same thing with ocamlformat is a good idea and I would certainly welcome that.

The only thing I'd suggest is for the dummy formatted to be equivalent to `cat`. Basically it should just read the file (or stdin) and output it to stdout. This will make updating the tests a little less fragile.

giltho:
I'll implement the ocamlformat mock in the github action branch based on this

tsnobip:
@rgrinberg I followed your suggestion, I changed the implementation to just piping out to StdOut everything from StdIn and updated the formatting tests accordingly.
I did the same for Ocamlformat and removed it from the test dependencies in dune, it seems to work quite well @giltho.

giltho:
Ah amazing, then I'll just rebase the github action branch on this once it is merged :)

ocaml-lsp-server/src/base_formatter.mli
```diff
@@ -0,0 +1,15 @@
...
+
+type command_result =
+  { stdout : string
+  ; stderr : string
+  ; status : Unix.process_status
+  }
+
+val run_command : string -> ?stdin_value:string -> string list -> command_result
+
+val _PATH : Stdune__.Path.t list lazy_t
```
rgrinberg: This should be `Stdune.Path`.

rgrinberg:
@tsnobip @cannorin the differences between ocalmformat and refmt weren't significant enough so I merged them into a single module. I also removed supports for modes that are unlikely to ever be used in LSP. The code should be much shorter now.

@tsnobip the tests that make sure an error message is present if a binary isn't found don't seem to be reliable enough. One could easily run them in an environment where `refmt` or `ocamlformat` are already present and get a test failure. I don't think that binary detection is fragile enough to be worth testing - the code is already mature as I extracted it from dune. What about we create mock refmt/ocamformat that raise and write to stderr? This way we can test that we forward error messages properly.

rgrinberg:
> It would be useful to have a custom setting to define the formatting width (it defaults to 80 columns) since there's no .refmt configuration file for refmt. But I didn't find any existing use of LSP custom settings so I'm not sure how to implement that. It could be added later.

Shouldn't this be something that is stored inside the repository rather than the editor? Otherwise contributors will overwrite each other's formatting changes.

imbsky:
Your opinion is quite correct, and VSCode extensions such as Prettier sometimes break the style because of it.

tsnobip:
> What about we create mock refmt/ocamformat that raise and write to stderr?

Yes I thought about that case too, but I'm not sure how to implement that, by checking if stdIn starts with a given character that is obviously incorrect, like a closing parens? 

If we don't test that and remove the tests when the binary is absent, I'm actually not sure it would be worth making any formatting test at all since the remaining tests are pretty much limited.

I didn't want to change ocamlformat implementation because I didn't know what was the road map about it but indeed I found some cases were not used at all, it looks much simpler now that it's cleaned up.

tsnobip:
> Shouldn't this be something that is stored inside the repository rather than the editor? Otherwise contributors will overwrite each other's formatting changes.

Yes, making it a workspace setting is indeed more logical in that case than a user setting, but I thought the workspace / user setting distinction was something that was taken care of by the editor itself? I don't know enough the LSP to argue about that.

giltho:
Workspace settings are just the settings given by the editor. In VSCode in particular, one defines global settings, and workspace settings. The settings sent to the editors are a mix of both, workspace settings having priority over user settings in the case of overlap between the two of them.

That being said, and without knowing much about `refmt` since I don't do much reason myself I think that:
- A PR should be submitted to refmt itself to support `.refmt` configuration files, which is pretty standard for formatting tools (`.prettyrc`, `.beautifyrc`, `.ocamlformat`). I would even suggest using the name `.refmtrc` to match the js tools conventions.
- Until such a change is made in refmt, either the parameter is non-configurable for ocaml-lsp, or we can accept workspace parameters. But in the long run, this should not be stored in the editor configuration, as pointed out by Rudi

rgrinberg:
> Yes I thought about that case too, but I'm not sure how to implement that, by checking if stdIn starts with a given character that is obviously incorrect, like a closing parens?

One simple approach is to have the binary recognize a special string as faulty. That seems like a simple enough check to implement on our side.

> If we don't test that and remove the tests when the binary is absent, I'm actually not sure it would be worth making any formatting test at all since the remaining tests are pretty much limited.

Well we are still making sure that we are calling a binary and presenting the diff to the editor in a way that it can understand it. That is quite valuable.

tsnobip:
> A PR should be submitted to refmt itself to support .refmt configuration files, which is pretty standard for formatting tools (.prettyrc, .beautifyrc, .ocamlformat). I would even suggest using the name .refmtrc to match the js tools conventions.

If I understood correctly, ReasonML core team wants refmt to be quite opiniated so the only setting they wanted to allow is column width, not sure they would accept that PR. In any case, I don't think it's that important if that setting is not configurable at first, and indeed if it is it should be done as a workspace setting.

tsnobip:
> One simple approach is to have the binary recognize a special string as faulty. That seems like a simple enough check to implement on our side.

Yes that seems easily doable.

> Well we are still making sure that we are calling a binary and presenting the diff to the editor in a way that it can understand it. That is quite valuable.

Sure, you're right.

ocaml-lsp-server/src/base_formatter.ml
```diff
@@ -0,0 +1,49 @@
...
+  { stdout : string
+  ; stderr : string
+  ; status : Unix.process_status
+  }
+
+let run_command command ?stdin_value args : command_result =
+  let command =
+    match args with
+    | [] -> command
+    | _ -> Printf.sprintf "%s %s" command (String.concat ~sep:" " args)
```
SquidDev: Would it be better to use `Unix.open_process_args_full`, rather than appending the arguments together? I could imagine this being a problem if the filename has spaces (as `ocamlformat --name=a folder/file.ml` is incorrect).

I realise this was in the code before PR, so may have been discussed already!

giltho:
> If I understood correctly, ReasonML core team wants refmt to be quite opiniated so the only setting they wanted to allow is column width, not sure they would accept that PR.

@jordwalke what do you think about the column with parameter of refmt ? How should one specify it for a workspace ? Let's make ocaml-lsp as nice as possible for Reason users :)

ocaml-lsp-server/src/ocamlformat.mli
```diff
@@ -1,44 +0,0 @@
...
-    | File of string
-end
-
-module Output : sig
-  type _ t =
-    | File : string -> unit t
-    | Stdout : string t
-end
-
-module Options : sig
```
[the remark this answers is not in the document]
rgrinberg: > We will have to set --lines option to support textDocument/{rangeFormatting, onTypeFormatting}, so I'll leave minimal support of passing options instead of completely removing it.

It should be quite simple to add back. We'll just ned to pass a range in the main function and then change `args` to generate arguments for that range.

ocaml-lsp-server/src/base_formatter.ml
rgrinberg: Good call. I suppose we accidentally omitted it in the last PR.

Khady:
> How should one specify it for a workspace ?

one can use a .editorconfig file specifying the expected line width for re and rei files. Then the editor can pass this value to the lsp server. A possible issue is that different directories in a single project can use different width.

rgrinberg:
If the editor is responsible for this then we will need to implement this logic per editor.

If refmt implements this, then this will be available to everyone automatically.

rgrinberg:
@tsnobip `yarn link` is unable to shadow the `ocamlformat` binary that I have installed. I think we need a different method of making the binaries available in `PATH`.

rgrinberg:
For now I reverted the mock setup and the reason tests. Let's try to get that in a separate PR with a test setup that is more reliable. We can merge what we have here since it's already an improvement.

tsnobip:
> @tsnobip yarn link is unable to shadow the ocamlformat binary that I have installed. I think we need a different method of making the binaries available in PATH.

You're right :( 
It makes things definitely more complicated, the only solution that comes to my mind is uninstalling the installed formatters before the test and reinstalling after, but the whole thing is becoming a  mess.

rgrinberg:
What about using workspaces? Perhaps if we define refmt in a separate package, we'll  be able to link it as a dependency in the tests?

tsnobip:
I think it'll be the same with workspaces, opam packages will still have better precedence. 

Having the binaries in the path don't make the tests fail though, do they? Couldn't we keep the tests the way they are? Even the ones testing the absence of binary could be kept if we just remove the line failing when the binary is indeed present.

rgrinberg:
I don’t think it’s a good idea. A lot of time is wasted on tests that are unreliable and don’t work the same way on everyone’s machines.

tsnobip:
You know better, I'll ping you if I have some other ideas to solve this.

PR #113, opened 2020-03-02
Always return something on request

giltho:
Changed the signature of request to return either a success (store * resp) or a Response.Error
The server should always answer requests, even in case of error, it should return an error response.

### DIFF
--- a/lsp/src/rpc.ml
+++ b/lsp/src/rpc.ml
@@ -87,7 +87,8 @@ type 'state handler =
       -> ('state * Initialize.Result.t, string) result
   ; on_request :
       'res.    t -> 'state -> Initialize.ClientCapabilities.t
-      -> 'res Client_request.t -> ('state * 'res, string) result
+      -> 'res Client_request.t
+      -> ('state * 'res, Jsonrpc.Response.Error.t) result
   ; on_notification :
       t -> 'state -> Client_notification.t -> ('state, string) result
   }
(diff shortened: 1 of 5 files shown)


rgrinberg:
One issue this PR introduced is the mass spamming of errors like this in Emacs:

```
LSP :: Invalid_argument("get_every_pattern: [{\"filename\":\"ocaml.ml\",\"start\":{\"line\":379,\"col\":32},\"end\":{\"line\":379,\"col\":40},\"ghost\":false,\"attrs\":[],\"kind\":\"value_binding\",\"children\":[{\"filename\":\"ocaml.ml\",\"start\":{\"line\":379,\"col\":36},\"end\":{\"line\":379,\"col\":40},\"ghost\":false,\"attrs\":[],\"kind\":\"pattern (ocaml.ml[379,10448+36]..ocaml.ml[379,10448+40])\\n  Tpat_var \\\"poly/17813\\\"\\n\",\"children\":[]},{\"filename\":\"ocaml.ml\",\"start\":{\"line\":379,\"col\":40},\"end\":{\"line\":379,\"col\":40},\"ghost\":true,\"attrs\":[{\"start\":{\"line\":379,\"col\":40},\"end\":{\"line\":379,\"col\":40},\"name\":\"merlin.hole\"}],\"kind\":\"expression\",\"children\":[]}]}]") [8 times]
```

giltho:
urgh
I'm trying to understand if these errors are legitimate.
This is the server answering every request, as it should, and not silencing any issue. I guess Emacs is showing the errors ?
Now, are these bugs that I introduced or are those just errors that were there before but were invisible ?

giltho:
The "get_every_pattern" error seems to be coming from a wrong call to `get_every_pattern` in `merlin/src/analysis/destruct`. This happens after calls to `code_actions` being fired
I think the source is the same for the "No node given at range" thing which is due to the `case_analysis` call in code actions

Issue #14, opened 2019-12-13
Add Azure CI


rgrinberg:
This will give us testing on Mac & Windows.

c-cube:
what about the new github-actions, btw?

imbsky:
GitHub Actions is great. The GitHub Actions team is also working on bumping the cache limit now. I definitely recommend it.

prometheansacrifice:
I have a working Azure Pipelines setup from vscode-merlin that I can contribute here. Is everyone onboard with Azure?

imbsky:
Just out of curiosity, We already have Travis and it works well. In many cases, caching is not being used very effectively and I don't think it makes much sense to use Azure. If you want to automate releases, you can do it with Travis, or GitHub Actions without an extra account. Why do you want to use Azure?

prometheansacrifice:
Caching has been working really well on vscode-merlin

prometheansacrifice:
We're seeing ~20min runs despite slow e2e tests

imbsky:
I see. Well, we can also use the caching feature on TravisCI.
https://docs.travis-ci.com/user/caching/

imbsky:
Unless we need to do something that only Azure can do, I don't think we need to force moving.

rgrinberg:
> I have a working Azure Pipelines setup from vscode-merlin that I can contribute here. Is everyone onboard with Azure?

I'm on board.

giltho:
Shall we close this since GH-actions has been merged ?

imbsky:
Yes.

PR #114, opened 2020-03-02
Github action

giltho:
So, the goal is to have testing for all platforms.
I think people were ok with azure (#14) but I have no experience with it so I just went with github actions until someone does azure.

However, for now, there is an issue: ocamlformat tests don't seem to be passing on windows. OCamlformat gets executed but returns the same thing as passed. If anyone has an idea :)


rgrinberg:
Where do I view the output for a github action? Or is this something that we need to enable first?

imbsky:
You don't' need to enable it. and you should see the results here until this PR is merged.
https://github.com/giltho/ocaml-lsp/actions

giltho:
Yes, the one I'm refering to is [this one](https://github.com/giltho/ocaml-lsp/actions/runs/48415715) to be exact.
The other seem to fail because of quantic behaviour from `setup-ocaml`...

rgrinberg:
Perhaps it makes sense to keep the current CI until the issues with setup-ocaml are resolved?

@avsm why are some of the jobs canceled?

giltho:
I restarted the last job and it is running. So it might just have been a server issue from either github or homebrew.

avsm:
It's a feature. If one job fails (in this case https://github.com/giltho/ocaml-lsp/runs/480740513?check_suite_focus=true) then the other jobs are all cancelled. See https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast for how to modify it if you want.

giltho:
Hi @avsm 
I was talking about this run for example : https://github.com/giltho/ocaml-lsp/runs/480793491?check_suite_focus=true
Where the job that died is caused by a failure in `setup-ocaml`

rgrinberg:
@giltho I guess we should change the behavior Anil mentioned above. We don't want CI to stop just because windows failed.

avsm:
No log output there @giltho -- afraid I can't do much with that unless you can find a case where it actually runs the action :-) Not seen that happen in any other uses though, so might be a GH infrastructure issue. They've had a number of outages in the past few weeks.

giltho:
Yes, unfortunately it's very hard to debug without log, I also think this might be GH bugging

giltho:
So with the fail-fast set to false, it is still very weird
https://github.com/giltho/ocaml-lsp/runs/482970879?check_suite_focus=true
Everything passes except:
 - windows because of formatting
 - unbuntu 4.06.1, with that same "silent failure" as [this previous one](https://github.com/giltho/ocaml-lsp/runs/480793491?check_suite_focus=true) but in ubuntu instead of windows

giltho:
So I have no idea what this is due to but I think what causes the absence of log is this:
```ts
try {
    let ocaml_version = core.getInput('ocaml-version');
    await installer.getOpam(ocaml_version);
  } catch (error) {
    core.setFailed(error.message);
  }
```
And inside the `installer.ts`, there's things like this:
```ts
try {
    downloadPath = await tc.downloadTool("https://cygwin.com/setup-x86_64.exe");
  } catch (error) {
    core.debug(error);
    throw `Failed to download cygwin: ${error}`;
  }
```

Now in theory, any error would be logged thanks to `core.debug`, but let's assume that somehow it's not working, then, since what is thrown is a string, then `error.message` is undefined, and `setFailed` will not show anything.

That being said, in the last actions run, none of the setup-ocaml failed.....

giltho:
Reran it and now opam depext fails in one of the runs ?! This definitely looks like a github infra thing...
Sorry to have bothered you @avsm

avsm:
That's frustrating -- I can definitely merge PRs that sprinkle more logging to help track this sort of thing down in the future. (I guess it would also fail if the Cygwin distfile site goes down, which it occasionally done)

giltho:
@rgrinberg After all, I just decided to skip the tests that don't work on windows.
However : 
 - `make build` is broken on 4.06.1 because of the generator script that uses custom let bindings.  But that should be built at all in CI I guess ?
 - There was an issue with the `textdocument/definition` and `textdocument/typeDefinition` on windows
```js
 ● textDocument/definition › returns location of a type definition

    expect(received).toMatchObject(expected)

    - Expected
    + Received

    @@ -8,8 +8,8 @@
            "start": Object {
              "character": 0,
              "line": 1,
            },
          },
    -     "uri": "file://test.ml",
    +     "uri": "file:///test.ml",
        },
      ]
```
I don't know if that's expected or not on windows or if that's an actual bug. In the meanwhile I changed it to
```js
os.type() === "Windows_NT" ? "file://test.ml" : "file:///test.ml"
```

rgrinberg:
> make build is broken on 4.06.1 because of the generator script that uses custom let bindings. But that should be built at all in CI I guess 

I think that should be fixed now. We are using ocaml-syntax-shims to provide this syntax for all versions of OCaml

> I don't know if that's expected or not on windows or if that's an actual bug. In the meanwhile I changed it to

Seems fine.

giltho:
It's finally passing yay !

rgrinberg:
Looks fine but let’s leave Travis alone for now. Once this setup is proven, we can get rid of Travis

giltho:
This is very reasonable

.github/workflows/build.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+    - uses: actions/setup-node@v1
+    - uses: giltho/setup-ocaml@master
+      with:
+        ocaml-version: ${{ matrix.ocaml-version }}
+    - uses: actions/checkout@v1
+      with:
+        submodules: true
+    - run: opam pin add lsp.dev -n .
+    - run: opam pin add ocaml-lsp-server.dev.dev -n .
```
rgrinberg, line 22: Why is `.dev` doubled here? Also, isn't `opam install . --deps-only` enough? Why the extra pin step?

giltho: The double `.dev` is a typo, and to be honest I don't really remember the process there. I'll change it and see if anything fails

giltho: Ah, the extra pin step comes from here:
https://github.com/marketplace/actions/setup-ocaml

giltho: I changed the double .dev but not the pin dev, it follow's the action's documentation, I don't know if I should remove it or not

rgrinberg: I pushed the simplified form and it seems to work just fine.

rgrinberg:
Changes are merged in #114

PR #117, opened 2020-03-03
quick fix for failures in code actions

### LINKED ISSUE #115: Code actions failures
#113 Has introduced a lot of error responses from the server that seem to be happening on codeAction requests (see note from @rgrinberg in the discussion).

giltho:
fixes #115

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -175,13 +175,13 @@ let code_action store (params : Lsp.CodeAction.Params.t) =
             (code_action_of_case_analysis params.textDocument.uri res)
         ]
       with
-      
-      | Failure _ (* TODO: find a better way of catching this, this is a bit dirty *)
+      | Failure _
+      (* TODO: find a better way of catching this, this is a bit dirty *)
       | Invalid_argument _
       | Destruct.Not_allowed _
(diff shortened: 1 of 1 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -175,6 +175,9 @@ let code_action store (params : Lsp.CodeAction.Params.t) =
             (code_action_of_case_analysis params.textDocument.uri res)
         ]
       with
+      | Failure _
+      (* TODO: find a better way of catching this, this is a bit dirty *)
+      | Invalid_argument _
```
rgrinberg: I would prefer if you just changed our vendored copy of merlin to raise a new exception. Invalid_argument can be raised by quite a few different calls, so it's too dangerous to catch it here.

giltho: That is better indeed ! 
What's the workflow there ? Should I submit a PR to the merlin repo first ?

rgrinberg: Submit a PR to my fork: https://github.com/rgrinberg/merlin/tree/remove-lsp

Submitting upstream is also good if you think your solution is good enough. I usually rebase that branch from time to time periodically.

PR #118, opened 2020-03-04
Fix incorrect type for workspace/applyEdit request

SquidDev:
[The specification][spec] states that only a single edit should be supplied. Multiple edits can be embedded inside `WorkspaceEdit.t`.

I realise that the lsp package doesn't currently have support for server requests, so fixing this may not be a high priority. I had a quick play around with bodging in support locally, and found this bug as a result.

[spec]: https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#workspace_applyEdit


rgrinberg:
Thanks for the fix!

PR #120, opened 2020-03-09
Logging through LSP

cannorin:
This PR makes the server to send log outputs to client through `window/logMessage`.

This will make debugging far easier within VSCode or other editors.

To match [the LSP spec](https://microsoft.github.io/language-server-protocol/specification#window_showMessage) and avoid an infinite loop [while logging about request sending/receiving](https://github.com/ocaml/ocaml-lsp/blob/master/lsp/src/rpc.ml#L18), this PR also introduces a type `Logger.Title.t`.

### DIFF
--- a/lsp/src/logger.ml
+++ b/lsp/src/logger.ml
@@ -27,6 +27,26 @@
 
 open Import
 
+module Title = struct
+  type t =
+    | Error
+    | Warning
+    | Info
+    | Debug
+    | LocalDebug
+    | Notify
+    | Custom of string
+
+  let to_string = function
+    | Error -> "error"
+    | Warning -> "warn"
+    | Info -> "info"
+    | Debug -> "debug"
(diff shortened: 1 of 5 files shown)


rgrinberg:
The code looks fine but I'd like to test this out first. Also, what about logging from merlin? Do you think we should redirect merlin logs here as well?

giltho:
I don't know how much logs merlin spits out. Maybe that should depend on the `trace` parameter in the initialize request ? There is not much information about `trace` for the server-side in the spec, I [raised an issue](https://github.com/microsoft/language-server-protocol/issues/943) about it.

In the meanwhile, merlin could be logged through the same process, but I guess we need to understand if it changes performances on the most used editors. Maybe some of them do not handle elegantly a heavy stream of `logMessage`

rgrinberg:
Tested this out and it seems to work fine.

PR #132, opened 2020-03-17
Implement textDocument/prepareRename

cannorin:
TBD:
- [x] Does it need a test?

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -28,6 +28,9 @@ let initializeInfo : InitializeResult.t =
     CompletionOptions.create ~triggerCharacters:[ "." ] ~resolveProvider:false
       ()
   in
+  let renameProvider =
+    `RenameOptions (Lsp.Types.RenameOptions.create ~prepareProvider:true ())
+  in
   let capabilities =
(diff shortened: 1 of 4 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -35,7 +35,10 @@ let initializeInfo : InitializeResult.t =
       ~referencesProvider:(`Bool true) ~documentHighlightProvider:(`Bool true)
       ~documentFormattingProvider:(`Bool true)
       ~selectionRangeProvider:(`Bool true) ~documentSymbolProvider:(`Bool true)
-      ~renameProvider:(`Bool true) ()
+      ~renameProvider:
+        (`RenameOptions
+          { prepareProvider = Some true; workDoneProgress = None })
```
rgrinberg: Please use the constructor for RenameOptions. This way, our code will not be broken once new fields are added

cannorin: For unknown reasons `module RenameOptions` is not included in the `Import` module and it is failing to compile:
Screenshot from 2020-03-18 17-35-28

cannorin: It's not just in the `ocaml-lsp-server/src/import.ml`. It is inconvenient for me to maintain it by hand btw.

rgrinberg:
Merged manually. Thanks.

PR #139, opened 2020-04-04
Bump minimum required dune version to 2.0

imbsky:


rgrinberg:
I think the minimum is 2.3 actually. This can be cross referenced with the dune-project file.

imbsky:
What does cross-reference mean?

rgrinberg:
I mean that the language set in dune-project and the minimum version of dune should agree.

rgrinberg:
Wait sorry, I was looking at a different project. Indeed 2.0 is enough for ocaml-lsp. Sorry for the noise.

imbsky:
No worries

Issue #141, opened 2020-04-05
ocamlformat is not listed as a dependency


yannl35133:
It seems that ocamlformat is needed :
```
[Error - 15:51:16] Request textDocument/formatting failed.
  Message: Unable to find ocamlformat binary
  Code: -32600 
[Error - 15:51:16] [lsp] Unable to find ocamlformat binary
```
(VSCode)

This was resolved when I installed it. I think it should be added as a dependency

giltho:
Hi !
This behaviour is intended. `ocamlformat` is not a requirement for the lsp to work, it is only a requirement for the formatting.
However we could probably improve the mmessage to be something like "Please install ocamlformat"

rgrinberg:
> This was resolved when I installed it. I think it should be added as a dependency

That would be a little too strict. ocamlformat is in fact not a dependency for reason projects for example.

PR #142, opened 2020-04-06
better message for formatting missing binary.

giltho:
Followup and fixing #141

### DIFF
--- a/ocaml-lsp-server/src/fmt.ml
+++ b/ocaml-lsp-server/src/fmt.ml
@@ -48,7 +48,7 @@ type error =
   | Unknown_extension of { name : string }
 
 let message = function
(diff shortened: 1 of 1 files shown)


ocaml-lsp-server/src/fmt.ml
```diff
@@ -48,7 +48,7 @@ type error =
   | Unknown_extension of { name : string }
 
 let message = function
-  | Missing_binary { binary } -> sprintf "Unable to find %s binary" binary
+  | Missing_binary { binary } -> sprintf "Unable to find %s binary. You need to install %s manually to use the formatting feature." binary binary
```
rgrinberg, line 51: Could you run `$ make fmt`? Long lines like this should be wrapped.

Issue #116, opened 2020-03-03
Batching document change notifications


jfeser:
The emacs lsp-mode package sends a lot of very granular text change notifications (https://github.com/emacs-lsp/lsp-mode/issues/362). Other lsp servers apparently batch these notifications and send diagnostics asynchronously (https://github.com/emacs-lsp/lsp-mode/issues/362#issuecomment-446549480). ocaml-lsp should probably do the same. 

Right now it looks like the type checker runs for every change?

I've looked at implementing this, but I'm not sure how to fit it in with the existing code.

giltho:
This would need an asynchronous server (using lwt or async) we do not have yet.
But in theory yes :)

rgrinberg:
Yes indeed. Although the plan is to use neither Lwt nor Async :)

jfeser:
> neither Lwt nor Async :)

What were you thinking of using? I'd be happy to take a shot at an implementation, but I don't want to make invasive changes that you don't want upstream.

rgrinberg:
I’d like to adopt dune’s fiber library. It’s a much simpler & smaller take on concurrency. Tbh, this is not a project I would recommend a new contributor to take on, but I don’t want to curb anyone’s ambitions ;)

Have a look at dune’s fiber and let me know if you have any questions. If you feel like you can take this on, we can discuss it in private.

jfeser:
I don't see code in the fiber library for doing async io. Is that somewhere else in dune?

rgrinberg:
That is something that is yet to be written. It should be something that isn’t too hard to whip up however. We don’t need anything fancy with libev, spinning up some background threads for reading/writing should be no problem.

rgrinberg:
Anyone who needs this feature can try out this branch: https://github.com/ocaml/ocaml-lsp/pull/181

rgrinberg:
Fixed in master

PR #151, opened 2020-04-15
[WIP] Batched document changes

jfeser:
Quick implementation of batching diagnostics after a document changes (#116). 

The current code will wait 2 seconds after a change before running `send_diagnostics`. If another change occurs in that time, the timer resets. The goal is to avoid running Merlin on every change to improve performance.

This code isn't ready to merge, because it busy waits for reads and for updates, but maybe it'll be a starting point for this feature.

### DIFF
--- a/lsp/src/rpc.ml
+++ b/lsp/src/rpc.ml
@@ -15,6 +15,28 @@ and state =
 
 let { Logger.log } = Logger.for_section "lsp"
 
+let threaded f x =
+  let open Fiber.O in
+  let mutex = Mutex.create () in
+  let var = ref None in
+  Thread.create
+    (fun x ->
+      Mutex.lock mutex;
+      var := Some (f x);
+      Mutex.unlock mutex)
+    x
+  |> ignore;
+
+  let rec loop () =
+    Mutex.lock mutex;
+    let v = !var in
+    Mutex.unlock mutex;
+    match v with
+    | Some v -> Fiber.return v
+    | None -> Fiber.yield () >>= fun () -> loop ()
+  in
+  loop ()
+
 let send rpc json =
   log ~title:Logger.Title.LocalDebug "send: %a"
     (fun () -> Yojson.Safe.pretty_to_string ~std:false)
@@ -58,6 +80,8 @@ let read rpc =
   | r -> Ok r
   | exception _exn -> Error "Unexpected packet"
 
+let read rpc = threaded read rpc
+
 let send_response rpc (response : Jsonrpc.Response.t) =
   let json = Jsonrpc.Response.yojson_of_t response in
   send rpc json
@@ -97,11 +121,14 @@ type 'state handler =
 
 let start init_state handler ic oc =
   let open Result.O in
-  let read_message rpc = read rpc >>= Message.of_jsonrpc in
+  let read_message rpc =
+    Fiber.map (read rpc) ~f:(fun m -> m >>= Message.of_jsonrpc)
+  in
 
   let handle_message prev_state f =
+    let open Fiber.O in
     let start = Unix.gettimeofday () in
-    let next_state = f () in
+    let+ next_state = f () in
     let ellapsed = (Unix.gettimeofday () -. start) /. 1000.0 in
     log ~title:Logger.Title.LocalDebug "time elapsed processing message: %fs"
       ellapsed;
@@ -113,13 +140,16 @@ let start init_state handler ic oc =
   in
 
   let rec loop rpc state =
+    let open Fiber.O in
     match rpc.state with
-    | Closed -> ()
+    | Closed -> Fiber.return ()
     | Ready ->
-      let next_state =
+      let* next_state =
         handle_message state (fun () ->
+            let+ msg = read_message rpc in
             let open Client_request in
-            read_message rpc >>= function
+            let open Result.O in
+            msg >>= function
             | Message.Request (id, E (Initialize params)) ->
               let* next_state, result =
                 handler.on_initialize rpc state params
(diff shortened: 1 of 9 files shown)


rgrinberg:
Thanks for tackling this. I'll try to review it this week.

Are you aware of the test failure?

```
- FAIL __tests__/TextDocument.test.ts (17.583s)
6663-   ● TextDocument: incremental sync › updates in the middle of the line
6664- 
6665-     : Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.Error:
6666- 
6667-       15 |   }
6668-       16 | 
6669-     > 17 |   it("updates in the middle of the line", async () => {
6670-          |   ^
6671-       18 |     let languageServer = await LanguageServer.startAndInitialize();
6672-       19 |     await languageServer.sendNotification("textDocument/didOpen", {
6673-       20 |       textDocument: Types.TextDocumentItem.create(
6674- 
6675-       at new Spec (node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
6676-       at Suite.<anonymous> (__tests__/TextDocument.test.ts:17:3)
6677-       at Object.<anonymous> (__tests__/TextDocument.test.ts:6:1)
```

Khady:
could the 2s be configured in some way? Unfortunately the diagnostics are not updated in background when a compilation occurs. So when I open a file, I often have to do a dummy edit to trigger a refresh of the diagnostics. Waiting 2s when the result can usually be immediate is a bit annoying.

rgrinberg:
2 seconds is indeed a bit too sluggish. We'll make it configurable, but I think a default like 750ms makes more sense.

jfeser:
I'll add a flag to configure the delay. I think the test failure is related, since there's a timeout on the test.

I'd like to get some ideas on how to remove the busy waiting. Right now an idle server spins in a loop where it polls for new messages and checks the diagnostics timers. In practice this means that every server uses 100% of a core.

lsp/src/rpc.ml
```diff
@@ -1,10 +1,67 @@
 open Import
 open Types
 
+module Event_queue = struct
```
rgrinberg, line 4: Can we move this to a separate module? I think deserves its own interface file.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -46,36 +46,102 @@ let initializeInfo : InitializeResult.t =
...
-          | _ -> DiagnosticSeverity.Error
-        in
-        let message =
-          Loc.print_main Format.str_formatter error;
-          String.trim (Format.flush_str_formatter ())
-        in
-        Diagnostic.create ~range ~message ~severity ~relatedInformation:[]
-          ~tags:[] ())
-  in
+module Diagnostics = struct
```
rgrinberg, line 49: Let's move this to its own file as well.

lsp/src/rpc.ml
```diff
@@ -95,101 +137,136 @@ type 'state handler =
...
+      let response =
+        let error =
+          Jsonrpc.Response.Error.make ~code:ServerNotInitialized
+            ~message:"not initialized" ()
+        in
+        Jsonrpc.Response.error id error
+      in
+      send_response rpc response;
+      Ok state
+    | Error err -> Error err
```
rgrinberg, line 187: Why was the bind undone here?

lsp/src/rpc.ml
```diff
@@ -95,101 +137,136 @@ type 'state handler =
...
+          | Some res -> res
+        in
+        let response = Jsonrpc.Response.ok id yojson_result in
+        send_response rpc response;
+        Ok next_state
+      | Error e ->
+        let response = Jsonrpc.Response.error id e in
+        send_response rpc response;
+        Error e.message )
+    | Error err -> Error err
```
rgrinberg, line 221: Same here. The old code used `let*` and that was a lot more readable.

ocaml-lsp-server/src/document_store.mli
```diff
@@ -8,6 +8,8 @@ val get : t -> Lsp.Uri.t -> (Document.t, Lsp.Jsonrpc.Response.Error.t) result
 
 val get_opt : t -> Lsp.Uri.t -> Document.t option
 
+val get_full_opt : t -> Lsp.Uri.t -> (Document.t * float) option
```
rgrinberg, line 11: I think we might as well make `value` public. The `float` argument is hard to understand otherwise.

rgrinberg:
Why did you revert the Fiber commit? The threaded implementation is fine, but we really don't want threads to leak into the API.

rgrinberg:
I see  you've gotten rid of the busy waiting - nice. I feel like doing a 150ms wait before every check would have been good enough as well.

rgrinberg:
#165 extracts your useful fiber refactoring without any of the batching. Let's build the matching on top of that.

jfeser:
I removed the fiber commit because I couldn't figure out how to remove the busy wait using the fiber api. The fiber api isn't thread safe, and I ran into problems mixing threads and fibers. We need at least a few threads (for reading, waiting to send diagnostics), so I figured that the fibers weren't adding much to the equation.

Alternatively, this could just be me misunderstanding how fibers are intended to be used.

What should I do with the threaded implementation? I can rebase it on the fiber refactoring if that would be helpful.

rgrinberg:
@jfeser #181 implements debouncing using fibers + thread pools. Your feedback on that work would be appreciated.

jfeser:
Awesome! I'll have a look.

PR #154, opened 2020-04-17
Windows tests tweaks

rgrinberg:
Extracted from #114


rgrinberg:
@giltho this is a small tweak of your original commits.

PR #159, opened 2020-04-20
Improve GitHub Actions configuration

imbsky:


giltho:
Oh nice they finally updated checkout v2 to handle submodules !

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,51 @@
...
+        ocaml-version:
+          - 4.09.1
+          - 4.08.1
+          - 4.06.1
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2.1.0
```
giltho, line 27:
```suggestion
        uses: actions/checkout@v2
```

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          submodules: true
+
+      - name: Use latest LTS Node.js release
+        uses: actions/setup-node@v1.4.1
```
giltho, line 32:
```suggestion
        uses: actions/setup-node@v1
```

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          node-version: 12.x
+
+      - name: Use OCaml ${{ matrix.ocaml-version }}
+        uses: avsm/setup-ocaml@v1.0
```
giltho, line 37:
```suggestion
        uses: avsm/setup-ocaml@v1
```

imbsky:
Yes, that change was a real nuisance to users. We have also been working on improving actions cache recently and I think it will be able to set up OCaml CI environment faster in the near future.

imbsky:
I knew that, but I intentionally choose this versioning method. Well, that sounds reasonable too.

giltho:
> I knew that, but I intentionally choose this versioning method. Well, that sounds reasonable too.

Ah sorry then 😅 I do find the syntax for choosing versions quite unclear though

imbsky:
I agree...

giltho:
argghhhh `@v1` is not tracking anything for `setup-ocaml`  because it's not using semantic versioning (it released `v1.0` instead of `v1.0.0`). So it should indeed be v1.0

imbsky:
It seems that setup-ocaml does not have the v1 tag.

imbsky:
Changed.

imbsky:
Thanks for your review! You and I should have been invited to the setup-ocaml repository as collaborators, so let's make some tweaks to the tag and more. For details, you should have received a message from avsm in Slack.

PR #163, opened 2020-04-22
Add diagnostic for ocamlmerlin_reason

rgrinberg:
If it's missing, the error will be reported as a diagnostic.

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -147,9 +147,7 @@ module Formatter = struct
     make_error ~code ~message ()
 
   let run rpc store doc =
-    let src = Document.source doc |> Msource.text in
-    let fname = Document.uri doc |> Lsp.Uri.to_path in
-    match Fmt.run ~contents:src ~fname with
+    match Fmt.run doc with
     | Result.Error e ->
       let message = Fmt.message e in
       let error = jsonrpc_error e in
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -46,36 +46,61 @@ let initializeInfo : InitializeResult.t =
   in
   InitializeResult.create ~capabilities ~serverInfo ()
 
+let ocamlmerlin_reason = "ocamlmerlin-reason"
+
 let send_diagnostics rpc doc =
-  let command =
-    Query_protocol.Errors { lexing = true; parsing = true; typing = true }
+  let reason_merlin_available =
+    match Document.syntax doc with
+    | Ocaml -> true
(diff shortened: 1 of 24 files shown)


rgrinberg:
@giltho what do you think of this approach?

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -47,35 +47,57 @@ let initializeInfo : InitializeResult.t =
   InitializeResult.create ~capabilities ~serverInfo ()
 
 let send_diagnostics rpc doc =
-  let command =
-    Query_protocol.Errors { lexing = true; parsing = true; typing = true }
+  let reason_merlin_available =
+    match Document.syntax doc with
+    | Ocaml -> true
+    | Reason -> Option.is_some (Bin.which "ocamlmerlin_reason")
```
imbsky, line 53:
```suggestion
    | Reason -> Option.is_some (Bin.which "ocamlmerlin-reason")
```

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -47,35 +47,57 @@ let initializeInfo : InitializeResult.t =
...
-        in
+  let uri = Document.uri doc |> Lsp.Uri.to_string in
+  match reason_merlin_available with
+  | false ->
+    let notif =
+      let diagnostics =
         let message =
-          Loc.print_main Format.str_formatter error;
-          String.trim (Format.flush_str_formatter ())
+          "Could not detect ocamlmerlin_reason. Please install reason"
```
imbsky, line 61:
```suggestion
          "Could not detect ocamlmerlin-reason. Please install reason"
```

giltho:
I like the diagnostic instead of failure, it's less noisy (especially in vscode) and it has the same clarity. 

The only remark I would make is that it probably should not fail in case the extension is not recognised. In a perfect world, it would detect the syntax, but I think it should at least default to something like `OCaml` + `Impl`. How and when the language server is activated is not described by the protocol, and the choice does not have to be made by the extension of the file. For example in VSCode, you can select the language mode to be `OCaml` on a file that is unnamed and has not been saved yet.

rgrinberg:
> The only remark I would make is that it probably should not fail in case the extension is not recognised. In a perfect world, it would detect the syntax, but I think it should at least default to something like OCaml + Impl. How and when the language server is activated is not described by the protocol, and the choice does not have to be made by the extension of the file. For example in VSCode, you can select the language mode to be OCaml on a file that is unnamed and has not been saved yet.

Good call. I've removed the dependence on extensions. Take another look.

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: Good catch. Thanks.

PR #167, opened 2020-05-02
Update build-and-test.yml

imbsky:
We recently released the v1 tag and decided to recommend its use.


giltho:
Yep, good catch forgot to update it here

PR #168, opened 2020-05-03
add command name

Nymphium:
When we want to use the lsp, we must see `ocaml-lsp-server/src/dune` to check the command name at present.
This PR adds the command name to `README.md`.


giltho:
Thanks for suggesting the add ! I just made it a tiny bit more verbose :)

imbsky:
Looks good.

PR #169, opened 2020-05-05
Remove the dune hack

giltho:
Removes the small hack added in #43 due to [this bug](https://github.com/ocaml/dune/issues/2982) that has been resolved since.


giltho:
Nope, this breaks esy's install process as well as opam's
```
opam pin add ocaml-lsp-server <git-https>
```

giltho:
I don't really know what the process is there. Should we ask people to pin `lsp` too ?
I don't know to fix this for esy before the release

andreypopp:
@giltho what's the failure with esy?

giltho:
@andreypopp same as for opam
The opam files now installs the server using
```
dune build -p ocaml-lsp-server @install
```
So I had to modify the dune file of the server [here](https://github.com/ocaml/ocaml-lsp/blob/63561e066eac6f434cfe9f8c8202e8499795cd92/ocaml-lsp-server/src/dune#L6) to call
`lsp.stdune` and `lsp.fiber`. This assumes that `lsp` is installed. Since it hasn't been released, I'm assuming esy just doesn't fetch it.
The CI doesn't fail here because both lsp and ocaml-lsp-server are pinned at the beginning

rgrinberg:
> I don't really know what the process is there. Should we ask people to pin lsp too ?

I don't think so because `ocaml-lsp-server` shouldn't depend on an lsp package publicly. So we'll still need custom build instructions so that we don't include `-p`

giltho:
This one works and is simpler

ocaml-lsp-server.opam.template
```diff
@@ -1,26 +1,13 @@
...
-  ]
-  # dune 2.0 doesn't promote *.install if --ignore-promoted-rules is passed,
-  # we can ask it to do it with --promote-install-files but it's dune 2.0 only
-  # so we do this manually to be compat with dune 1.0 and dune 2.0
-  [
-    "cp"
-    "%{_:build}%/_build/default/%{name}%.install"
-    "%{name}%.install"
+    "ocaml-lsp-server/src/main.exe"
+    "@install"
```
rgrinberg, line 9: I think we should build `ocaml-lsp-server.install` here. That should include the binary that you've written out unnecessary targets. You should also use the `--release` flag here to set the profile correctly.

giltho: Oh I didn't know building `ocaml-lsp-server.install` as a target was an option !
Done :)

johnhaley81:
I just tried to update `ocaml-lsp-server` and I'm running into errors using `esy`. I tracked it down to this PR but maybe I'm missing something?

Here's my console output:

```
info building @opam/ocaml-lsp-server@github:ocaml/ocaml-lsp:ocaml-lsp-server.opam#642124fc5d2fd566aadbbc0bb4375620775f491b@d41d8cd9
error: build failed with exit code: 1
  build log:
    # esy-build-package: building: @opam/ocaml-lsp-server@github:ocaml/ocaml-lsp:ocaml-lsp-server.opam#642124fc5d2fd566aadbbc0bb4375620775f491b
    # esy-build-package: pwd: /Users/john/.esy/3/b/opam__s__ocaml_lsp_server-c2acf39a
    # esy-build-package: running: 'dune' 'build' '-j' '4' 'ocaml-lsp-server.install' '--release'
    dune: unknown option `--release'.
    Usage: dune build [OPTION]... [TARGET]...
    Try `dune build --help' or `dune --help' for more information.
    error: command failed: 'dune' 'build' '-j' '4' 'ocaml-lsp-server.install' '--release' (exited with 1)
    esy-build-package: exiting with errors above...
    
  building @opam/ocaml-lsp-server@github:ocaml/ocaml-lsp:ocaml-lsp-server.opam#642124fc5d2fd566aadbbc0bb4375620775f491b
esy: exiting due to errors above
```

esy version 0.6.4

rgrinberg:
If anyone else is seeing this error and is confused:

You must use a recent version of dune to build ocaml-lsp. As of today, at least dune 2.5 is required.

PR #170, opened 2020-05-07
Add short contribution "guide"

rusty-key:
Added steps I had to perform on clean repo as a short contribution guide.

Also gitignored `_opam` to support local switch setup ~and added `cmdliner` to dependencies as it was missing~.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ Note that degrees of support for each LSP request are varying.
 
 ## Contribution
 
-```
+```bash
 # clone repo with submodules
 git clone --recursive git@github.com:ocaml/ocaml-lsp.git
 
--- a/README.md
+++ b/README.md
@@ -89,10 +89,10 @@ git clone --recursive git@github.com:ocaml/ocaml-lsp.git
 git submodule update --init --recursive
 
 # create local switch (or use global one)
-opam sw create . 4.09.0
+opam switch create . ocaml-base-compiler.4.09.1
 
 # install dependencies
-opam install .
+opam install . --deps-only --with-test
 
 # build
 make build
--- a/README.md
+++ b/README.md
@@ -85,10 +85,10 @@ Note that degrees of support for each LSP request are varying.
 # clone repo with submodules
 git clone --recursive git@github.com:ocaml/ocaml-lsp.git
 
-# if you already clone, pull submodules
+# if you already cloned, pull submodules
 git submodule update --init --recursive
 
-# creat local switch (or use global one)
+# create local switch (or use global one)
 opam sw create . 4.09.0
 
 # install dependencies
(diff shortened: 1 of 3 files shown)


ocaml-lsp-server.opam
```diff
@@ -23,6 +23,7 @@ depends: [
   "ocamlfind" {>= "1.5.2"}
   "ocaml" {>= "4.06"}
   "dune" {>= "2.0"}
+  "cmdliner"
```
imbsky, line 26: You should edit the dune-project instead as described at the top of the file.

giltho, line 26: cmdliner is vendored here, so it's not an opam dependency
Also, note for the future, the `opam` file is generated automatically by dune. So if cmdliner was to be added as a dependency, it should be done by adding it in the `dune-package` file

README.md
```diff
@@ -79,6 +79,28 @@ The server supports the following queries:
...
+
+```bash
+# clone repo with submodules
+git clone --recursive git@github.com:ocaml/ocaml-lsp.git
+
+# if you already clone, pull submodules
+git submodule update --init --recursive
+
+# creat local switch (or use global one)
+opam sw create . 4.09.0
```
imbsky, line 92: It's probably more clear to newcomers to use switch instead of sw. (This is just a suggestion, feel free to ignore it.)

giltho, line 92: I tried reproducing the steps, but I got this
```
opam sw create . 4.09.0
[ERROR] Compiler selection '4.09.0' is ambiguous. matching packages: { ocaml-base-compiler.4.09.0, ocaml-system.4.09.0 }
```
It depends on which system compiler you have, but let's replace this to make it explicit ?
Also, we might as well go with 4.09.1
```suggestion
opam switch create . ocaml-base-compiler.4.09.1
```

README.md
```diff
@@ -79,6 +79,28 @@ The server supports the following queries:
...
+
+# install dependencies
+opam install .
```
imbsky, line 95:
```suggestion
opam install . --deps-only --with-test
```

README.md
imbsky, line 92:
```suggestion
opam switch create . 4.09.0
```

imbsky: Oh, I missed it. (I deleted my suggestion about this to avoid confusion.)

imbsky:
@giltho I missed a few things. Thank you for correcting/supplementing them!

rusty-key:
What if I add `make init` something like:
```
init: 
	git submodule update --init --recursive
	rm -rf _opam _build
	opam switch create . ocaml-base-compiler.4.09.1
	opam install . --deps-only --with-test
```
So it's a one command bootstrapping?

imbsky:
I think it's enough to show the instruction in README.

rusty-key:
> I think it's enough to show the instruction in README.

Then it's ready.

For some reason I cannot re-request review, so please take a look

giltho:
yep LGTM :)
Merging in

giltho:
thank you !

PR #171, opened 2020-05-08
add :dev constraint to ocamlformat

bikallem:
While following the steps in https://github.com/ocaml/ocaml-lsp#contribution. I noticed that ```opam install . --deps-only --with-test``` step doesn't install ```ocamlformat``` package. Adding the constraint of ```:dev``` to ```dune-project``` fixes this. 

Signed-off-by: Bikal Gurung <gbikal@gmail.com>

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -50,7 +50,7 @@
   ocaml-syntax-shims
   ppx_yojson_conv_lib
   (ocamlformat
-   (and :dev :with-test))
+   (or :dev :with-test))
   (ocamlfind
    (>= 1.5.2))
   (ocaml
--- a/dune-project
+++ b/dune-project
@@ -1,30 +1,32 @@
 (lang dune 2.0)
+
 (using menhir 2.0)
+
 (using cinaps 1.0)
+
 (name lsp)
 
 (implicit_transitive_deps false)
 
 (license ISC)
-(maintainers "Rudi Grinberg <me@rgrinerg.com>")
-(authors
- "Andrey Popp <8mayday@gmail.com>"
- "Rusty Key <iam@stfoo.ru>"
- "Louis Roché <louis@louisroche.net>"
- "Oleksiy Golovko <alexei.golovko@gmail.com>"
- "Rudi Grinberg <me@rgrinberg.com>")
 
-(source (github ocaml/ocaml-lsp))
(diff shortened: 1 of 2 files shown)


dune-project
```diff
@@ -45,7 +49,11 @@ possible and does not make any assumptions about IO.
   menhir
   ocaml-syntax-shims
   ppx_yojson_conv_lib
-  (ocamlformat :with-test)
-  (ocamlfind (>= 1.5.2))
-  (ocaml (>= 4.06))
-  (dune (>= 2.0))))
+  (ocamlformat
+   (and :dev :with-test))
```
giltho, line 53:
```suggestion
   (or :dev :with-test))
```

We need it both for tests and for dev

imbsky:
I think `dev` isn’t for such use cases. There’s currently no such thing as npm's `devDependencies` in opam.

bikallem:
> 
> 
> I think `dev` isn’t for such use cases. There’s currently no such thing as npm's `devDependencies` in opam.

According to this, ```dev``` seems to work similar to npm's ```devDependencies```. https://discuss.ocaml.org/t/ann-opam-2-0-7-and-2-1-0-alpha/5583/22?u=bikalgurung. Note however, I haven't been able to confirm this in the opam manual since it is not documented clearly.  

I tried this setting locally, and it seems to work, i.e. install ocamlformat when --deps-only is given.

giltho:
I tend to agree with @imbsky after reading the opam documentation. Khady's answer is true, and `dev` will match what you say, however, if {dev} was to be used like this, every package would depend on `merlin {dev}`, which I've never seen anywhere

Khady:
given that [this issue](https://github.com/ocaml/opam/issues/4175) was closed, I believe that the dev annotation is meant to cover this use case.

dune-project
bikallem: Done with https://github.com/ocaml/ocaml-lsp/pull/171/commits/7554ae21cb3879ff8a37dc40189913a4fe12ac20

giltho:
So my main concern is, until we release it, pinning ocaml-lsp-server and using it for your day-to-day development will require to install ocamlformat (since it is not built from an archive but from the git repo). I don't know how esy handles opam's dev dependencies, but for example it makes no sense for a reason developer to have to install ocamlformat

PR #172, opened 2020-05-08
Allow setting line width for refmt

rusty-key:
There is currently no way of setting up line width for `refmt` other than passing flag to the command. 

While we at this point, I am suggesting to introduce `--refmt-width` argument to `ocamllsp` and pass it to `refmt`. On editor extension side a setting for refmt can be introduced and passed to `ocamllsp` instance: https://github.com/ocamllabs/vscode-ocaml-platform/pull/184

Related issue in reason repo: https://github.com/facebook/reason/issues/2567

### DIFF
--- a/ocaml-lsp-server/src/fmt.ml
+++ b/ocaml-lsp-server/src/fmt.ml
@@ -57,10 +57,11 @@ type formatter =
   | Reason of Document.Kind.t
   | Ocaml of Lsp.Uri.t
 
-let args = function
+let args ~refmt_width t =
+  match t with
   | Ocaml uri -> [ sprintf "--name=%s" (Lsp.Uri.to_path uri); "-" ]
   | Reason kind -> (
-    [ "--parse"; "re"; "--print"; "re" ]
+    [ "--parse"; "re"; "--print"; "re"; "--print-width"; string_of_int refmt_width ]
     @
     match kind with
     | Impl -> []
@@ -89,10 +90,10 @@ let exec bin args stdin =
   | Unix.WEXITED 0 -> Result.Ok res.stdout
   | _ -> Result.Error (Unexpected_result { message = res.stderr })
 
-let run doc =
+let run ~refmt_width doc =
   let open Result.O in
   let* formatter = formatter doc in
-  let args = args formatter in
+  let args = args ~refmt_width formatter in
   let* binary = binary formatter in
   let contents = Document.source doc |> Msource.text in
   exec binary args contents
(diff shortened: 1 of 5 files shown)


giltho:
Hey !
Following up on and summarizing what has been said in ocamllabs/vscode-ocaml-platform#184:
-> passing the option to the ocamllsp command line arguments is not the right way to do things

-> The best way is to pass the argument to the formatting options because it is what is intended to. However, in our case it might be a tiny bitsy more difficult to implement given that the FormattingOption type is generated and [custom fields](https://github.com/ocaml/ocaml-lsp/blob/master/lsp/src/types.mli#L1596-L1602) are not handled right now. The type generator could be extended with a mechanism that adds a `custom_` field of type
```
[> `String of string, `Bool of boolean, `Int of int, `Float of float]
```
for this `[key: string]: boolean | number | string`
Once that is done the implementation is trivial both here and in the vscode extension.

-> The 2nd best way is to use workspace configuration. But then you would need to implement the whole configuration system which is probably even more trouble.


So I have not said anything new from what has already been said in the discussion on your vscode-ocaml-platform PR apart from implementation details. Feel free to ping if you need any help.

rusty-key:
Hey @giltho! Thanks for the summary. The part I am struggling with is actually on the client side: I cannot understand when and how to pass `DocumentFormattingParams` from the client. Looking at [spec](https://microsoft.github.io/language-server-protocol/specification#textDocument_formatting) I thought that I need to register it via `ClientCapabilities` during initialization, but it looks like that's not what you are talking about. Can you direct me in the right direction?

giltho:
Ah damn... Took me a while to find it: https://github.com/microsoft/vscode-languageserver-node/blob/363e9228a87d3df51427474ecc596740c3fecfd5/client/src/codeConverter.ts#L616

The params sent by the vscode client are hardcoded... Language clients like Typescript's for VSCode are custom-written, and are not using the high-level LanguageClient class, so that's why they can use additional parameters...

So the first possiblity would be to fix the vscode-language client, but that's very much annoying, because we'd have to wait on the next vscode update....

I guess you should go with the workspace configuration option

rgrinberg:
@giltho is right. Let us know when this PR is updated and is ready for review again.

rusty-key:
@rgrinberg, I'll have some time over this weekend to refactor the PR

rgrinberg:
Please re-open when that weekend comes. Note that master already has the infrastructure for maintaining a `Configuration.t` value. So implementing this feature should be quite easy.

PR #174, opened 2020-05-11
Preview of fiber based scheduler

rgrinberg:
@jfeser @giltho, this is a preview of the scheduler required to make #151 work with fibers.

In particular, it contains two comoponents:

* A way to run computations in separate threads. We will be using this for async
  IO

* A timer mechanism to implement batched updates. This implementation slightly
  differs with #151 by having the "waker" run in a separate thread. Otherwise, this
  snippet:

```ocaml
 Option.iter !next_time ~f:(fun t ->
   Thread.delay (t -. Unix.gettimeofday ()));
```

 is problematic. What happens when a new timer that's due before `!next_time` is scheduled?

Some initial review would be helpful.

### DIFF
--- a/lsp/src/scheduler.ml
+++ b/lsp/src/scheduler.ml
@@ -210,7 +210,7 @@ let wake_loop t =
   loop ()
 
 let create () =
-  let rec t =
+  let t =
     { events_pending = 0
     ; events = Queue.create ()
     ; mutex = Mutex.create ()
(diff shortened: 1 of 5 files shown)


giltho:
This looks really nice :)
But I'm really not an expert in concurrency so I might have missed some things

giltho:
This looks really nice :) I don't think I have anything to change to that
But I'm really not an expert in concurrency so I might have missed some things

Issue #161, opened 2020-04-22
lsp-format with non-ascii char in comment.


Inkbottle007:
`ocaml-lsp` is from April 15. I use it with `emacs` and `lsp-mode`.

Here the initial file, containing a single comment, with non ascii character:
```
$ cat bin/test2.ml 
(* We’ll use *)
```

```
$ cat bin/test2.ml | hd
00000000  28 2a 20 57 65 e2 80 99  6c 6c 20 75 73 65 20 2a  |(* We...ll use *|
00000010  29 0a 0a                                          |)..|
00000013
```

Now we see how `ocamlformat` performs on it:

```
$ ocamlformat bin/test2.ml | hd
00000000  28 2a 20 57 65 e2 80 99  6c 6c 20 75 73 65 20 2a  |(* We...ll use *|
00000010  29 0a                                             |).|
00000012
```

At this point I invoke `lsp-format-buffer` on the file. Here the result:

```
$ cat bin/test2.ml | hd
00000000  28 2a 20 57 65 e2 80 99  6c 75 20 65 20 65 20 2a  |(* We...lu e e *|
00000010  29 0a                                             |).|
00000012
```
It's all mixed up.

Doing that several times did end up with the server crashing, but it might be unrelated.

giltho:
This might be the same root cause as #148

giltho:
@Inkbottle007 would you mind trying again with the last master branch and check if it still happens ?

Inkbottle007:
After `opam upgrade`, it seems I can't reproduce the issue any more.

PR #175, opened 2020-05-14
Use utf16 encoding offsets for documents

### LINKED ISSUE #148: Illegal character error.
When I use ocaml-lsp with emacs, I have a following issue: when I accidentally type in Russian and then press backspace erasing all the visible characters, I still have invisible bytes left, leading to illegal character error.

Say, when I type "дуе" instead of "let", then press backspace three times, I'm getting three errors: `Illegal character (\131); Illegal character (\181); Illegal character (\180)`. ("дуе" is represented as "\208\180\209\131\208\181", so it seems like it deletes only first byte of each symbol.)

I don't have such an issue with neither other `lsp` implementations, nor `merlin`.

giltho:
Fixes #148

Probably also #161, we'll confirm with issue author


rgrinberg:
Looks good, but you need to add uutf as a submodule and make sure that it builds with dune.

giltho:
oh yeah I forgot
Can't I just add it to the dune-project as a dependency ? Do we need to vendor it ?

rgrinberg:
Looks like there’s even an appropriate fork out there: https://github.com/dune-universe/uutf

We can’t add it as a normal dependency for a couple of reasons:

* A user might want to pick a different version of uutf for their project. This should be independent of the version of uutf their lsp server is using. It’s a bit unfortunate, but we’re just working around a limitation of opam
* We can’t use any non dune dependencies. The server should be buildable without installing ocamlbuild/findlib.

giltho:
Ah very good points ! Never thought about that
Ok, on it then :)

giltho:
Thanks for the link to the fork!

ocaml-lsp-server/src/document.ml
```diff
@@ -27,7 +27,7 @@ module Syntax = struct
     | id -> failwith ("Unexpected language id " ^ id)
 end
 
-let { Logger.log } = Logger.for_section "ocaml-lsp-server"
+(* let { Logger.log } = Logger.for_section "ocaml-lsp-server" *)
```
rgrinberg, line 30: Why are we commenting this stuff out?

ocaml-lsp-server/src/document.ml
```diff
@@ -1,4 +1,4 @@
-open Import
+(* open Import *)
```
rgrinberg, line 1: this can be `open! Import` if it's causing warnings.

ocaml-lsp-server/src/document.ml
giltho: Ah I forgot to clean that up
The logger is only used here for the logging of the whole new text every time a document changes. Which 90% of the "over-logging" we do.
I think the server is stable enough to remove this now.

rgrinberg: Yeah, I never found this message to be overly useful either.

PR #176, opened 2020-05-15
Better URI handling

### LINKED ISSUE #95: Uri.to_path un-escaping incomplete
I remarked that some special characters escape sequences are not replaced in the `Uri.to_path` function, for instance:
- `%2B` which maps to `+`
- `%3D` which maps to `=`
- `%5E` which maps to `^`

There is almost certainly other character escape sequences not currently handled (I just happened upon these in my expérimentations), and it might benefit from the automatic translation of #87 ?
In any case, this is not blocking for me, so on this one I can wait for a generic solution, :)

giltho:
fixes #95
Handle URI the same way vscode does. `to_path` and `from_path` are strongly inspired from [uriToFsPath](https://github.com/microsoft/vscode-uri/blob/master/src/index.ts#L1026-L1051) and [file](https://github.com/microsoft/vscode-uri/blob/master/src/index.ts#L749-L774) in vscode-uri.

I also vendored ocaml-uri (although it uses dune), and I had to use [this trick](https://github.com/giltho/ocaml-lsp/blob/0fa731a635432bb9ddf6e1cce988e93a3f533843/lsp/src/dune#L16-L28) because the library is also called `Uri`. There might be a better solution


giltho:
Ah 😞  Didn't think of ocaml-uri's own dependencies 😞

lsp/src/dune
```diff
@@ -12,6 +13,20 @@
 (ocamllex
  (modules text_document_text))
 
+(library
+ (name ocaml_uri)
+ (public_name lsp._ocaml_uri)
```
rgrinberg, line 18: I might be missing something, but why do we need this wrapper library?

giltho: I'm probably missing something too, but if I use `Uri` inside `uri.ml` by simply adding `uri` to the lsp libraries, I get this:

```
File "lsp/src/uri.ml", line 3, characters 9-14:
3 | type t = Uri.t
             ^^^^^
Error: The module Uri is an alias for module Lsp__Uri, which is missing
```

giltho:
@rgrinberg I don't really know what the strategy is here. Should I add `re` and `stringext` to the vendored libs too ? That seems like a lot to add just for handling URIs.... But I don't see another good way of being rfc-compliant

rgrinberg:
i guess we can skip the vendoring for now. We haven't released the package yet, so we have still room to experiment with how to package the dependencies properly.

lsp/src/dune
rgrinberg: I see. That's quite annoying. In this case, I think we should just rename our `uri.ml` to something like `uri0.ml` then. And then rename it back in `lsp.ml` and `lsp.mli`

giltho: Ah, good idea, it's a much better solution

giltho: But there's no `lsp.ml` and `lsp.mli` 😅 
I'm going to create lsp.ml

rgrinberg:
Merging this to avoid needless conflicts

PR #177, opened 2020-05-15
Jsonrpc.Session

rgrinberg:
An abstract implementation of the jsonrpc state machine. Correctly handles out
of order requests.

I'm trying to upstream `fork_and_race` first. So for now it's just a dummy
implementation. The funcrorization over `Chan` is useful while we still don't
have a Fiber based stream.

### DIFF
--- a/lsp/src/jsonrpc.ml
+++ b/lsp/src/jsonrpc.ml
@@ -11,6 +11,12 @@ module Id = struct
     | `String s -> Either.Left s
     | `Int i -> Right i
     | json -> Json.error "Id.t" json
+
+  let to_dyn x = Dyn.Encoder.opaque x
+
+  let hash x = Hashtbl.hash x
+
+  let equal = ( = )
 end
 
 module Constant = struct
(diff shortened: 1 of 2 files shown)


lsp/src/jsonrpc.ml
```diff
@@ -323,3 +329,88 @@ module Response = struct
...
+      in
+      match res with
+      | Either.Right () -> Chan.close t.chan
+      | Left (Request r) ->
+        let* resp = t.on_request r in
+        let* () = Chan.send t.chan (Response resp) in
+        loop ()
+      | Left (Response r) -> (
+        match Table.find t.pending r.id with
+        | None -> Code_error.raise "no pending request" []
```
giltho, line 384: I wouldn't trust a server/client to always be correct. Maybe it should just log and continue if it receives a response that does not correspond to a request ?

lsp/src/jsonrpc.ml
```diff
@@ -323,3 +329,88 @@ module Response = struct
...
+    let rec loop () =
+      let open Fiber.O in
+      let* res =
+        fork_and_race
+          (fun () -> Chan.recv t.chan)
+          (fun () -> Fiber.Ivar.read t.stop_requested)
+      in
+      match res with
+      | Either.Right () -> Chan.close t.chan
+      | Left (Request r) ->
```
giltho, line 378: This would be missing a `Left (Request r) when Option.is_some r.id` case ?
on_notification is never called here otherwise

rgrinberg: Good catch. I fixed it.

lsp/src/jsonrpc.ml
rgrinberg: Okay, I ignored it. It doesn't mater so much for now since we'll be dealing with well behaved clients/servers, so let's just do what's easier.

PR #180, opened 2020-05-16
Update opam metadata

rgrinberg:
* Don't use :dev dependencies
* Add more authors
* Menhir isn't an lsp dependency


rgrinberg:
Yeah, for some reason dune 2.5.0 is problematic on Windows? Perhaps @prometheansacrifice knows?

prometheansacrifice:
Turns out there is not `ocamlfind-secondary` on fdopen's fork. Earlier versions of Dune (for instance 2.4.0) were prebuilt binaries - 2.4.0b

If I'm able to port the secondary toolchain for Windows, I'll keep you all posted

PR #181, opened 2020-05-17
Reimplement jsonrpc on top of rpc and add debouncing

rgrinberg:
Will give us proper out of order request execution


rgrinberg:
Small update on how this PR is progressing:

* We now have a functioning client implementation
* There's also a new server implementation that I'm eager to switch to. The benefit of it is that it's fiber based and shares quite a lot of code with the client.
* Out of order request execution is supported.
* I'm planning to drop all the result types everywhere. I find them to be annoying in non pure code as robust error handling must catch exceptions anyway
* There's now a cute streaming abstraction: Fiber_stream. It's based on https://github.com/snapframework/io-streams and the benefit that it gives us is the ability to use our lsp client/servers in an IO agnostic way. One benefit of this is that it makes it trivial to write tests.
* The new implementation has individual threads for reading/writing on all FD's. This should make the server as non-blocking as possible. I'm not using select at the moment because it doesn't handle signals, files, x-platform, and the performance gains aren't needed anyways.

rgrinberg:
I've added debouncing. For now, it's just a fixed value of 0.25 seconds.

rgrinberg:
Support for cancelling requests was brought up in this issue https://github.com/ocamllabs/vscode-ocaml-platform/issues/247

This PR makes it possible to implement cancellation. Of course, there are still some difficulties that need to be addressed.

Right now, every request and notification is handled in a synchronous manner. In other words, we don't proceed to read the next request until we've handled the current one. This makes it easy to implement the guarantee that the internal state of the server is being updated consistently. However, it clearly breaks cancellation. I have a rough idea on how to fix this, but it will likely be done in a separate PR (let's not add too many bugs at once) and people's feedback on the design could be useful.

rgrinberg:
Are there any windows users around? Could you please try out this branch and let me know how it's working?

There are some test failures regarding termination, but hopefully things should work fine on Windows.

mnxn:
> Are there any windows users around? Could you please try out this branch and let me know how it's working?

I can confirm that this works on my Windows 10 machine. Good work.

rgrinberg:
@jfeser To clarify, all debouncing magic is occurring inside `Scheduler.schedule` and the time/waker threads. The time thread is responsible for running the scheduled events. While the wake thread is responsible for waking up the time thread whenever some timer is due. The wake thread cannot be the one that is sleeping because a new timer might always require it to wake up "ahead of time". This is the main difference with your original design.

rgrinberg:
Looks like we need this [commit](https://github.com/ocaml/ocaml-lsp/pull/181/commits/52a31c5e69ba44507680311ce9ca9d9e84b9ce36) to make things terminate. I don't quite understand why, but let's leave it for now.

rgrinberg:
Things are starting to look good now, and I'll consider merging this branch once the upstream fiber issues are resolved.

@giltho I've removed all the time logging. My opinion is that we should use stderr for more important stuff than that and make sure that all stderr output is visible during the tests. As a replacement, I suggest that we keep a table of the last N requests and their various timings. A dedicated LSP command could be used to dump/visualize this information. wdyt?

lsp/src/fiber_stream.mli
```diff
@@ -0,0 +1,40 @@
...
+  val create : ('a option -> unit Fiber.t) -> 'a t
+
+  val write : 'a t -> 'a option -> unit Fiber.t
+
+  val of_ref : 'a list ref -> 'a t
+
+  val null : unit -> 'a t
+end
+
+(** [connect i o] read from [i] and write them to [o]. Close [o] when [i] is
```
giltho, line 33:
```suggestion
(** [connect i o] reads from [i] and writes to [o]. Closes [o] when [i] is
```
More importantly, if I look at the corresponding implementation, I don't understand how the channel gets closed on exhaustion ?
```ocaml
let connect i o =
  let open Fiber.O in
  let rec go () =
    let* a = In.read i in
    let* () = Out.write o a in
    go ()
  in
  go ()
```
I might be missing something here

giltho:
re logging, I really think we should be using the logTrace feature of lsp that is a bit unofficial but that is used by VSCode. Basically, depending on the parameter in the initial `trace` value passed at initialisation, one should send `$logTraceNotifications`, and when it becomes official `$logTrace` to keep track of this kind of things. Then VSCode, or whatever client, is able to produce logs that are viewable in the lsp log inspector.

That being said, stderr should still be used for important logs and as an output for testing as you said

lsp/src/fiber_stream.mli
rgrinberg: When `i` is closed, reading it will return `None`. Once we read that `None`, we will use write it to `o` which will close it.

rgrinberg:
> Basically, depending on the parameter in the initial trace value passed at initialisation, one should send $logTraceNotifications, and when it becomes official $logTrace to keep track of this kind of things. Then VSCode, or whatever client, is able to produce logs that are viewable in the lsp log inspector.

Okay, we can do that. On the client side, where can the user set the trace level? Is this a setting we need to expose in the extension and then pass it when initializing the server?

In any case, there's no need to log timing information eagerly. It's just as simple to save this information internally and dump it on demand when the user asks for it.

rgrinberg:
Another commit makes it possible to cancel queued tasks meant to run in the thread pool. This will be used to implement request canceling.

rgrinberg:
@giltho The `Removable_queue` module could use some review.

lsp/src/fiber_stream.mli
giltho: But, looking at `In` and `Out` implementation, `Out.write` and `In.read` have the following implementations:
```ocaml
(* Out *)
let write f x = f x

(* In *)
let read t = t ()
```
I don't understand how anything gets "closed", and therefore how the loop ever stops ?
For example, in this case:
```ocaml
let fiber = connect (In.of_list []) (Out.of_ref (ref []))
```
Isn't this just going to be an infinite loop ?

giltho:
@rgrinberg Removable_queue looks good, sorry for not catching the pop issue earlier

Re logging, in VSCode, one needs to set `ocaml.trace.server` to one of these `trace?: 'off' | 'messages' | 'verbose'`. I don't know how it works in other editors. The feature is still a bit grey. I've submitted a PR to update the spec and I'm still waiting for it to be merged (https://github.com/microsoft/language-server-protocol/pull/953)

lsp/src/fiber_stream.mli
rgrinberg: Oops, I just noticed the problem and pushed a fix. We should add 

Just to clarify, a closed `In` stream is one that returned `None`. Any subsequent reads should return `None` as well. A closed `Out` stream is one that received a `None`. Writing more elements after `None` should be a no-op (perhaps it should even be an error - I'll check haskell's implementation).

rgrinberg:
> Re logging, in VSCode, one needs to set ocaml.trace.server to one of these trace?: 'off' | 'messages' | 'verbose'

I think we'll also need to pass this option when initializing the `LanguageClient`. @mnxn can we fix our client to pass this option?

PR #185, opened 2020-05-18
Upgrade to ocamlformat 0.14.2

imbsky:
After all, from 0.14.2, it seems to get the same behavior with the same configuration as before 0.14.0.


imbsky:
https://github.com/ocaml-ppx/ocamlformat/releases/tag/0.14.2

Issue #184, opened 2020-05-18
Important features no longer work on Windows


mnxn:
I recently updated ocaml-lsp to the latest version and immediately I noticed that important features weren't working on Windows:

- The language server does not see external modules and marks each reference to such module as an error. 

- Formatting .ml and .mli files does nothing with no errors.

The VS Code extension has the following errors in `main.ml` even though it builds and runs fine from the terminal.

> Unbound module Stdio
> Hint: Did you mean Stdlib?

> Unbound module My_lib

I use OCaml for Windows installed from here: https://fdopen.github.io/opam-repository-mingw/installation/

I've tried this on a clean install with only ocaml-lsp and ocamlformat installed.

The VS Code output pane does not have any useful information, it only has `time elapsed processing message`. The VS Code developer tools console also has no information.

Before this, I was running a version of ocaml-lsp that was built a few weeks ago, which was running with no problems.

giltho:
Hey @mnxn, what version of ocaml are you using ?
Also, do you have the commit hash of the previous version that you were using that worked ? It would be helpful to find the breaking commit

mnxn:
I am using OCaml 4.10.0. I unfortunately do not know exactly which version was working before. I will investigate today and try to find which update broke it.

rgrinberg:
@mnxn how are you launching the LSP server via vscode? Are you using the global toolchain or some sandbox?

mnxn:
I tried both global toolchain and opam sandbox and the behavior is the same.

rgrinberg:
Is the server responsive at all? Are you able to get any auto completion for example?

mnxn:
Auto completion and types on hover are functional.

mnxn:
I went through some of the commits and collected the following data (sorted from oldest to newest):

| Commit Hash | External Modules Work? | Formatting Works? | 
| - | - | - |
| 6f54399a99bbc7c0eb455a568ec723529877708f | Yes | Yes |
| e1073ef2da58af8e09c4c3a46f3f120a58fae355 | Yes | Yes |
| 4a03e80530712e027871529ff4d7fdb38a544202 | No | No |
| 319f4f73530fa3f95987d893473f5057266649e0 | No | No |

The culprit seems to be https://github.com/ocaml/ocaml-lsp/commit/4a03e80530712e027871529ff4d7fdb38a544202

rgrinberg:
[...]

PR #187, opened 2020-05-18
Revert switch to URI

rgrinberg:
This seems to break windows

@mnxn could you confirm this fixes things for you?

@giltho  i've reverted things in a gentle way so it's easier to introduce the
changes back.


mnxn:
> @mnxn could you confirm this fixes things for you?

This does indeed fix all the problems I had in #184. Works perfectly now.

PR #189, opened 2020-05-21
Add support for ocaml.interface language id

rgrinberg:
To support this PR:
https://github.com/ocamllabs/vscode-ocaml-platform/pull/209


rgrinberg:
good call. I also grouped all the options under an `ocamllsp` key.

PR #190, opened 2020-05-23
Test support of ocamlformat-ignore

Khady:
In the process of trying to reproduce https://github.com/ocamllabs/vscode-ocaml-platform/issues/223 I wrote this test. Apparently the problem was on my side. But you might want to keep the test.


rgrinberg:
Just a note: the interpretation of ocamlformat-ignore files is done completely by ocamlformat (thankfully). Nevertheless, I'll accept this test.

Khady:
Yes, this test mostly insures that ocamlformat is called with the right `--name` parameter.

PR #191, opened 2020-05-23
Set ocaml-lsp as source of diagnostics

Khady:
Set `ocaml-lsp` as the source of the diagnostics. It is convenient when multiple tools can provide diagnostics. For example if I have a build task in addition to the lsp server. With this addition I can see which errors are coming from ocamllsp and which are coming from my build task.

image

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -53,6 +53,7 @@ let initializeInfo : InitializeResult.t =
 let ocamlmerlin_reason = "ocamlmerlin-reason"
 
 let send_diagnostics rpc doc =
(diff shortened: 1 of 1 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -53,6 +53,7 @@ let initializeInfo : InitializeResult.t =
 let ocamlmerlin_reason = "ocamlmerlin-reason"
 
 let send_diagnostics rpc doc =
+  let diagnostic_create = Diagnostic.create ~source:"ocaml-lsp" in
```
rgrinberg, line 56: how about `ocamllsp`? That's the name of the binary at least.

Khady: Fine by me. I wasn't sure which name to use.

PR #195, opened 2020-05-26
Disable code lens for mli/rei files

rgrinberg:
The current code lens functionality only provides inferred types. Those
aren't very useful for mli files

@Khady I believe we agreed on this earlier


Khady:
I remember it was discussed, but not where. Anyway, yes they are pretty useless in interface files :+1:

PR #199, opened 2020-06-13
Handle dummy positions

rgrinberg:
Dummy positions are now turned into ranges that span the first line of
the file.

There's also a warning if merlin returns a dummy position as the
definition source.


ulugbekna:
LGTM

PR #205, opened 2020-06-21
Convert doc comments to markdown

ulugbekna:
A little reworked version of #104. Differences:

- vendored `omd` and `octavius`

Note: There was a problem with vendoring `omd` because it is being rewritten: the last published version on opam does not use dune (hence, cannot be vendored with dune) and the current version lacks necessary functionality. So I rewrote the last published version of omd to dune and vendored that repo.

- I removed grouping of various tags for now (in @jorisgio's implementation, params where combined under "Params" header, etc.). I thought it is better if we agree on the interface first before doing something like that.

The new version:

image

- added a test 

~~Note: currently, the test I wrote is failing, and I cannot figure why. The produced result matches the expected one in `console.log`.~~

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,3 +12,12 @@
 [submodule "vendor/uutf"]
 	path = vendor/uutf
 	url = https://github.com/dune-universe/uutf.git
+[submodule "vendor/octavius"]
+	path = vendor/octavius
+	url = https://github.com/ocaml-doc/octavius.git
+  ignore = dirty
+[submodule "vendor/omd"]
+	path = vendor/omd
+	url = https://github.com/ulugbekna/omd.git
+	branch = ocaml-lsp
+  ignore = dirty
--- a/ocaml-lsp-server/src/doc_to_md.ml
+++ b/ocaml-lsp-server/src/doc_to_md.ml
@@ -1,4 +1,4 @@
-open Stdune
+open Import
 
 module Oct = Octavius
 
@@ -158,7 +158,7 @@ let translate doc =
   match parse (Lexing.from_string doc) with
   | Error e ->
     let msg = Errors.message e.error in
-    log ~title:"parse doc comment" "invalid doc comments %s" msg;
+    log ~title:Logger.Title.Notify "invalid doc comments %s" msg;
     Raw (Omd.to_markdown [ Raw doc ])
   | Ok doc ->
     let doc = comment_to_markdown doc in
--- a/ocaml-lsp-server/src/doc_to_md.ml
+++ b/ocaml-lsp-server/src/doc_to_md.ml
@@ -14,10 +14,14 @@ let space = Omd.Text " "
 
 let new_line = Omd.NL
 
-let rec put_in_between elem = function
-  | [] -> []
-  | [ el ] -> [ el ]
-  | el :: tail -> [ el; elem ] @ put_in_between elem tail
+(* [put_in_between elem lst] inserts [elem] between all elements of [lst] *)
+let put_in_between elem lst =
+  let rec loop acc = function
+  | [] -> acc
+  | hd :: [] -> hd :: acc
+  | hd :: tail -> loop (elem :: hd :: acc) tail
+  in
+  List.rev (loop [] lst)
 
 let heading_level level heading =
   let open Omd in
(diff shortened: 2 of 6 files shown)


.gitmodules
```diff
@@ -12,3 +12,12 @@
...
 	path = vendor/uutf
 	url = https://github.com/dune-universe/uutf.git
+[submodule "vendor/octavius"]
+	path = vendor/octavius
+	url = https://github.com/ocaml-doc/octavius.git
+  ignore = dirty
+[submodule "vendor/omd"]
+	path = vendor/omd
+	url = https://github.com/ulugbekna/omd.git
+	branch = ocaml-lsp
```
imbsky, line 22:
```suggestion
	url = https://github.com/ocaml/omd.git
```

imbsky: Now it uses dune.

ulugbekna: The current implementation does use dune but lacks functionality such as `to_markdown: t -> string` function. 
I was hoping to merge this and then work on a PR to Omd (I expected a PR to Omd to make longer work and review time).

imbsky: Oh, I missed it. @nojb is taking care of it these days, so it might be a good idea to contact them.

nojb: Sorry coming late to the discussion; we are getting a release ready which breaks a bunch of things, so it would be useful to hear what you are using and what you would need so that we can take it into account.

nojb: OK, I see from reading the code that you are generating Markdown AST programmatically and would like to convert it to string.

To be honest it would be rather easy (easier?) to generate the markdown on your side directly instead of going through Omd. It is the parsing of Markdown that is tricky :)

But I *think* it wouldn't be hard for us to include this functionality in the new release; I will look into it.

ocaml-lsp-server/src/doc_to_md.ml
```diff
@@ -0,0 +1,157 @@
...
+let to_code_block code = Omd.Code_block (ocaml, code)
+
+let space = Omd.Text " "
+
+let new_line = Omd.NL
+
+let rec put_in_between elem = function
+  | [] -> []
+  | [ el ] -> [ el ]
+  | el :: tail -> [ el; elem ] @ put_in_between elem tail
```
rgrinberg, line 20: This function isn't tail recursive.

ocaml-lsp-server/src/doc_to_md.mli
```diff
@@ -0,0 +1 @@
+val translate : string -> [ `Raw of string | `Markdown of string ]
```
rgrinberg, line 1: Can you turn this into a normal variant?

ocaml-lsp-server/test/e2e/__tests__/textDocument-hover.test.ts
```diff
@@ -93,8 +97,98 @@ describe("textDocument/hover", () => {
...
+          \`\`\`
+          ---
+          This function has a nice documentation
+          `,
+      },
+    });
+  });
+
+  it("returns type inferred under cursor with documentation with tags (markdown formatting)", async () => {
+    languageServer = await LanguageServer.startAndInitialize({
```
rgrinberg, line 112: Is this server being terminated somewhere?

ulugbekna: Hi! Thanks for the review. Isn't this snippet from the test file responsible for terminating the server?

```ts
afterEach(async () => {
    await LanguageServer.exit(languageServer);
})
```

rgrinberg:
Thanks, that’s the case.

I noticed the tests weren’t uniform on this matter. I’ve now changed them to use afterEach for all tests.

ulugbekna:
I think we could merge the current PR as is unless there are other comments. I will work on the TODOs in later separate PRs.

ocaml-lsp-server/src/doc_to_md.ml
```diff
@@ -0,0 +1,165 @@
+open Stdune
```
rgrinberg, line 1: you should `open Import` instead.

rgrinberg:
Thanks. This PR needs rebasing now however.

rgrinberg:
I merged after applying `$ make fmt` and getting of local opens. We try to stick to only opening common modules this way. Otherwise, it's best avoided.

PR #207, opened 2020-06-24
Fix #206: Select markdown language id based on document syntax

### LINKED ISSUE #206: Hover markdown string always returns OCaml as language label
In writing the hover component in Onivim2, we discovered that the language server always returns `ocaml` as the language label for the markdown strings. (onivim/oni2#1974)

I'm pretty sure this is the only source, as it was the only one I could find, but it shows that `"ocaml"` is hardcoded into the string.
https://github.com/ocaml/ocaml-lsp/blob/058c9b44f64b6e7957f057b1078e6ba6c91cc10e/ocaml-lsp-server/src/ocaml_lsp_server.ml#L272

bryphe:
The `hover` request is hard-coded to always specify a language type of `ocaml`, but it's actually sending Reason code over in the Reason case - but the language specifier is always `ocaml`.

This selects the language id conditionally based on the syntax of the document.

Fixes #206

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -261,7 +261,12 @@ let on_request :
       | _ -> None
     in
 
-    let format_contents ~as_markdown ~typ ~doc =
+    let format_contents ~syntax ~as_markdown ~typ ~doc =
+      let languageId = 
+        match syntax with
+        | Document.Syntax.Ocaml -> "ocaml"
+        | Document.Syntax.Reason -> "reason"
+      in
       let doc =
         match doc with
         | None -> ""
@@ -269,7 +274,7 @@ let on_request :
       in
       `MarkupContent
         ( if as_markdown then
-          { MarkupContent.value = Printf.sprintf "```ocaml\n%s%s\n```" typ doc
+          { MarkupContent.value = Printf.sprintf "```%s\n%s%s\n```" languageId typ doc
           ; kind = MarkupKind.Markdown
           }
         else
@@ -284,6 +289,7 @@ let on_request :
     match query_type doc pos with
     | None -> Ok (store, None)
     | Some (loc, typ) ->
+      let syntax = Document.syntax doc in
       let doc = query_doc doc pos in
       let as_markdown =
         match client_capabilities.textDocument with
@@ -293,7 +299,7 @@ let on_request :
             ~set:(Option.value contentFormat ~default:[ Markdown ])
         | _ -> false
       in
-      let contents = format_contents ~as_markdown ~typ ~doc in
+      let contents = format_contents ~syntax ~as_markdown ~typ ~doc in
       let range = Range.of_loc loc in
       let resp = Hover.create ~contents ~range () in
       Ok (store, Some resp) )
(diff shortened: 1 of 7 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -261,15 +261,20 @@ let on_request :
       | _ -> None
     in
 
-    let format_contents ~as_markdown ~typ ~doc =
+    let format_contents ~syntax ~as_markdown ~typ ~doc =
+      let languageId = 
+        match syntax with
+        | Document.Syntax.Ocaml -> "ocaml"
+        | Document.Syntax.Reason -> "reason"
```
giltho, line 268: This is a detail, but this is the inverse function of `Document.of_language_id`, would you mind moving it in `document.ml` and name it `to_language_id` ?

bryphe: Sure thing! Thanks for the feedback.

bryphe: Added in https://github.com/ocaml/ocaml-lsp/pull/207/commits/d83470a129dbc13827bd98820d912005e51ce061

giltho:
If you're right this is annoying, I'm guessing `yarn add -g reason-cli` should be enough..
But [in theory](https://github.com/ocaml/ocaml-lsp/blob/058c9b44f64b6e7957f057b1078e6ba6c91cc10e/ocaml-lsp-server/src/ocaml_lsp_server.ml#L57-L60) the error message would be different if it just ocamlmerlin-reason missing

giltho:
my bad, it's actually `yarn global add`,  `-g` is npm...

giltho:
Ok I could see that one coming actually, I was hoping it would not happen but it did. I don't see a good way of using ocamlmerlin-reason with the opam workflow.

The best solution I can see is to create a workflow for esy + reason testing, with the adequate test suite. I'll put that on my todolist, and try to do it tomorrow before merging

bryphe:
> The best solution I can see is to create a workflow for esy + reason testing, with the adequate test suite. I'll put that on my todolist, and try to do it tomorrow before merging

This is a good idea! We could add reason as a dependency for the `esy.json`, and run the tests via esy - sounds promising 👍  Thank you for the help!

giltho:
So I started doing that, and there's a bit of an issue... `reason-cli` that contains `ocamlmerlin-reason`, has the following line in its [esy.json](https://github.com/reasonml/reason-cli/blob/master/package.json#L62): 
```json
"ocaml": "~4.2.3001"
```
while `ocaml-lsp` needs `ocaml >= 4.6`...
I might be missing something, I'm not extremely familiar with the reason environment. reason-cli hasn't been updated for 2 years, and there's no information about how to use it with `esy`
@bryphe do you know how to do that ?

How do you install ocamlmerlin-reason usually ?

andreypopp:
@giltho `@opam/reason` package should have it

giltho:
Ah nice, thank you @andreypopp ! :)
So I don't even have to set up anything to do with esy actually, this is good news

bryphe:
Thanks @giltho and @andreypopp for your help getting this in!

ocaml-lsp-server/test/e2e/__tests__/textDocument-hover.test.ts
```diff
@@ -67,6 +67,38 @@ describe("textDocument/hover", () => {
     });
   });
 
+  it("returns correct syntax for reason (markdown formatting)", async () => {
+    languageServer = await LanguageServer.startAndInitialize({
```
rgrinberg, line 71: Is the server being shut down at the end of the test somewhere?

PR #211, opened 2020-07-02
Fix #208

### LINKED ISSUE #208: Lost diagnostics after moving from commit b3cc35d to commit c6d6321 with vim-lsp
Hello,

I just noticed that I lost warning/error reports after upgrading to c6d6321 from b3cc35d.

This happens with with vim 8.2 1-1096, neovim v0.5.0-dev (nightly), and vim-lsp HEAD@ 11b1350.

I am not quite sure what do provide to better help diagnosing the issue, but I’d be happy to help as much as I can.

rgrinberg:
It's incorrect to ignore fibers outside of detach. This PR moves sending
diagnostics inside detach.


rgrinberg:
I'll be simplifying some of this fibers and scheduling stuff soon. Upstream removed `fork`, that should simplify the scheduler quite a bit.

Issue #215, opened 2020-07-03
Code completion should show documentation.


mnxn:
Language servers for other languages provide documentation on code completion.

Java:
![java](https://user-images.githubusercontent.com/25037249/86500021-2e91ff80-bd43-11ea-88fd-b7a9780c515b.png)

ocamllsp already shows documentation on hover:

![ocaml_hover](https://user-images.githubusercontent.com/25037249/86500042-44072980-bd43-11ea-8410-db30555dc587.png)

The same documentation should be visible on code completion, but it isn't:

![ocaml_compl](https://user-images.githubusercontent.com/25037249/86500058-5d0fda80-bd43-11ea-8f62-222c99217b31.png)

rgrinberg:
It does now. Things aren't ideal as we're fetching it eagerly, but @ulugbekna is working on that.

PR #217, opened 2020-07-04
Show docs on code-completion request

ulugbekna:
Implements #215. 

How it looks like (with markdown support):

image

image


rgrinberg:
Thanks. I made some style tweaks and share the markdown code between documentation and hover. The way we handle missing fields seems confusing to me, I wonder if it would be simpler to just say that if any of the fields are missing, this is the equivalent to `[]`. Defaulting the last field `[Markdown]` seems confusing to me and I can't imagine any active lsp clients that rely on this.

PR #219, opened 2020-07-06
Add support for ocamllex/menhir syntax

rgrinberg:
We don't actually support any functionality, it's just a new language type. This
is a pre-req for ml/mli switching however.

### DIFF
--- a/ocaml-lsp-server/src/document.ml
+++ b/ocaml-lsp-server/src/document.ml
@@ -20,16 +20,31 @@ module Syntax = struct
   type t =
     | Ocaml
     | Reason
+    | Ocamllex
+    | Menhir
 
-  let of_language_id = function
-    | "ocaml.interface" -> Ocaml
-    | "ocaml" -> Ocaml
-    | "reason" -> Reason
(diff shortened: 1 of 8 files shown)


ocaml-lsp-server/src/document.ml
```diff
@@ -20,16 +20,31 @@ module Syntax = struct
...
+    ; ("ocaml", Ocaml)
+    ; ("reason", Reason)
+    ; ("ocaml.ocamllex", Ocamllex)
+    ; ("ocaml.menhir", Menhir)
+    ]
+
+  let of_language_id id =
+    match List.assoc all id with
+    | Some id -> id
+    | None -> Code_error.raise "invalid language id" [ ("id", String id) ]
```
giltho, line 43: Extremely minor, but `"id"` clashes with Request ids. It's not a real clash, but it avoids a tiny confusion
```suggestion
    | None -> Code_error.raise "invalid language id" [ ("language_id", String id) ]
```

rgrinberg: Fixed.

PR #221, opened 2020-07-07
Implements "completionItem/resolve"

ulugbekna:
Implements `completionItem/resolve` for lazy fetch of docs on code completion.

`textDocument/completion` seems to take now 2-4 times less time. 

Due to the merlin's API, we need to provide it with a document and a position for it to be able to fetch docs. I modify the given document to "fake" the completion. (is there a simpler/cheaper solution?)

I will rebase when we settle on the solution. 

The PR also seems to fix (merlin's?) bug: currently, ocaml-lsp on `Hashtbl.Make<cursor>` shows an empty documentation, but with the new implementation there is documentation.

### DIFF
--- a/ocaml-lsp-server/src/compl.ml
+++ b/ocaml-lsp-server/src/compl.ml
@@ -96,7 +96,7 @@ let complete doc position ~markdown =
   let completion =
     let complete =
       Query_protocol.Complete_prefix
-        (prefix, position, completion_kinds, true, true)
+        (prefix, position, completion_kinds, false, true)
     in
     Query_commands.dispatch pipeline complete
   in
--- a/ocaml-lsp-server/src/compl.ml
+++ b/ocaml-lsp-server/src/compl.ml
@@ -1,5 +1,22 @@
 open Import
 
+module Resolve = struct
+  type t = CompletionParams.t
+
+  let uri (t : t) = Uri.t_of_yojson (`String t.textDocument.uri)
+
+  let yojson_of_t = CompletionParams.yojson_of_t
+
+  let t_of_yojson = CompletionParams.t_of_yojson
+
+  let of_completion_item (ci : CompletionItem.t) =
+    match ci.data with
+    | Some json -> t_of_yojson json
+    | None ->
+      Code_error.raise
+        "client did not pass additional data to completion resolve" []
+end
+
 let completion_kind kind : CompletionItemKind.t option =
   match kind with
   | `Value -> Some Value
@@ -16,7 +33,15 @@ let make_string chars =
   let chars = Array.of_list chars in
   String.init (Array.length chars) ~f:(Array.get chars)
 
-let prefix_of_position source position =
+(** [prefix_of_position ~short_path source position] computes prefix before
+    given [position].
+
+    @param short_path determines whether we want full prefix or cut at ["."],
+    e.g. [List.m<cursor>] returns ["m"] when [short_path] is set vs ["List.m"]
+    when not.
+
+    @return prefix of [position] in [source] and its length *)
+let prefix_of_position ~short_path source position =
   match Msource.text source with
   | "" -> ""
   | text ->
@@ -34,15 +59,44 @@ let prefix_of_position source position =
         | 'a' .. 'z'
         | 'A' .. 'Z'
         | '0' .. '9'
-        | '.'
         | '\''
         | '_' ->
           find (ch :: prefix) (i - 1)
+        | '.' ->
+          if short_path then
+            make_string prefix
+          else
+            find (ch :: prefix) (i - 1)
         | _ -> make_string prefix
     in
 
     let (`Offset index) = Msource.get_offset source position in
     find [] (index - 1)
+
+let suffix_of_position source position =
+  match Msource.text source with
+  | "" -> ""
+  | text ->
+    let len = String.length text in
+
+    let rec find suffix i =
+      if i >= len then
+        suffix
+      else
+        let ch = text.[i] in
+        (* The characters for an infix function are missing *)
+        match ch with
+        | 'a' .. 'z'
+        | 'A' .. 'Z'
+        | '0' .. '9'
+        | '\''
+        | '_' ->
+          find (ch :: suffix) (i + 1)
+        | _ -> suffix
+    in
+
+    let (`Offset index) = Msource.get_offset source position in
+    make_string (List.rev @@ find [] index)
 
 let range_prefix (lsp_position : Position.t) prefix : Range.t =
   let start =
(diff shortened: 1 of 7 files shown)


ulugbekna:
`e85c58c` may be controversial and can be removed, but code seems to look less verbose.

rgrinberg:
> e85c58c may be controversial and can be removed, but code seems to look less verbose.

It's not controversial but AFAIK, it broke 4.06 last time I tried a change like this.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -602,27 +584,39 @@ let on_request :
...
+  | TextDocumentRename req -> rename state req
+  | TextDocumentFoldingRange req -> folding_range state req
+  | SignatureHelp _ -> not_supported ()
+  | ExecuteCommand _ -> not_supported ()
+  | TextDocumentLinkResolve l -> Fiber.return @@ Ok (l, state)
+  | TextDocumentLink _ -> Fiber.return @@ Ok (None, state)
+  | WillSaveWaitUntilTextDocument _ -> Fiber.return @@ Ok (None, state)
+  | CodeAction params -> code_action rpc params
+  | CompletionItemResolve ({ data; _ } as compl) ->
+    assert (Option.is_some data);
```
rgrinberg, line 596: Might as well do `let data = Option.value_exn data` instead.

ocaml-lsp-server/src/compl.ml
```diff
@@ -128,5 +121,56 @@ let complete doc position ~markdown =
...
       List.map ~f:(fun entry -> `Replace (range, entry)) entries
   in
-  let items = List.mapi ~f:(item ~markdown) items in
+  let textDocument =
+    TextDocumentIdentifier.create ~uri:(Document.uri doc |> Lsp.Uri.to_string)
+  in
+  let compl_info =
+    CompletionParams.(
+      create ~textDocument ~position:lsp_position ()
+      |> yojson_of_t |> Json.t_of_yojson)
```
rgrinberg, line 130: Is `Json.t_of_yojson` really necessary? It's just the identity function and it exists to just play nice with the code generators.

ocaml-lsp-server/src/compl.ml
```diff
@@ -128,5 +121,56 @@ let complete doc position ~markdown =
       let range = range_prefix lsp_position prefix in
       List.map ~f:(fun entry -> `Replace (range, entry)) entries
   in
-  let items = List.mapi ~f:(item ~markdown) items in
+  let textDocument =
+    TextDocumentIdentifier.create ~uri:(Document.uri doc |> Lsp.Uri.to_string)
+  in
+  let compl_info =
+    CompletionParams.(
```
rgrinberg, line 128: Local open should be avoided.

ocaml-lsp-server/src/compl.ml
```diff
@@ -16,7 +16,7 @@ let make_string chars =
   let chars = Array.of_list chars in
   String.init (Array.length chars) ~f:(Array.get chars)
 
-let prefix_of_position source position =
+let prefix_of_position ?(short = false) source position =
```
rgrinberg, line 19: Let's make this argument mandatory and add a doc string to explain its purpose (or pick a better name perhaps?)

rgrinberg:
By the way, I just remembered that `completionItem/resolve` is absurdly slow in Emacs + lsp-mode. I wanted to avoid doing this,  but I don't think it's possible - we need a way to detect the client.

I propose that we introduce a `Client` module:

```ocaml
type t =
  | Emacs
  | Other
```

Make it a part of `State`. Construct an appropriate value from `clientInfo` on initialization. Wdyt?

rgrinberg:
Here's the clientInfo from Emacs for reference:

```ocaml
  "clientInfo": {
    "name": "emacs",
    "version": "GNU Emacs 26.3 (build 1, x86_64-apple-darwin19.3.0, Carbon Version 162 AppKit 1894.3)\n of 2020-03-23"
  },
```

ocaml-lsp-server/src/ocaml_lsp_server.ml
ulugbekna: I do use `Option.value_exn` later, but wanted to document the assumption and thought assert failure would have a more clear stack trace. Removed.

ocaml-lsp-server/src/compl.ml
ulugbekna: You're right. Removed.

ocaml-lsp-server/src/compl.ml
ulugbekna: I thought your comment on local opens to my previous PR was only for `let open ... in`. Here I thought the readability doesn't suffer. Fixed.

ocaml-lsp-server/src/compl.ml
ulugbekna: If `short` is set, the prefix computation for `List.m<cursor>` returns just `m`; otherwise, it returns prefix `List.m`. I first called the param as `short_path` but opted for brevity. Both documenting or giving a more descriptive name (any ideas?) sound good to me.

ulugbekna:
> Make it a part of State. Construct an appropriate value from clientInfo on initialization. Wdyt?

Keeping client info in state sounds like something we would want at some point. A couple of questions:

1. Can't slow `completionItem/resolve` be solved on the `lsp-mode` side? If not, then we could handle docs differently for emacs users. 
2. Should client info code go in this PR?

ocaml-lsp-server/src/compl.ml
rgrinberg: Both sound fine. `short_path` sounds clearer to me.

rgrinberg:
> Can't slow completionItem/resolve be solved on the lsp-mode side? If not, then we could handle docs differently for emacs users.

Let me see if this was improved. Last time I asked, I was told it's a limitation of the completion framework in Emacs.

> Should client info code go in this PR?

It can be done separately if you'd like. I don't mind either way.

ulugbekna:
Addressed all comments and rebased. 

As per client info, I would prefer for it to go another PR. However, if we don't want to ruin the emacs users' experience with this PR merged before lsp-mode fix lands (if ever), I can think of adding it in this PR but I guess that would mean that docs would be fetched eagerly on emacs.

rgrinberg:
> As per client info, I would prefer for it to go another PR. However, if we don't want to ruin the emacs users' experience with this PR merged before lsp-mode fix lands (if ever), I can think of adding it in this PR but I guess that would mean that docs would be fetched eagerly on emacs.

Let's leave the Emacs stuff alone for now actually. I'd like to test things again first as it seems like there were some changes in lsp-mode.

Could you add completionItem/resolve test by the way?

ulugbekna:
We need to compute the "suffix" because when we type `List.mp`, the `prefix_of_position ~short_path=true` computes `List.m` instead of `List.mp`, which when applying completion resulted in `List.mapp` passed to merlin.

ocaml-lsp-server/src/compl.ml
```diff
@@ -31,16 +31,20 @@ let completion_kind kind : CompletionItemKind.t option =
 
 let make_string chars =
   let chars = Array.of_list chars in
-  String.init (Array.length chars) ~f:(Array.get chars)
+  let arr_len = Array.length chars in
+  (String.init (Array.length chars) ~f:(Array.get chars), arr_len)
```
rgrinberg, line 35: Why do we return the length here? The caller can always obtain it with `String.length`

ulugbekna: As `make_string` anyway computes the length of the string, I thought why to do that twice.

rgrinberg:
String.length is constant time anyways. It’s just a field access

rgrinberg:
Thanks. I made `~short_path` mandatory because there were only 2 callers and squashed some commits. Merging.

PR #225, opened 2020-07-13
Remove foldingRangeProvider server capability

ulugbekna:
If the lsp server doesn't have `foldingRange` capability, vs code (and maybe other editors?) provides more granular folding ranges (based on indentation perhaps). 

| ocaml-lsp folding ranges | vs code folding ranges|
|---  | ---   |
| image | image|

With this initial state of the PR people can try which version works best for them. 

#### Actions based on feedback

- if the folding is better without this server capability on all editors, we can go on and remove all folding range handling code.
- if this only works for vs code users, we can either create a PR to vscode-ocaml-platform extension removing this client capability or add client-based logic to ocaml-lsp

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -48,7 +48,7 @@ let initialize_info : InitializeResult.t =
       ~referencesProvider:(`Bool true) ~documentHighlightProvider:(`Bool true)
       ~documentFormattingProvider:(`Bool true)
       ~selectionRangeProvider:(`Bool true) ~documentSymbolProvider:(`Bool true)
-      ~foldingRangeProvider:(`Bool true) ~experimental ~renameProvider ()
+      ~experimental ~renameProvider ()
   in
   let serverInfo =
     let version = Version.get () in


rgrinberg:
The folding mode provided by the server isn't based on indentation hence is more accurate. It should stay as the default.

If you'd prefer indentation based folding, what about configuring your editor to do so?

By the way, it's easy enough to make the folds in the lsp server more granular. Just add sections based on patterns clauses.

ulugbekna:
> If you'd prefer indentation based folding, what about configuring your editor to do so?

Fair enough, but I was more aiming at better defaults for ocaml-lsp. Without this capability, we could as well win efficiency by offloading folding range request handling. 

> Just add sections based on patterns clauses. 

Could you elaborate?

rgrinberg:
Sure, I’ll elaborate. Currently, folding currently relies on merlin’s Outline command. This command mostly dumps toplevel values in a consistent way. To make it work more like traditional folding, we’d want it fold the cases in a match expression. Seems like that wouldn’t be so hard to add, but it would probably require us to create a Folding rather than an Outline command in merlin. Since users of Outline aren’t going to be interested in seeing all this detail.

ulugbekna:
Thanks! Will try to look into it some time later.

PR #226, opened 2020-07-14
fix Infix completion

### LINKED ISSUE #51: LSP: completion of infix operators in incorrect
```ocaml
open Angstrom

let f = word >>CURSOR
```

With cursor on CURSOR, ask for completion. Then select `>>|` in the answer. The final result is

```ocaml
open Angstrom

let f = word >>>>|
```

It looks like the prefix is ignored when completing an operator.

giltho:
Fixes #51 

Prefix lookup is now taking infix operator characters into account, and prefix is used to generate a textedit that overrides it.

### DIFF
--- a/ocaml-lsp-server/test/e2e/__tests__/completionItem-resolve.test.ts
+++ b/ocaml-lsp-server/test/e2e/__tests__/completionItem-resolve.test.ts
@@ -3,7 +3,7 @@ import * as LanguageServer from "./../src/LanguageServer";
 
 import * as Types from "vscode-languageserver-types";
 
-xdescribe("textDocument/completion", () => {
+describe("textDocument/completion", () => {
   let languageServer: LanguageServer.LanguageServer = null;
 
   async function openDocument(source) {
@@ -21,7 +21,7 @@ xdescribe("textDocument/completion", () => {
     label: string,
     position: Types.Position,
   ) {
-    let response = await languageServer.sendRequest("completionItem/resolve", {
+    return languageServer.sendRequest("completionItem/resolve", {
       label: label,
       data: {
         textDocument: {
@@ -51,13 +51,15 @@ xdescribe("textDocument/completion", () => {
       Types.Position.create(0, 5),
     );
 
-    expect(response).toMatchObject({
-      documentation: {
-        kind: "markdown",
-        value:
-          "`List.map2 f [a1; ...; an] [b1; ...; bn]` is\n`[f a1 b1; ...; f an bn]`.\nRaise `Invalid_argument` if the two lists are determined\nto have different lengths. Not tail-recursive.",
-      },
-    });
+    expect(response).toMatchInlineSnapshot(`
+      Object {
+        "documentation": " [List.map2 f [a1; ...; an] [b1; ...; bn]] is
+         [[f a1 b1; ...; f an bn]].
+         Raise [Invalid_argument] if the two lists are determined
+         to have different lengths.  Not tail-recursive. ",
+        "label": "map2",
+      }
+    `);
   });
 
   it("can get documentation at arbitrary position", async () => {
@@ -70,12 +72,12 @@ xdescribe("textDocument/completion", () => {
       Types.Position.create(0, 5),
     );
 
-    expect(response).toMatchObject({
-      documentation: {
-        kind: "markdown",
-        value: "`find_all` is another name for `List.filter`.",
-      },
-    });
+    expect(response).toMatchInlineSnapshot(`
+      Object {
+        "documentation": " [find_all] is another name for {!List.filter}. ",
+        "label": "find_all",
+      }
+    `);
   });
 
   it("can get documentation at arbitrary position (before dot)", async () => {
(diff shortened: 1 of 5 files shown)


ulugbekna:
I'm afraid it broke code completion on VS code. 

image

ulugbekna:
I wrote my (crude) vision of how we could do code completion for infix operators (only starting with '>' for this example) [here](https://github.com/ulugbekna/ocaml-lsp/tree/infix-completion).

giltho:
Yeah, I didn't merge because I wanted to add some more tests I wasn't 100% confident about it.
However, I don't see how that could break completion

ulugbekna:
The offending line is in `complete` fn: 

```ocaml
...
let items =
    completion.entries |> List.map ~f:(fun entry -> `Replace (range, entry))
...
```

I think the way we create `\`Replace` may be wrong. `prefix_of_position` with `short_path=false` returns `List.m` on instead of the desired `m`, which should be replaced with, say, `map`. So we should do smth like this

```ocaml
  let range =
    range_prefix lsp_position
      (prefix_of_position ~short_path:true (Document.source doc) position)
  in
  let items =
    completion.entries |> List.map ~f:(fun entry -> `Replace (range, entry))
```

(with these lines, identifier completion seems to work well but not infix operators)

ulugbekna:
With ">" added to trigger characters, completion for infix operators seems to work as well. 

image

giltho:
The issue isn't triggering it's that when one choses `>>=` in the list of possibilities, it adds it entirely instead of also replacing the beginning of it (#51).
I had to trigger it myself.

So you're right, I should use the short range for replacement. However, I don't know why triggering on `.` got killed.
There is something happening with VSCode that I don't understand.

If I write the following code :
```ocaml
module Test = struct
   let some_value = 0
end

let x = Test.CURSOR
```
Completion isn't triggered, and even if I trigger it manually, I get "No suggestions".
However, looking at the logs, the server does send the (almost) right item :

```jsonc
"items": [
        {
            "label": "some_value",
            "kind": 12,
            "detail": "int",
            "deprecated": false,
            "sortText": "0000",
            "textEdit": {
                "range": {
                    "start": {
                        "line": 6,
                        "character": 8 // This part is wrong, the range should be of size 0
                    },
                    "end": {
                        "line": 6,
                        "character": 13
                    }
                },
                "newText": "some_value"
            }
            ...
      }
]
```

For some reason, vscode is filtering it
I'm gonna fix the range and see if this still happens

giltho:
Yep, I'm guessing VSCode is clever and filters completion items that override too much of the document, it's now fixed

ulugbekna:
I don't get completion on infix operators on vs code. Strange 🤔

image

giltho:
Yes, you have to trigger it (alt + escape).
But it should still be correct for people who customise it to be triggered
on other characters

giltho:
@rgrinberg I noticed you never used snapshot testing, do you have anything against it ? I think using inline snapshots doesn't reduce clarity but there are arguments against snapshot so I can change if needed

rgrinberg:
Sanpshot tests are fine with me. I haven't used them because I don't know how to use them.

ocaml-lsp-server/test/e2e/__tests__/completionItem-resolve.test.ts
```diff
@@ -3,7 +3,7 @@ import * as LanguageServer from "./../src/LanguageServer";
 
 import * as Types from "vscode-languageserver-types";
 
-xdescribe("textDocument/completion", () => {
```
ulugbekna, line 6: Ah, thanks for fixing this up. I followed the example of completion tests and used `xdescribe()` without knowing it's an alias to skip tests 😅

ocaml-lsp-server/test/e2e/__tests__/completionItem-resolve.test.ts
```diff
@@ -70,12 +72,12 @@ xdescribe("textDocument/completion", () => {
...
-    });
+    expect(response).toMatchInlineSnapshot(`
+      Object {
+        "documentation": " [find_all] is another name for {!List.filter}. ",
+        "label": "find_all",
+      }
+    `);
   });
 
   it("can get documentation at arbitrary position (before dot)", async () => {
```
ulugbekna, line 83: This should in reality test whether the docs for `LargeFile` are fetched in `Stdlib.LargeFil.in_channel_length`. So the request and response should look like 

```typescript
    let response = await queryCompletionItemResolve(
      "LargeFile",
      Types.Position.create(0, 15),
    );

    expect(response).toMatchInlineSnapshot(`
      Object {
        "documentation": " Operations on large files.
        This sub-module provides 64-bit variants of the channel functions
        that manipulate file positions and file sizes.  By representing
        positions and sizes by 64-bit integers (type [int64]) instead of
        regular integers (type [int]), these alternate functions allow
        operating on files whose sizes are greater than [max_int]. ",
        "label": "LargeFile",
      }
    `);
```

I can commit fix to this PR.

ocaml-lsp-server/test/e2e/__tests__/completionItem-resolve.test.ts
giltho: No worry haha. I also just discovered it, the name really is confusing. I disabled it again on some ocaml versions because order changes and it kills the tests.

ocaml-lsp-server/test/e2e/__tests__/completionItem-resolve.test.ts
giltho: That would be great if you could do that :)

PR #229, opened 2020-07-21
Small Scheduler refactoring

ulugbekna:


lsp/src/scheduler.ml
```diff
@@ -110,9 +110,7 @@ end = struct
     t
 
   let add_work (type a) (t : a t) (w : a) =
-    ( match t.state with
-    | Running _ -> ()
-    | _ -> Code_error.raise "invalid state" [] );
+    if not (is_running t) then Code_error.raise "invalid state" [];
```
rgrinberg, line 113: How do callers know when calling `add_work` is allowed?

ulugbekna: I don't quite get the question. Do you mean that previously callers could look into the implementation and see when they can safely call `add_work`? Isn't that just a bad practice? 

I think we could either just document that the function can throw an exception or return `_ result` but not rely that the caller reads the implementation.

rgrinberg: Sorry, I misread the diff. The cleanup looks right to me.

PR #230, opened 2020-07-26
[WIP] Cancellation support

ulugbekna:


ulugbekna:
The last commit is a step forward in cancelling requests in jsonrpc, but it's not working (tbh don't know why yet) - the request is not cancelled.

lsp/src/client_notification.ml
```diff
@@ -30,6 +32,7 @@ let yojson_of_t = function
   | TextDocumentDidChange params ->
     DidChangeTextDocumentParams.yojson_of_t params
   | TextDocumentDidClose params -> DidCloseTextDocumentParams.yojson_of_t params
+  | CancelRequest params -> CancelParams.yojson_of_t params
```
rgrinberg, line 35: The changes introducing `CancelRequest` can be merged.

lsp/src/jsonrpc.mli
```diff
@@ -135,4 +135,6 @@ end) : sig
   val notification : _ t -> Message.notification -> unit Fiber.t
 
   val request : _ t -> Message.request -> Response.t Fiber.t
+
+  val try_cancel_request : _ t -> [ `Cancelled | `Failed ] Fiber.t
```
rgrinberg, line 139: I'm not sure that this is the right API. We should be able to ask for a cancellation on an arbitrary request (given the Id).

lsp/src/client_notification.ml
rgrinberg: I cherry picked this into #322

PR #233, opened 2020-08-18
implement handling 'didSwitchImplIntf' - switching between impl and intf files

ulugbekna:
- [x] implement switching between impl and intf files if it exists
- [x] propose possible filenames if the opposite file doesn't exist, i.e., if a user is trying to switch from 'foo.ml', s/he will be proposed filenames `foo.mli`, `foo.mly`, `foo.mll`, `foo.rei` (I am also working on handling this on vscode extension side but bindings are hard); each file has a list of extensions that it maps to in the order of most likely to least likely
- [x] write tests

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -40,7 +40,12 @@ let initialize_info : InitializeResult.t =
   let capabilities =
     let experimental =
       `Assoc
-        [ ("ocamllsp", `Assoc [ ("interfaceSpecificLangId", `Bool true) ]) ]
+        [ ( "ocamllsp"
+          , `Assoc
+              [ ("interfaceSpecificLangId", `Bool true)
+              ; Switch_impl_intf.capability
+              ] )
+        ]
     in
     ServerCapabilities.create ~textDocumentSync ~hoverProvider:(`Bool true)
       ~definitionProvider:(`Bool true) ~typeDefinitionProvider:(`Bool true)
@@ -508,6 +513,7 @@ let definition_query (state : State.t) uri position merlin_request =
   let result = location_of_merlin_loc uri result in
   Ok (result, state)
 
+(** handles requests for OCaml (syntax) documents *)
 let ocaml_on_request :
     type resp.
        State.t Server.t
@@ -652,9 +658,23 @@ let on_request :
     let+ doc = Document_store.get_opt store uri in
     Document.syntax doc
   in
-  match syntax with
-  | Some (Ocamllex | Menhir) -> not_supported ()
-  | _ -> ocaml_on_request server req
+  match req with
+  | Client_request.UnknownRequest { meth; params } -> (
+    match
+      [ (Switch_impl_intf.meth, Switch_impl_intf.on_request) ]
+      |> List.assoc_opt meth
+    with
+    | None ->
+      Fiber.return
+        (Error
+           (make_error ~code:InternalError ~message:"Unknown method"
+              ~data:(`Assoc [ ("method", `String meth) ])
+              ()))
+    | Some handler -> handler ~params state )
+  | _ -> (
+    match syntax with
+    | Some (Ocamllex | Menhir) -> not_supported ()
+    | _ -> ocaml_on_request server req )
 
 let on_notification server (notification : Client_notification.t) :
     State.t Fiber.t =
--- a/ocaml-lsp-server/test/e2e/__tests__/textDocument-completion.test.ts
+++ b/ocaml-lsp-server/test/e2e/__tests__/textDocument-completion.test.ts
@@ -3,7 +3,9 @@ import * as LanguageServer from "./../src/LanguageServer";
 
 import * as Types from "vscode-languageserver-types";
 
-const describe_opt = LanguageServer.ocamlVersionGEq("4.08.0") ? describe : xdescribe;
+const describe_opt = LanguageServer.ocamlVersionGEq("4.08.0")
+  ? describe
+  : xdescribe;
 
 describe_opt("textDocument/completion", () => {
   let languageServer = null;
--- a/ocaml-lsp-server/docs/ocamllsp/switchImplIntf-spec.md
+++ b/ocaml-lsp-server/docs/ocamllsp/switchImplIntf-spec.md
@@ -1,0 +1,34 @@
+#### Switch Implementation/Interface Request
+
+Switch Implementation/Interface Request is sent from client to server  to get
+URI(s) of the file(s) that the current file can switch to, e.g.,  if the user
+has "foo.ml" and "foo.mli" files, the client, who want to  switch from one to
+the other, sends this request.
+
+If there are one or more files, to which the currently open file can  switch to,
+exist in the same folder, then URIs of all those existing  files are returned.
+In case there is no file to switch to in that folder,  the most likely candidate
+for creation is returned, e.g., if a user wants  to switch from "foo.ml", but no
+files already exist in the project that  could be returned, a URI for "foo.mli"
+is returned.
+
+##### Client capability
+
+nothing that should be noted
+
+##### Server capability
+
+property name: `handleSwitchImplIntf`
+property type: `boolean`
+
+##### Request
+
+- method: `ocamllsp/switchImplIntf`
+- params: `DocumentUri` (see [`DocumentUri`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#uri) in LSP specification)
+
+##### Response
+
+- result: DocumentUri[] (non-empty)
+- error: code and message set in case an exception happens during the willSaveWaitUntil request.
+
+
--- a/ocaml-lsp-server/src/import.ml
+++ b/ocaml-lsp-server/src/import.ml
@@ -38,6 +38,7 @@ module CodeLens = CodeLens
 module DocumentHighlight = DocumentHighlight
 module DocumentHighlightKind = DocumentHighlightKind
 module DocumentSymbol = DocumentSymbol
+module DocumentUri = DocumentUri
 module SymbolInformation = SymbolInformation
 module CompletionItem = CompletionItem
 module CompletionList = CompletionList
--- a/ocaml-lsp-server/src/switch_impl_intf.ml
+++ b/ocaml-lsp-server/src/switch_impl_intf.ml
@@ -1,0 +1,61 @@
+open Import
+
+let capability = ("handleSwitchImplIntf", `Bool true)
+
+let meth = "ocamllsp/switchImplIntf"
+
+(** See the spec for 'ocamllsp/switchImplIntf' *)
+let switch (state : State.t) (param : DocumentUri.t) :
+    (Json.t, Jsonrpc.Response.Error.t) result =
+  let file_uri = Uri.t_of_yojson (`String param) in
+  let filepath = Uri.to_path file_uri in
+  let ml, mli, re, rei, mll, mly = ("ml", "mli", "re", "rei", "mll", "mly") in
+  let open Result.O in
+  let+ doc = Document_store.get state.store file_uri in
+  let extensions_to_switch_to =
+    match Document.syntax doc with
+    | Ocaml -> (
+      match Document.kind doc with
+      | Intf -> [ ml; mly; mll; re ]
+      | Impl -> [ mli; mly; mll; rei ] )
+    | Reason -> (
+      match Document.kind doc with
+      | Intf -> [ re; ml ]
+      | Impl -> [ rei; mli ] )
+    | Ocamllex -> [ mli; rei ]
+    | Menhir -> [ mli; rei ]
+  in
+  let path_without_extension = Filename.remove_extension filepath ^ "." in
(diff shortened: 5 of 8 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -511,6 +516,46 @@ let definition_query (state : State.t) uri position merlin_request =
...
+          loop (`String file_to_switch_to :: acc) rest
+    in
+    `List (loop [] exts |> List.rev)
+  in
+  let extensions_to_switch_to =
+    (* ordering of extensions to switch goes from most to least likely candidate *)
+    match Document.syntax doc with
+    | Ocaml -> (
+      match Document.kind doc with
+      | Intf -> [ ml; re; mly; mll ]
```
rgrinberg, line 548: `re` is unlikely that it shouldn't be offered.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -511,6 +516,46 @@ let definition_query (state : State.t) uri position merlin_request =
...
+      | Impl -> [ mli; mll; mly; rei ] )
```
rgrinberg, line 549: `rei`, `mly`, and `mll`.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -511,6 +516,46 @@ let definition_query (state : State.t) uri position merlin_request =
...
+    | Reason -> (
+      match Document.kind doc with
+      | Intf -> [ re; ml ]
+      | Impl -> [ rei; mli ] )
+    | Ocamllex -> [ ml; mli ]
```
rgrinberg, line 554: Should only be `mli`

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -511,6 +516,46 @@ let definition_query (state : State.t) uri position merlin_request =
...
+    | Menhir -> [ mli; ml ]
```
rgrinberg, line 555: `mli` only.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -511,6 +516,46 @@ let definition_query (state : State.t) uri position merlin_request =
...
+    @return a list of single filepath to switch to if it already exists or a
+    list of filepaths to create a file *)
+let switch_impl_intf (state : State.t) (filepath : string) :
+    (Json.t, Jsonrpc.Response.Error.t) result =
+  let uri = Uri.of_path filepath in
+  let open Result.O in
+  let+ doc = Document_store.get state.store uri in
+  let ml, mli, re, rei, mll, mly = ("ml", "mli", "re", "rei", "mll", "mly") in
+  let path_without_extension = Filename.remove_extension filepath ^ "." in
+  let find_switch (exts : string list) =
```
rgrinberg, line 531: This logic is slightly backwards. It should be something like this:

1) Get all the candidates
2) If there's only a single candidate, choose it
3) If there's more than one candidate, pick the one that already exists.

rgrinberg:
For mll, mly files it's possible to offer `.re` files as counterparts but I feel like this is something so rare that we should just not worry about it.

ocaml-lsp-server/src/ocaml_lsp_server.ml
ulugbekna: > Assuming the input file is grammar.mly, executing
>```
>        ocamlyacc options grammar.mly
>```
>produces OCaml code for a parser in the file grammar.ml, and its interface in file grammar.mli.

[source](https://caml.inria.fr/pub/docs/manual-ocaml/lexyacc.html#s%3Aocamllex-overview)

IIUC, menhir and ocamlyacc use the same extension, so I think it makes sense to keep both with priority to `mli`.

ocaml-lsp-server/src/ocaml_lsp_server.ml
ulugbekna: >Assuming the input file is lexer.mll, executing
>
 >       ocamllex lexer.mll
>produces OCaml code for a lexical analyzer in file `lexer.ml`.

[source](https://caml.inria.fr/pub/docs/manual-ocaml/lexyacc.html#s:ocamllex-overview)

Maybe should only be `ml`?

ocaml-lsp-server/src/ocaml_lsp_server.ml
ulugbekna: Why is it that we remove re from the list for ocaml intf files but keep rei for ocaml impl files?

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -634,6 +679,22 @@ let ocaml_on_request :
   | Client_request.TextDocumentOnTypeFormatting _ ->
     Fiber.return @@ Ok (None, state)
   | Client_request.SelectionRange req -> selection_range state req
+  | Client_request.UnknownRequest
+      { meth = "ocamllsp/didSwitchImplIntf"; params } ->
+    (* return a list of a single filepath to switch to if it exists or a list of
+       potential filepaths to create a file and to switch to *)
+    ( match params with
+    | Some (`List [ `String filepath; _ ]) ->
```
ulugbekna, line 687: We do this quite ugly and brittle pattern matching because we use `sendRequest` fn on the vscode extension side, which has the signature given below and takes a variable number of arguments (params to send). 

```typescript
(method) MessageConnection.sendRequest<unknown>(method: string, ...params: any[]): Promise<unknown> (+11 overloads)
```

So when the extension sends a single string param, the params sent have the form `["<string>", null]`. It would be great to get just a string on the ocaml-lsp side but I'm not sure how. 

cc @mnxn

ocaml-lsp-server/src/ocaml_lsp_server.ml
ulugbekna: Let's have the following logic: 

```
if file is <extension> file then
  (case 1) if only one candidate exists then open that file else 
  (case 2) if several candidates exist then open a dialog box for user to pick which to open (or open all? - useful) else
  (case 3) if no candidate exists then <discussed below>
```

For cases 1 and 2, the behavior for files is clear, let's discuss case 3 - no candidate exists.

ml: create mli file directly
mli: create ml file directly
re: create rei file directly
rei: create re file directly
mll: offer to run `ocamllex <given filepath>` or do nothing (?)
mly: offer to choose from menhir and ocamlyacc and run or do nothing (?)

Wdyt?

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: The `.ml` is *always* generated by ocamllex. Hence it's an error if both `foo.mll` and `foo.ml` exist in the same directory (dune will complain about duplicate targets). However, ocamllex will generate `foo.mli` *unless* `foo.mli` already exists in the source. Hence it's possible for the user to jump to the mli.

The same behavior goes for ocamlyacc and menhir but with the `.mly` extension.

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: How about this:

If no candidates exist, then we present all possibilities to the user. The user can then pick and save this file. Consequently, the next invocations will no longer need to prompt the user. That seems simple and also consistent for all file types.

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: Just an accidental omission by me. Cross language intf <-> impl jumping is fishy and we shouldn't suggest it to the user. However, I'd like to propose the following:

* If only `foo.ml` exists, jumping to intf will go to `foo.mli`
* If both `foo.ml` and `foo.rei` exists, then we'll respect the user's quirky setup and jump between these two.

ocaml-lsp-server/src/ocaml_lsp_server.ml
mnxn: I took a look at the generated JS, and it seems that it is passing the optional and unit parameters as undefined to `sendRequest`:
```js
sendRequest(client, "ocamllsp/didSwitchImplIntf", $$document.fileName, undefined, undefined)
```

I pushed a fix that handles the parameters correctly and has the following generated code:
```js
function sendRequest(t, meth, data, token, param) {
  if (token !== undefined) {
    return t.sendRequest(meth, data, token);
  } else {
    return t.sendRequest(meth, data);
  }
}
```

rgrinberg (approved):
Looks good. Please merge after adding some tests.

ulugbekna:
Not to bikeshed, but how do we like the command name `ocamllsp/didSwitchImplIntf`? I feel like `didSwitchImplIntf` is akin to an event that happened, which isn't true. Can we go with `ocamllsp/switchImplIntf`?

ulugbekna:
Otherwise, I feel like PR is *finally* ready

mnxn:
> Not to bikeshed, but how do we like the command name `ocamllsp/didSwitchImplIntf`? I feel like `didSwitchImplIntf` is akin to an event that happened, which isn't true. Can we go with `ocamllsp/switchImplIntf`?

When I was originally naming the methods, I followed the C++ extension, which was named `cpptools/didSwitchHeaderSource`: https://github.com/microsoft/vscode-cpptools/blob/9f8a5ad797051497cbd0b1d122c85c5d87a9c026/Extension/src/LanguageServer/client.ts#L410

I looked over the [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/specification-current/) and the methods that use `did` seem to be notifications. Renaming the methods to remove `did` seems like the right thing to do.

rgrinberg:
> Otherwise, I feel like PR is finally ready

The title still says its WIP 🤔

ulugbekna:
Yeah, sorry, decided to fix some things

ulugbekna:
Sorry for a "noisy" PR.

I think now it is *really* ready and can be merged (after a review preferably) :-)

rgrinberg:
Looks good. All I did is run `$ make fmt` and move all your code to a separate module. I'd like to minimize the amount of method specific logic in ocaml_lsp_server.

ocaml-lsp-server/test/e2e/__tests__/textDocument-completion.test.ts
```diff
@@ -3,7 +3,9 @@ import * as LanguageServer from "./../src/LanguageServer";
 
 import * as Types from "vscode-languageserver-types";
 
-const describe_opt = LanguageServer.ocamlVersionGEq("4.08.0") ? describe : xdescribe;
+const describe_opt = LanguageServer.ocamlVersionGEq("4.08.0")
+  ? describe
+  : xdescribe;
```
ulugbekna, line 8: seems a little bizarre formatting

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -652,9 +658,21 @@ let on_request :
...
   in
-  match syntax with
-  | Some (Ocamllex | Menhir) -> not_supported ()
-  | _ -> ocaml_on_request server req
+  match req with
+  | Client_request.UnknownRequest { meth; params } -> (
+    match
+      [ (Switch_impl_intf.meth, Switch_impl_intf.on_request) ]
+      |> List.assoc_opt meth
+    with
```
ulugbekna, line 666: I wonder if using extensible GADTs to represent requests and responses would result in more elegant handling of lsp implementation-specific requests.

ulugbekna (commented):
lgtm

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: Could be. But keep in mind that requests are server specific, so it's not completely correct to extend the request type in general. This would cause problems for those developing more than one server at a time or combining servers somehow.

rgrinberg:
@ulugbekna Are you able to reproduce this failure locally?

```
FAIL __tests__/ocamllsp-switchImplIntf.ts
14
  ● ocamllsp/switchImplIntf › test switches
15

16
    expect(received).toEqual(expected) // deep equality
17

18
    - Expected  - 1
19
    + Received  + 0
20

21
      Array [
22
        "file:///Users/runner/work/ocaml-lsp/ocaml-lsp/ocaml-lsp-server/test/e2e/test_files/test.ml",
23
    -   "file:///Users/runner/work/ocaml-lsp/ocaml-lsp/ocaml-lsp-server/test/e2e/test_files/test.mll",
24
      ]
25

26
      52 |   ) => {
27
      53 |     let response = await ocamllspSwitchImplIntf(requestParam);
28
    > 54 |     expect(response).toEqual(expectedResponse);
29
         |                      ^
30
      55 |   };
31
      56 | 
32
      57 |   /**
```

I'm unable to :/

rgrinberg:
Seems like the tests are flaky. We'll need to take care of that later in a different PR.

PR #236, opened 2020-08-27
Update to latest merlin & update deps

rgrinberg:
* We now support 4.11
* Unvendor csexp. There's no need to vendor it as it offers a stable API


smorimoto:
Would you like to add a test for 4.11.0 in another PR?

rgrinberg:
Thanks for the reminder. Let me just add it here.

PR #253, opened 2020-10-02
Fix completion bug #252

### LINKED ISSUE #252: Record accessor functions that don't exist are offered as completions
As shown in the screenshot, the first three autocompletion items offered look like accessor functions to pull values out of a record, but after accepting such a completion, we're told that it doesn't exist (both from realtime LSP diagnostics, and an error if running the compiler).

Screenshot 2020-09-30 at 13 12 09

`dune`:
```text
(executable
 (name main))
```

`main.ml`
```ocaml
module M = struct
  type t =
    { x : int
    ; y : float
    ; z : string }
end

let r : M.t = {x = 1; y = 2.2; z = "hello"}
let _ = M.y r
```

compiler output:
```text
File "main.ml", line 9, characters 8-11:
9 | let _ = M.y r
            ^^^
Error: Unbound value M.y
```

---

`opam list --installed-roots`:
```text
# Packages matching: root
# Name              # Installed # Synopsis
dune                2.7.1       Fast, portable, and opinionated build system
ocaml-base-compiler 4.11.1      Official release 4.11.1
ocaml-lsp-server    ~dev        pinned to version ~dev at git+https://github.com/ocaml/
```

ulugbekna:
Fixes #252

### DIFF
--- a/ocaml-lsp-server/src/compl.ml
+++ b/ocaml-lsp-server/src/compl.ml
@@ -135,9 +135,6 @@ let item index full_entry ~compl_info =
     ~sortText:(Printf.sprintf "%04d" index)
     ~data:compl_info ?textEdit ()
 
-let completion_kinds =
-  [ `Constructor; `Labels; `Modules; `Modules_type; `Types; `Values; `Variants ]
-
 let complete doc lsp_position =
   let position = Position.logical lsp_position in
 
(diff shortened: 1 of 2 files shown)


rgrinberg:
What's the problem with the switchImplIntf tests currently?

ulugbekna:
You mentioned they are flaky [here](https://github.com/ocaml/ocaml-lsp/pull/233). I think they fail sometimes without an apparent reason, especially on Linux. 

I can enable them in another PR and trigger the CI a couple of times to see if the problem persists.

ulugbekna:
And they anyway need to be updated in the light of https://github.com/ocamllabs/vscode-ocaml-platform/issues/382

rgrinberg:
Yeah, but disabling them doesn't seem like the way to go. If they are failing with the same flaky result, but passing locally, just merge anyway.

rgrinberg (approved):
Looks good. This complicates the implementation of the keyword completion I had in mind, but we'll think of another way of doing it.

PR #254, opened 2020-10-02
fix bug in 'switchImplIntf':

ulugbekna:
Fixes https://github.com/ocamllabs/vscode-ocaml-platform/issues/382:
 
  one couldn't switch from a new file that's not saved on disk
  (in vscode it's a file with URI scheme 'untitled')

  The previous switching logic depended on document store, which only
  knows files that were opened with a 'textDocument/didOpen'
  notification. VS Code doesn't send such notifications for unsaved files,
  hence the bug. Now switching handled file URIs directly without
  dependence on document store, which works for any 'switchImplIntf'
  request with a valid URI.

Note: handling of URIs is brittle as is the current URI module. I would strongly be in favor of adding a vendored dep `ocaml-uri`.

TODO:
- [x] add proper error handling using `result`
- [x] add tests for 'untitled' scheme (without loss of generality for other schemes that follow URI conventions of `scheme:/path`)

### DIFF
--- a/ocaml-lsp-server/src/switch_impl_intf.ml
+++ b/ocaml-lsp-server/src/switch_impl_intf.ml
@@ -4,53 +4,50 @@ let capability = ("handleSwitchImplIntf", `Bool true)
 
 let meth = "ocamllsp/switchImplIntf"
 
-(** See the spec for 'ocamllsp/switchImplIntf' *)
+(** see the spec for [ocamllsp/switchImplIntf] *)
 let switch (param : DocumentUri.t) : (Json.t, Jsonrpc.Response.Error.t) result =
-  let filepath =
-    let file_uri = Uri.t_of_yojson (`String param) in
-    let possible_filepath = Uri.to_path file_uri in
-    if possible_filepath = Uri.to_string file_uri then
-      (* remove URI scheme *)
-      String.split ~on:':' possible_filepath |> List.tl |> List.hd
-    else
-      possible_filepath
-    (* Note: URI lib such as ocaml-uri would help eliminate brittle URI handling *)
+  let fpath =
+    match String.split ~on:':' param with
+    | [ scheme; path ] ->
+      if scheme = "file" then
+        Uri.t_of_yojson (`String param) |> Uri.to_path
+      else
+        path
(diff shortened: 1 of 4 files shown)


ocaml-lsp-server/src/switch_impl_intf.ml
```diff
@@ -5,21 +5,27 @@ let capability = ("handleSwitchImplIntf", `Bool true)
...
-    (Json.t, Jsonrpc.Response.Error.t) result =
-  let file_uri = Uri.t_of_yojson (`String param) in
-  let filepath = Uri.to_path file_uri in
+let switch (param : DocumentUri.t) : (Json.t, Jsonrpc.Response.Error.t) result =
+  let filepath =
+    let file_uri = Uri.t_of_yojson (`String param) in
+    let possible_filepath = Uri.to_path file_uri in
+    if possible_filepath = Uri.to_string file_uri then
+      (* remove URI scheme *)
+      String.split ~on:':' possible_filepath |> List.tl |> List.hd
```
rgrinberg, line 14: How about just:
```
match String.split with
| _ :: x :: _ - >x
| _ -> assert false
```

rgrinberg:
>  VS Code doesn't send such notifications for unsaved files,
hence the bug.

Does that mean there's no lsp assistance until the file is saved?

ulugbekna:
> > VS Code doesn't send such notifications for unsaved files,
> > hence the bug.
> 
> Does that mean there's no lsp assistance until the file is saved?

Yes; extension understands that's an OCaml file but we miss ocamllsp support for that file. We could artificially send didOpen notif when that unsaved file is created, but we'd have to handle cases: what happens when it's saved, what happens when it's closed without saving, can merlin handle files with non file URIs, etc.

rgrinberg:
> Yes; extension understands that's an OCaml file but we miss ocamllsp support for that file. We could artificially send didOpen notif when that unsaved file is created, but we'd have to handle cases: what happens when it's saved, what happens when it's closed without saving, can merlin handle files with non file URIs, etc.

Nah, there's no need. I was just curious if this was indeed the case. It seems like a flaw in LSP itself and we shouldn't bother patching over it.

PR #255, opened 2020-10-06
Disable destruct in intf files

rgrinberg:
This action does not make sense in interface files


ulugbekna:
We could pass `state` instead of `rpc` to `code_action` as in other request handlers such as `hover`, etc. Otherwise, lgtm.

rgrinberg:
Good call. I'll do that once I move destruct to its own module

PR #262, opened 2020-10-19
fix windows URI parsing bug (mentioned in #184 at the end)

ulugbekna:
I hope I got this right this time. 

It'd be great if someone could test it on windows in particular (@db4 or @smorimoto maybe?)

### DIFF
--- a/ocaml-lsp-server/src/switch_impl_intf.ml
+++ b/ocaml-lsp-server/src/switch_impl_intf.ml
@@ -15,11 +15,13 @@ let scheme_path_of_uri s =
     (* handles URI: scheme:path as in untitled:/path/to/test.ml or
        untitled:/c:/test.ml *)
   in
-  let scheme_path_of_uri () = (Str.matched_group 1 s, Str.matched_group 2 s) in
+  let scheme_path_of_uri () =
+    try Some (Str.matched_group 1 s, Str.matched_group 2 s) with _ -> None
+  in
   if Str.string_match uri_regex_1 s 0 then
-    Some (scheme_path_of_uri ())
+    scheme_path_of_uri ()
   else if Str.string_match uri_regex_2 s 0 then
-    Some (scheme_path_of_uri ())
+    scheme_path_of_uri ()
   else
     None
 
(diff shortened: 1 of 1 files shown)


rgrinberg:
@mnxn has a windows machine as well

mnxn:
If I switch from an existing `a.ml` to an untitled scheme `a.mli` and try to switch back, it doesn't work.

Switching to `a.mli`:
```
[Trace - 8:52:03 AM] Sending request 'ocamllsp/switchImplIntf - (4)'.
Params: "file:///c:/dev/projects/promise_jsoo/a.ml"


[Trace - 8:52:03 AM] Received response 'ocamllsp/switchImplIntf - (4)' in 1ms.
Result: [
    "file:///c%3A/dev/projects/promise_jsoo/a.mli"
]
```

Trying to switch back to `a.ml`:
```
[Trace - 8:52:10 AM] Sending request 'ocamllsp/switchImplIntf - (5)'.
Params: "untitled:/c:/dev/projects/promise_jsoo/a.mli"


[Trace - 8:52:10 AM] Received response 'ocamllsp/switchImplIntf - (5)' in 1ms.
Result: [
    "file:////c%3A/dev/projects/promise_jsoo/a.ml"
]
```

I think the problem is the 4 forward slashes in `file:////c%3A/dev/projects/promise_jsoo/a.ml`.

The developer console has this error:
```
[UriError]: If a URI does not contain an authority component, then the path cannot begin with two slash characters ("//")
```

ulugbekna:
1) I don't have a windows machine and installing windows virtual box doesn't allow to have WSL to install opam et al., does anybody know what I could do? 

2) I used `Str` module here. Should I replace it with something else because of thread-safety?

rgrinberg:
Yes, please drop it. Consider one of:

* Do not use regular expressions at all if its simple enough.
* Use ocamllex
* Use re if regular expressions cannot be avoided.

mnxn:
>  1. I don't have a windows machine and installing windows virtual box doesn't allow to have WSL to install opam et al., does anybody know what I could do?

@ulugbekna: I don't think installing WSL will help test for Windows since it will act like linux.

You can install [OCaml for Windows](https://fdopen.github.io/opam-repository-mingw/installation/) for a native mingw Windows version. It's what Esy uses under the hood on Windows.

PR #264, opened 2020-10-21
Fix mutex change of ownership (OpenBSD)

xguerin:
OpenBSD is pretty strict about change of ownership for mutexes (`abort()`). This PR addresses that issue by moving mutex locking to their logically owning threads.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 - Do not use vendored libraries when building the lsp package (#260)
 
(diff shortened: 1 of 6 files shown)


xguerin:
AFAICT the cause of the failure on Mac is not related to the patch.

rgrinberg:
Thanks. A reference about mutexes on openbsd would be handy :) I'll take your word for it though

PR #260, opened 2020-10-17
delete vendor directory when installing lsp

rgrinberg:
Fix #259

PR #265, opened 2020-10-21
Vendor directory fixes

Gbury:
Follow-up to #260 
For now, there are two main changes:
- commit the opam file generated from the template
- add the unix dependency for the `unvendor.ml` script

The installation of a pinned version now fails at `dune subst`, with the error:
```
Error: vendor/cmdliner: No such file or directory
```
However I don't know enough about dune to know how to fix that, ^^

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,24 +1,24 @@
 [submodule "vendor/merlin"]
-	path = vendor/merlin
+	path = ocaml-lsp-server/vendor/merlin
 	url = https://github.com/rgrinberg/merlin
 	branch = lsp
 	ignore = dirty
 [submodule "vendor/cmdliner"]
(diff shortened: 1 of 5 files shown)


rgrinberg:
@Gbury give it a try now.

Gbury:
@rgrinberg I'm still getting the same error:
```
Error: ocaml-lsp-server/vendor/cmdliner: No such file or directory
```
It looks like the unvendor script is correctly removing the vendored directories, but `dune subst` tries and either fetch info or modify some things in these directories, and fail because they do not exist anymore.

rgrinberg:
Yeah, this seems like a dune subst bug that I will report upstream.

I disabled dune subst for lsp for now.

PR #280, opened 2020-10-27
Move fiber scheduler to own library

rgrinberg:
The goal is to make the lsp library useful without fibers. It also makes
it easier to move fiber tests in the correct place.


fiber-unix/src/dune
```diff
@@ -0,0 +1,5 @@
+(library
+ (name fiber_unix)
+ (public_name lsp.fiber_unix)
```
rgrinberg, line 3: This is required until we get rid of the dependency on fibers.

PR #283, opened 2020-10-29
Split lsp into sub libraries

rgrinberg:
* jsonrpc_fiber - fiber based implementation of jsonrpc
* lsp_fiber - fiber based implementation of lsp


rgrinberg:
This introduces a little bit of duplication, but it's necessary to make the lsp library more widely useful.

PR #286, opened 2020-10-30
Create documents in merlin thread

rgrinberg:
It does not seem to be safe to do these operations concurrently to
merlin


rgrinberg:
@ulugbekna will wait for explicit approval this time. Sorry for the impatience.

ulugbekna (approved):
Looks good to me. 

Rebased and made some renamings. Feel free to remove my commit if you don't agree with it -- thought it's better to make changes myself than bothering you.

PR #288, opened 2020-11-01
Simplify prefix_of_position

rgrinberg:
Do the entire operation in a single pass with the help of String.rfrom

### DIFF
--- a/ocaml-lsp-server/src/compl.ml
+++ b/ocaml-lsp-server/src/compl.ml
@@ -96,10 +96,10 @@ let suffix_of_position source position =
   | "" -> ""
   | text ->
     let len = String.length text in
-
-    let rec find suffix i =
+    let buf = Buffer.create 8 in
+    let rec find i =
       if i >= len then
-        suffix
+        ()
       else
         let ch = text.[i] in
         (* The characters for an infix function are missing *)
@@ -109,12 +109,14 @@ let suffix_of_position source position =
         | '0' .. '9'
         | '\''
         | '_' ->
-          find (ch :: suffix) (i + 1)
-        | _ -> suffix
+          Buffer.add_char buf ch;
+          find (i + 1)
+        | _ -> ()
     in
 
     let (`Offset index) = Msource.get_offset source position in
-    make_string (List.rev @@ find [] index)
+    find index;
+    Buffer.contents buf
 
 let range_prefix (lsp_position : Position.t) prefix : Range.t =
   let start =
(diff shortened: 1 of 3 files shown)


vendor/stdune/string.ml
```diff
@@ -271,7 +282,18 @@ let rfindi =
...
   in
-  fun s ~f -> loop s ~f (String.length s - 1)
+  fun ?from s ~f ->
+    let from =
+      let len = String.length s in
+      match from with
+      | None -> len - 1
+      | Some i ->
+        if i > len - 1 then
+          Code_error.raise "rfind: invalid from" []
```
ulugbekna, line 292: `rfind` -> `rfindi` ?

ocaml-lsp-server/src/compl.ml
```diff
@@ -79,42 +71,42 @@ let prefix_of_position ~short_path source position =
...
+    let len =
+      let until =
+        String.findi ~from text ~f:(function
+          | 'a' .. 'z'
+          | 'A' .. 'Z'
+          | '0' .. '9'
+          | '\''
+          | '_' ->
+            false
+          | _ -> true)
```
ulugbekna, line 104: I think it would be very helpful in reading the code if we moved predicates out in a separate function something like `is_valid_ident` (in our case it's more like `not_is_valid_ident`) here and in `prefix_of_position`.

ocaml-lsp-server/src/compl.ml
```diff
@@ -79,42 +71,42 @@ let prefix_of_position ~short_path source position =
...
+        | '.' -> short_path
+        | _ -> true)
     in
-
-    let (`Offset index) = Msource.get_offset source position in
-    find [] (index - 1)
+    let pos =
+      match pos with
+      | None -> 0
+      | Some pos -> pos + 1
```
ulugbekna, line 81: is this why we stopped supporting non-prefix completions?

ulugbekna (approved):
Index manipulations are a little tricky to review, but LGTM.

ocaml-lsp-server/src/compl.ml
rgrinberg: You mean the non-prefix completion I've referred to in slack? No. I removed non-prefix completion because it was slow and unreliable.

ocaml-lsp-server/src/compl.ml
rgrinberg: That's true. I did that.

vendor/stdune/string.ml
rgrinberg: good catch.

PR #292, opened 2020-11-02
Fix the dune files to fix compilation errors in 4.06.x

smorimoto:
To make it easier to compare to each dune file, I ordered all the libraries in alphabetical order, but that breaks the diff for this commit. For other things I did, I just added the `Result` library and `future_syntax` preprocess as needed. If you don't like it, I will undo the changes around the alphabetical sort.

### DIFF
--- a/fiber-test/dune
+++ b/fiber-test/dune
@@ -1,3 +1,3 @@
 (library
  (name fiber_test)
- (libraries stdune fiber))
+ (libraries fiber stdune))
--- a/fiber-unix/src/dune
+++ b/fiber-unix/src/dune
@@ -1,5 +1,5 @@
 (library
  (name fiber_unix)
  (public_name lsp.fiber_unix)
- (libraries stdune fiber threads.posix yojson)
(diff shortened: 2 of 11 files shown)


rgrinberg:
Seems good. Is there something wrong with our CI job for 4.06? It should have broke if we dropped 4.06 accidentally.

smorimoto:
In 4.06.x, steps such as installing dependencies are enabled, but building is disabled.

smorimoto:
If necessary, I will open a PR to address it in another PR.

rgrinberg:
Creating a PR that will enable building in 4.06 would be quite welcome. I think it would also be helpful to merge it before merging this PR. This will allow us to verify that this PR actually fixes the issues with 4.06.

smorimoto:
Sure, that's fine with me, but my friend is stuck with this and wants to merge this to the master as soon as possible. What do you think? I checked that there is no problem with this in my local environment.

rgrinberg:
Sure, I don't mind.

smorimoto:
Thanks.

PR #294, opened 2020-11-03
Implement TextDocumentDeclaration

mnxn:
Implements TextDocumentDeclaration in order to use "Go to Declaration" in editors like VSCode. This is useful for jumping to .mli files in order to read documentation.


rgrinberg:
Thanks. Could you rebase? The commit on master fixes the make target.

rgrinberg (approved):

mnxn:
I still need to add tests. They will likely be a bit more complicated than TextDocumentDefinition because it needs Merlin to recognized two files.

rgrinberg:
> I still need to add tests. They will likely be a bit more complicated than TextDocumentDefinition because it needs Merlin to recognized two files.

Indeed. You will need to use dune to build your two sources.

I'm wondering if this is where we should start experimenting with cram tests? Perhaps you could could create a little test binary:

```
  $ lsp_run_server_and_send_request.exe < cat <<EOF
  {
    // your locate query
  }
  > EOF
```

I don't really like the current typescript tests. Quite a bit of overhead and the integration with dune is poor.

mnxn:
I'd like to keep this PR small in scope, so I just did it in typescript with some supplemental files. We can move to cram tests in another PR.

PR #295, opened 2020-11-03
Use proper barrier for timer awaker

rgrinberg:
The previous implementation relied on Thread.delay which was not
interruptible once the awaker was already asleep and we were waiting to
write into the mvar.

Barrier is a poor man's implementation of `pthread_cond_wait`, which I tried to
avoid binding to. @mnxn could you check that this work on windows?

### DIFF
--- a/fiber-unix/src/barrier.ml
+++ b/fiber-unix/src/barrier.ml
@@ -6,40 +6,26 @@ type t =
 
 let create () =
   let r, w = Unix.pipe () in
-  Unix.set_nonblock r;
-  { r; w; buf = Bytes.create 8 }
+  { r; w; buf = Bytes.create 1 }
 
 let close t =
   Unix.close t.w;
   Unix.close t.r
 
-let read_one t =
-  match Unix.read t.r t.buf 0 (Bytes.length t.buf) with
-  | 0 -> Error `No_data
-  | _ -> Ok ()
-  | exception Unix.Unix_error (Unix.EAGAIN, _, _) -> Error `Unavailable
-  | exception Unix.Unix_error (Unix.EBADF, _, _) -> Error `Bad_fd
-  | exception Unix.Unix_error (e, _, _) -> failwith (Unix.error_message e)
+let rec empty_fd fd buf =
+  match Unix.select [ fd ] [] [] 0.0 with
+  | [], _, _ -> Ok ()
+  | _ ->
+    let _ = Unix.read fd buf 0 1 in
+    empty_fd fd buf
+  | exception Unix.Unix_error (Unix.EBADF, _, _) -> Error (`Closed (`Read true))
+  | exception Unix.Unix_error (e, _, _) ->
+    failwith ("read :" ^ Unix.error_message e)
 
-let read_all t =
-  let rec loop read_once =
-    match read_one t with
-    | Ok () -> loop true
-    | Error `Bad_fd -> Error (`Bad_fd, read_once)
-    | Error `Unavailable
-    | Error `No_data ->
-      Ok read_once
-  in
-  loop false
-
-let rec await ?(timeout = -1.) t =
+let await ?(timeout = -1.) t =
   match Unix.select [ t.r ] [] [] timeout with
   | [], _, _ -> Error `Timeout
-  | _ -> (
-    match read_all t with
-    | Ok true -> Ok ()
-    | Ok false -> await ~timeout t
-    | Error (`Bad_fd, read_once) -> Error (`Closed (`Read read_once)) )
+  | _ -> empty_fd t.r t.buf
   | exception Unix.Unix_error (Unix.EBADF, _, _) ->
     Error (`Closed (`Read false))
   | exception Unix.Unix_error (e, _, _) ->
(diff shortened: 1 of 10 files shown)


mnxn:
Definitely doesn't work on Windows:

```
ocamllsp: internal error, uncaught exception:
          Unix.Unix_error(Unix.ENOTSOCK, "unix_set_nonblock", "")
          Raised by primitive operation at Fiber_unix__Barrier.create in file "fiber-unix/src/barrier.ml", line 9, characters 2-21
          Called from Fiber_unix__Scheduler.create in file "fiber-unix/src/scheduler.ml", line 234, characters 30-47
          Called from Dune__exe__Ocaml_lsp_server.start in file "ocaml-lsp-server/src/ocaml_lsp_server.ml", line 750, characters 18-37
          Called from Lsp__Logger.with_sections in file "lsp/src/logger.ml", line 155, characters 8-12
          Re-raised at Lsp__Logger.with_sections in file "lsp/src/logger.ml", line 161, characters 4-19
          Called from Lsp__Logger.with_log_file in file "lsp/src/logger.ml", line 187, characters 10-34
          Re-raised at Lsp__Logger.with_log_file in file "lsp/src/logger.ml", line 193, characters 6-21
          Called from Cmdliner_term.app.(fun) in file "ocaml-lsp-server/vendor/cmdliner/src/cmdliner_term.ml", line 25, characters 19-24
          Called from Cmdliner.Term.run in file "ocaml-lsp-server/vendor/cmdliner/src/cmdliner.ml", line 117, characters 32-39
```

tmattio:
@rgrinberg There are bindings for `pthread_cond_wait` in `Stdlib.Condition`, which has an API similar to the `Barrier` module in this PR. But reading your code, it seems you need `pthread_cond_timedwait`, not `pthread_cond_wait`, right?

rgrinberg:
Indeed. I want to wait with a timeout

tmattio:
> Indeed. I want to wait with a timeout

Ok, are you open to doing this with bindings if I implement the code for Windows? I'm preparing a PR for the compiler and could use the same implementation

rgrinberg:
Sure we can replace this implementation eventually if pr gets merged. I strongly prefer not to add any C in this pr

tmattio:
Seems like on Windows, `Unix.pipe ()` calls `CreatePipe`, which creates anonymous pipes, which don't support async IO. So there's no way to use `Unix.set_nonblock` on the file descriptors returned by `Unix.pipe ()`.
Reading Microsoft doc on async IO (https://docs.microsoft.com/en-us/windows/win32/ipc/synchronous-and-overlapped-input-and-output), there would be another solution using `ReadFile` and `WriteFile`, but the OCaml bindings don't allow passing the correct arguments to enable async IO on the file descriptors.

I see two solutions:
- Make the read operations on the fd blocking on Windows (I tried doing this and did not notice any performance hit)
- Bind to `pthread_cond_wait`

I'm not very familiar with Windows' API, so maybe there's a solution I'm not seeing though.

rgrinberg:
@tmattio try running the tests now.

tmattio:
It hangs indefinitely. I can investigate why, but at this point, I really think some c stubs might be cleaner than introducing this behavior (blocking reads) on Windows. Will keep going down the rabbit hole though if you think that's better.

rgrinberg:
First let's understand why OCaml's `select` emulation for pipes is not working for us here. I'm considering to use pipes for other thing, so it would be good to know the limitations.

As for introducing C: I think it would be simpler to just busy poll on windows at 10Hz. The timer accuracy isn't all that important for us.

tmattio:
@rgrinberg I implemented a non blocking reading in 7b87071779cac72478934f292ef225ea31510d2f.

It creates a separate thread that will read from the fd and set the state to `Available` whenever it receives data.

When calling `Barriere.await`, we read this state and if set it to `Empty` if it was available, otherwise, we poll the value of the state at 30Hz, until we reach the timeout, or the state is evaluated as `Available`

If this implementation is Ok for you, I'll make it specific to Windows and keep the `Unix.set_nonblock` on Unix, that's probably cleaner.

rgrinberg:
> otherwise, we poll the value of the state at 30Hz, until we reach the timeout, or the state is evaluated as Available

Why not use select?

tmattio:
> Why not use select?

We use polling to read the state, not the file descriptor. Reading the file descriptor is blocking, so we don't need polling.

rgrinberg:
> We use polling to read the state, not the file descriptor. Reading the file descriptor is blocking, so we don't need polling.

That's exactly what I don't understand. Why should we poll if we can use select?

tmattio:
> That's exactly what I don't understand. Why should we poll if we can use select?

How can we use `select` to read the state?

We have two threads:
- The parent, which does not use the file descriptor, it only reads the state, so to wait for the state to be `Available`, I don't see how we could do without polling.
The child, which reads from the file descriptor and set the state. But reading is blocking anyway, so `select` wouldn't give us anything.

rgrinberg:
Looks good. @mnxn do you mind testing this PR out again?

mnxn:
> Looks good. @mnxn do you mind testing this PR out again?

Just tried it on Windows and it seems to work fine. No crashes this time.

PR #296, opened 2020-11-04
Improve Actions workflow

smorimoto:
- Bump patch versions of OCaml compilers
- Bump Node.js version
- Run the actual build on all versions and platforms
- Use `runner.os` instead of `env.OS` to make version-independent condition
- Unify workflow style with others

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -17,12 +17,12 @@ jobs:
         skip_test:
           - false
         ocaml-version:
-          - 4.11.0
+          - 4.11.1
         include:
-          - ocaml-version: 4.11.0
+          - ocaml-version: 4.11.1
             os: windows-latest
             skip_test: true
-          - ocaml-version: 4.10.0
+          - ocaml-version: 4.10.1
             os: windows-latest
             skip_test: true
           - ocaml-version: 4.09.1
@@ -34,9 +34,6 @@ jobs:
 
     runs-on: ${{ matrix.os }}
 
-    env:
-      SKIP_TEST: ${{ matrix.skip_test }}
-
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -55,18 +55,15 @@ jobs:
...
+      - run: opam exec -- make build
 
       - run: opam exec -- make test
         if: env.SKIP_TEST != 'true'
 
-      - name: test source is well formatted
-        run: opam exec -- make fmt
-        if: env.OCAML_VERSION == '4.10.0' && env.OS == 'ubuntu-latest'
+      - run: opam exec -- make fmt
+        if: env.OCAML_VERSION == '4.10.0' && runner.os == 'Linux'
```
rgrinberg, line 66: Let's switch this to 4.11 as well

rgrinberg:
Good work. Thanks. Seems like 4.06 is still failing?

Are we running the tests on windows?

.github/workflows/build-and-test.yml
smorimoto: Oh indeed.

smorimoto:
I think it was fixed with a new commit.

No, we are not currently running tests on Windows.

ocaml-lsp-server.opam.template
```diff
@@ -1,3 +1,16 @@
...
+  "dune-build-info"
+  "dune" {>= "2.5.0"}
+  "menhir"
+  "ocamlfind" {>= "1.5.2"}
+  "ocamlformat" {with-test}
+  "ppx_yojson_conv_lib"
+  "result" {>= "1.5"}
+  "stdlib-shims"
+  "yojson"
+  ("ocaml" {>= "4.06" & < "4.12"} | ("ocaml" {< "4.08"} & "ocamlfind-secondary"))
```
smorimoto, line 12: If you don't want to move everything to a template, we can install the `ocamlfind-secondary` only in the Actions runner.

smorimoto: This is just affected by the setup-ocaml spec, so I decided to handle it only in the Actions runner at the moment.

tmattio:
Closing in favor of #318 (which builds on this PR)

PR #298, opened 2020-11-04
Ignore all changes to the submodule uutf's work tree

smorimoto:


smorimoto:
Oh, the VSCode default autoformat feature is causing unnecessary diff... If you don't like it I will undo it and only commit the minimal changes.

ulugbekna (approved):

Issue #291, opened 2020-11-01
Certain completion requests cause the language server to crash


mnxn:
If I type the following in small dune project, the server crashes immediately after I type `f`:

```ocaml
let _ = Base.f
```

stack trace:
```
ocamllsp: internal error, uncaught exception:
          File "ocaml-lsp-server/vendor/merlin/src/ocaml/utils/411/load_path.ml", line 83, characters 2-8: Assertion failed
          Raised at Load_path.find_uncap in file "ocaml-lsp-server/vendor/merlin/src/ocaml/utils/411/load_path.ml", line 83, characters 2-46
          Called from Persistent_env.Persistent_signature.load.(fun) in file "ocaml-lsp-server/vendor/merlin/src/ocaml/typing/411/persistent_env.ml", line 42, characters 12-53
          Called from Persistent_env.find_pers_struct in file "ocaml-lsp-server/vendor/merlin/src/ocaml/typing/411/persistent_env.ml", line 253, characters 16-58
          Called from Persistent_env.find in file "ocaml-lsp-server/vendor/merlin/src/ocaml/typing/411/persistent_env.ml" (inlined), line 300, characters 6-45
          Called from Env.find_pers_mod in file "ocaml-lsp-server/vendor/merlin/src/ocaml/typing/411/env.ml", line 824, characters 2-86
          Called from Env.lookup_ident_module in file "ocaml-lsp-server/vendor/merlin/src/ocaml/typing/411/env.ml", line 2457, characters 16-31
          Called from Env.lookup_module_components in file "ocaml-lsp-server/vendor/merlin/src/ocaml/typing/411/env.ml", line 2538, characters 23-71
          Called from Env.find_all in file "ocaml-lsp-server/vendor/merlin/src/ocaml/typing/411/env.ml", line 2929, characters 8-92
          Called from Env.spellcheck.choices in file "ocaml-lsp-server/vendor/merlin/src/ocaml/typing/411/env.ml" (inlined), line 3143, characters 43-61
          Called from Env.spellcheck.(fun) in file "ocaml-lsp-server/vendor/merlin/src/ocaml/typing/411/env.ml", line 3149, characters 40-64
          Called from Misc.did_you_mean in file "ocaml-lsp-server/vendor/merlin/src/ocaml/utils/misc.ml", line 595, characters 8-22
[...]

PR #300, opened 2020-11-06
Remove uses of str_formatter

rgrinberg:
It's not thread safe and offers no convenience to make up for it.


rgrinberg:
@ulugbekna @mnxn I pushed a commit that fixes #291

mnxn (approved):
> I pushed a commit that fixes #291

Can confirm. The language server no longer crashes for me.

PR #305, opened 2020-11-10
Fix switch impl intf tests

ulugbekna:


ulugbekna:
go-to-declaration tests are failing; I don't think the PR has any changes that could cause that.

mnxn:
It doesn't seem to fail when running the tests locally for me. But it does when I do `dune build` in the `declaration_files` directory.

Maybe the test project is misconfigured. @rgrinberg: do you know what would cause this dune error?
> Command failed: dune build
> Entering directory '/Users/runner/work/ocaml-lsp/ocaml-lsp'
> Error: Don't know about directory
> ocaml-lsp-server/test/e2e/\_\_tests\_\_/declaration_files specified on the
> command line!

ulugbekna:
`ocaml-lsp` occurs twice in path, but it shouldn't, I think.

rgrinberg:
Sorry guys, I have no idea and I'm not interested in investigating. The tests pass on my laptop, so I suppose we can merge this.

Let's focus on moving the tests to OCaml. That should fix these issues for good.

smorimoto:
Node.js exec basically has a lot more unusual behavior than most people think. ​Honestly, I think it makes the most sense to rewrite all the tests to OCaml.

Issue #173, opened 2020-05-09
OCaml keywords are not suggested.


mnxn:
https://github.com/ocamllabs/vscode-ocaml-platform/issues/183

ocaml-lsp should suggest keywords like language servers for other programming languages. It's particularly annoying when writing `let ... in` expressions as pressing enter at the end of the line completes `in_channel` most of the time.

![image](https://user-images.githubusercontent.com/25037249/81464412-73038380-9176-11ea-8864-eda5326c2418.png)

`Keyword` is already a supported `CompletionItemKind`: https://github.com/ocaml/ocaml-lsp/blob/f3e1f4c543eef14a3cc42bc5efbb41c9137dc3a2/lsp/src/types.mli#L549-L575

I'd imagine that we could maintain a list of keywords for completion, possibly limiting the suggestions based on whether a `.ml` or `.mli` file is open.

rgrinberg:
Yes, this would be nice. I have a WIP PR to add keyword completions to merlin. I should probably revive it soon

giltho:
When finishing a line with `in`, it has become a reflex to type `space -> backspace -> enter` to avoid this...
This would be great :)

Kakadu:
It would be great to add this to 
https://github.com/ocaml/ocaml-lsp/milestone/1

rgrinberg:
Everyone who thumbed up this PR, please give master a test.

Kakadu:
> Everyone who thumbed up this PR, please give master a test.

At the moment I can't do that because of https://github.com/ocaml/ocaml-lsp/issues/291#issuecomment-721293389

yannl35133:
Not sure how that issue affects testing for this test, but I quickly conducted mine and the main complaint (`in` suggestion'ed into `in_channel`) is gone, so fixed for me

Kakadu:
> Not sure how that issue affects testing for this test, but I quickly conducted mine and the main complaint (`in` suggestion'ed into `in_channel`) is gone, so fixed for me

I had issues with invoking completion...

rgrinberg:
The issue has been fixed. Please try again

Kakadu:
It's better now. Thanks!

rgrinberg:
>  I'd imagine that we could maintain a list of keywords for completion, possibly limiting the suggestions based on whether a .ml or .mli file is open.

[...]

PR #307, opened 2020-11-17
Filter keyword completion by completion context

mnxn:
Implements the keyword filtering mentioned in https://github.com/ocaml/ocaml-lsp/issues/173#issuecomment-723665086

Keywords that are only used in implementations are filtered out of the possible keywords for "interface" completion contexts (`Module_type`, `Signature`, `Type`)


rgrinberg:
Looks good. Could you make also make a PR to my fork [here] at the lsp branch?

rgrinberg:
Merged. Thanks

PR #257, opened 2020-10-07
Infer interface action

rgrinberg:
- [ ] Handle unopened documented
- [ ] Wrap printing environment
- [ ] Get URI for impl doc
- [ ] Use end location instead of start location
- [ ] Add tests

Issue #247, opened 2020-09-11
Reason & 4.11


rgrinberg:
Reason gets in the way of easily testing ocaml-lsp and the latest version of OCaml.
We should either disable the tests or remove reason from the test suite altogether.

EduardoRFS:
Reason is working with OCaml 4.11 since day one, and we always have a version working with the new compile before release. If you want to pin the opam version

> https://github.com/facebook/reason/blob/master/reason.opam

rgrinberg:
Pinning is incredibly inconvenient and makes it painful to release ocaml-lsp. The released version in opam is still incompatible with 4.11

rgrinberg:
#287 fixes this issue

PR #308, opened 2020-11-17
Add inferred_intf code action

tmattio:
This PR builds on top of #257 to provide a code action to infer a module interface.

It implements the following todos from the previous PR:

- [x] Handle unopened documented
- [x] Wrap printing environment
- [X] Get URI for impl doc
- [x] Use end location instead of start location
- [X] Add tests

@rgrinberg I'm not sure what you mean by "Wrap printing environment" and why it's better to use the end location instead of the start, would appreciate some details/pointers on these 🙂 

Also, reading https://github.com/ocamllabs/vscode-ocaml-platform/issues/381, I'm under the impression that what we want is to create the interface when switching from the implementation to a non-existing interface. In this case, what's the reasoning behind implementing a new code action, and not using the existing `ocamllsp/switchImplIntf`?


rgrinberg:
>  I'm under the impression that what we want is to create the interface when switching from the implementation to a non-existing interface. In this case, what's the reasoning behind implementing a new code action, and not using the existing ocamllsp/switchImplIntf?

It seems like a fragile UI. What if I want to infer the interface from scratch after not being satisfied with the current interface, how do I ask the language server to infer the interface file? I prefer not to add functionality that relies on a special state to be active.

ocaml-lsp-server/src/inferred_intf.ml
```diff
@@ -0,0 +1,48 @@
...
+      let typer = Mpipeline.typer_result pipeline in
+      let sig_ : Types.signature =
+        let typedtree = Mtyper.get_typedtree typer in
+        match typedtree with
+        | `Interface _ ->
+          Printf.fprintf stderr "assert false\n";
+          assert false
+        | `Implementation impl -> impl.str_type
+      in
+      Format.asprintf "%a@." Printtyp.signature sig_)
```
rgrinberg, line 16: This call needs to be wrapped with:

```ocaml
Printtyp.wrap_printing_env env ~verbosity
```

This is done so that `-short-paths` is respected.

rgrinberg:
> and why it's better to use the end location instead of the start, would appreciate some details/pointers on these

I don't remember what I originally had in mind, but I think I considered two kinds of behaviors for the code action:
* Insert the inferred interface at the current position
* Replace the current interface file with the inferred interface

What seems like a decent compromise to me is:
We replace the range from the code action request with the inferred interface.

.ocamlformat
```diff
@@ -1,4 +1,4 @@
-version=0.15.0
+version=0.15.1
```
rgrinberg, line 1: I pushed this change directly to master.

ulugbekna:
The suggestion I have below is in terms of user experience/use cases with the extension backed by ocaml-lsp. If we like it, we can discuss _how_ in particular to implement these later.

What if we have two kinds of "infer interface": 
  (1) for a file, i.e., ocaml-lsp is given a file for which it then infers the interface
  (2) for a module, i.e., ocaml-lsp infers interface for a piece of code (not sure whether it should be a position in a document or a piece of code)

More details:

(1) "Infer interface for a file" use case
 User could press `alt+shift+o` (in contrast to `alt+o` that simply switches to intf file) to switch from an ml file to mli:
  - if mli doesn't exist, user sees an unsaved inferred mli file
  - if mli exists, user sees a vscode diff (either as vscode git diff (side-by-side comparison) or as vscode merge-conflict resolution where user can accept the change or not)

image

OR 

image

(2) "infer interface for a piece of code" use case
 User either selects a piece of code and calls a command "Infer interface" and gets the inferred interface in their clipboard 

or 

calls command "Infer interface" without selecting anything and the smallest module surrounding user's position in code gets interface inferred and copied to clipboard. 

----

Depending on these use cases, we can design the interface ocaml-lsp should have for its clients.

rgrinberg:
> if mli doesn't exist, user sees an unsaved inferred mli file

Sounds like a good idea. We might as well do this for `alt+o` though. It seems strictly better than presenting an empty file.

> if mli exists, user sees a vscode diff (either as vscode git diff (side-by-side comparison) or as vscode merge-conflict resolution where user can accept the change or not)

This seems like a feature that could be provided by vscode independent of any language server and plugin. Vscode knows the edit we're suggesting to the user, so it could spin up this diff on its own - for any code action.

> "infer interface for a piece of code" use case
User either selects a piece of code and calls a command "Infer interface" and gets the inferred interface in their clipboard

Or perhaps to log this inferred value somewhere. Right now, one serious limitation of the lsp server is that hover is crippled compared to merlin's type enclosing feature. This could be a way to compensate for that. Perhaps first, I'd like to see if 3.16 improves anything in this area.

But let's focus on creating the smallest viable feature in this PR, and save improving the UI in subsequent work. Feel free to make other tickets/pr's.

tmattio:
The only missing piece is the handling of unopened documents. From what I can see, there's no real clean way of doing it:

The content of the file could be retrieved from the server, but this would only work if the file is a `file://`, if the file is a `memory://` or `ftp://` for instance, we most likely won't be able to retrieve the content from the server.

I've been looking at how we could request the file to be loaded by the client from the server, but did not find anything in the documentation. LSP is pretty new to me, so maybe @ulugbekna @rgrinberg you know if there is a way to do this?
However, even if we find a way to do this, the client would have to execute the codeActionsProvider command manually for the interface file when the implementation file is loaded on the client-side, because I figure VSCode won't request the codeActions a second time without any user input. This does not seem like a good solution either.

Do you see any alternatives?

ocaml-lsp-server/src/inferred_intf.ml
```diff
@@ -0,0 +1,76 @@
...
+    match Document_store.get_opt store impl_uri with
+    | None -> Fiber.return (Ok None)
+    | Some impl -> (
+      let open Fiber.O in
+      let+ intf = infer_intf impl in
+      match intf with
+      | Error e -> Error (Jsonrpc.Response.Error.of_exn e)
+      | Ok intf -> Ok (Some (code_action_of_intf intf_uri intf params.range)) )
+    )
+  | Impl -> Fiber.return (Ok None)
```
rgrinberg, line 51: The `Impl` clause shuould be on top since it's shorter.

rgrinberg:
> I've been looking at how we could request the file to be loaded by the client from the server, but did not find anything in the documentation. LSP is pretty new to me, so maybe @ulugbekna @rgrinberg you know if there is a way to do this?

There is a way to ask the client to display a file in 3.16, but we can't rely on that.

> Do you see any alternatives?

Let's do the naive thing for now, and just look for the file directly on disk. If it doesn't work, we'll fail gracefully. If the file is `ftp://` or `memory://` I have no idea how the user is building the project and things aren't going to work anyway.

ocaml-lsp-server/src/position.ml
```diff
@@ -1,6 +1,15 @@
 open Import
 include Lsp.Types.Position
 
+let start = { line = 1; character = 1 }
+
+let end_of_string s =
+  let lines = String.split_lines s in
+  let length_last_line =
+    ListLabels.nth lines (List.length lines - 1) |> String.length
```
rgrinberg, line 9: `List.last` is already available and is safer to use.

rgrinberg: Actually, I don't see `end_of_string` being used anywhere anyway.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -164,22 +169,30 @@ let on_initialize rpc (ip : Lsp.Types.InitializeParams.t) =
...
+    | Some _
+    | None ->
+      let+ action = f () in
+      Option.map action ~f:(fun destruct -> `CodeAction destruct)
+  in
+  let* destruct_action =
+    code_action (CodeActionKind.Other Destruct_lsp.action_kind) (fun () ->
+        Destruct_lsp.code_action doc params)
+  in
+  let* inferred_intf_action =
```
rgrinberg, line 189: You should be able to generate these actions concurrently with `Fiber.parallel_map`. It doesn't matter now, but hopefully it will once we add non merlin based actions.

ocaml-lsp-server/src/inferred_intf.ml
```diff
@@ -0,0 +1,76 @@
...
+    | Some impl -> (
+      let+ intf = infer_intf impl in
+      match intf with
+      | Error e -> Error (Jsonrpc.Response.Error.of_exn e)
+      | Ok intf -> Ok (Json.t_of_yojson (`String intf), state) ) )
+  | Some _
+  | None ->
+    Fiber.return
+    @@ Error
+         (Jsonrpc.Response.Error.make ~code:InvalidRequest
```
ulugbekna, line 75: You probably followed `switchImplIntf` in error codes, but it feels like uses of `InvalidParams` and `InvalidRequest` in the selected code should probably be swapped.

ulugbekna (commented):
Thanks for the PR :) 

A couple of notes:

- For custom requests, I documented specs (description, capability, request, response) we came up with. If you found the specs for `ocamllsp/switchImplIntf` for example useful, consider doing something similar for this custom request.

- Since `inferred_intf` can provide a code action and reply a custom request, I would consider splitting that functionality into logical blocks either with modules or simply some rearrangements and comments

- I would document the use case for the feature (e.g., we want to provide a code action that is triggered only when a user's cursor is in the empty mli file AND a custom request that will fill in an intf counterpart file if it doesn't exist); that would help : 

(1) to review what the code is trying to do and why
(2) serve as a ready recipe for users and can be added to both changelog and README

- (likely to happen in future if at all) maybe format the generated intf file if the formatter is available (reuse `Fmt` lib in `ocaml-lsp-server`) so that user can look through it before they save the file, which would anyway trigger formatting

rgrinberg:
Why do we need a custom request in the first place? Isn't it possible to do everything through code actions?

ulugbekna:
> Why do we need a custom request in the first place? Isn't it possible to do everything through code actions?

IIUC, when we switch from `test.ml` to a non-existing `test.mli` and want it to be filled with inferred interface, if we only have a code action, we would need to automatically trigger the code action on user's behalf, which is less desirable I guess. 

A case when a new `mli` file should be filled in with interface could be piggy-backed on existing `ocamllsp/switchImplIntf` request with the response type changing to something like 

```ts
interface Response = DocumentUri /* if document exists */ 
                                   | { uri: DocumentUri, content: string } /* if mli file only doesn't exist */
```

(For this reason, I am proposing to first model how the user is supposed to use the feature(s), and then design how language server and client should communicate on which events and with what request/responses)

rgrinberg:
> A case when a new mli file should be filled in with interface could be piggy-backed on existing ocamllsp/switchImplIntf request with the response type changing to something like

That seems better. Even simpler is:

```
{
  documentUri: uri;
  content : string option
}
```

ocaml-lsp-server/src/position.ml
```diff
@@ -1,6 +1,15 @@
 open Import
 include Lsp.Types.Position
 
+let start = { line = 1; character = 1 }
```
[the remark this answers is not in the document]
tmattio: I think you're right. I'll check again. Anyway, I think it's not used anymore because we use the request range as the response range now.

tmattio:
> IIUC, when we switch from test.ml to a non-existing test.mli and want it to be filled with inferred interface, if we only have a code action, we would need to automatically trigger the code action on user's behalf, which is less desirable I guess.

Right, in addition to the design choice, I realized that code actions are not requested for an empty document. We could request them manually, but IMHO the fact that we would need to do this indicates that what we're going against the intended use of code actions.

> That seems better. Even simpler is:

Sounds like a good approach as well. The reason I went for a custom request is the need for flexibility that you expressed:

> I prefer not to add functionality that relies on a special state to be active.

It seemed to me that a custom request was a way to implement the feature in a re-usable way. But if the overhead this implies does not seem justified, I can refactor to just have it in `ocamllsp/SwitchImplIntf`

rgrinberg:
> It seemed to me that a custom request was a way to implement the feature in a re-usable way. But if the overhead this implies does not seem justified, I can refactor to just have it in ocamllsp/SwitchImplIntf

Actually, all of this seems like unnecessary complexity to save a keystroke for the user. Let's drop this requirement and see how the protocol evolves. I don't particularly care if the user has to trigger the code action manually.

tmattio:
> Let's drop this requirement and see how the protocol evolves.

That's a pity, I really liked the UX with the way it works at the moment.. IMO the complexity is well contained as all the logic can be found in a single module, so I would argue that the implementation is ok as-is (when I address the remaining todos obviously) and can be revisited easily later if needed, but if you really think the cost outweighs the benefits, I'll remove it.

rgrinberg:
The main measure of complexity that matters here is how self contained the feature is. The current implementation is not satisfactory from this perspective unfortunately. Keep in mind that features that require custom requests are also far less useful. We have plenty of non vscode users that would not benefit from this.

What you can do is use the `workspace/applyEdit` notification to try and insert the inferred interface if you'd really like. I'd suggest that you do this in a separate PR though. The feature is in a satisfactory even without this.

ulugbekna:
I'm in favor of having all features this PR introduces whether in this or a separate PR. 

I don't see anything wrong with enhancing `ocamllsp/switchImplIntf` or creating a new command as long as it's documented.

tmattio:
I reverted the custom request and will make another PR to handle the automatic generation of interfaces (most likely with the changes in `ocamllsp/switchImplIntf`), we can continue the discussion there.

I addressed the other PR reviews and implemented the handling of unopened implementation files. I'll run some more tests, but it should be ready to merge now if you don't have any other suggestions.

tmattio:
I've run some more tests with `vscode-ocaml-platform` and everything seems to be working as expected. I also added a test suite to make sure that this supports Reason syntax.

The CI failures seems unrelated to these changes.

ocaml-lsp-server/src/inferred_intf.ml
```diff
@@ -0,0 +1,85 @@
...
+  in
+  let title = String.capitalize_ascii "Insert inferred interface" in
+  CodeAction.create ~title ~kind:(CodeActionKind.Other action_kind) ~edit
+    ~isPreferred:false ()
+
+let read_file filename =
+  let ch = open_in filename in
+  let s = really_input_string ch (in_channel_length ch) in
+  close_in ch;
+  s
```
ulugbekna, line 33: you could do `Io.read_file`, which is contained in dune's stdlib `Stdune` that we use and have opened in `Import`

ocaml-lsp-server/src/inferred_intf.ml
```diff
@@ -0,0 +1,85 @@
...
+  | ".mly" -> "ocaml.menhir"
+  | ext -> failwith ("Unknown extension " ^ ext)
+
+let force_open_document (state : State.t) uri =
+  let open Fiber.O in
+  log ~title:Logger.Title.Warning "Forcing file open on %s" (Uri.to_string uri);
+  let filename = Uri.to_path uri in
+  log ~title:Logger.Title.Warning "Path is %s" filename;
+  let text = read_file filename in
+  log ~title:Logger.Title.Warning "Content is %s" text;
```
ulugbekna, line 52: do you want to keep all this logging?

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -163,23 +167,45 @@ let on_initialize rpc (ip : Lsp.Types.InitializeParams.t) =
...
+      ]
+      ~f:code_action
+  in
+  let code_action_results =
+    result_of_list code_action_results |> Result.map ~f:List.filter_opt
+  in
+  match code_action_results with
+  | Ok [] -> Fiber.return (Ok (None, state))
+  | Ok l -> Fiber.return (Ok (Some l, state))
+  | Error err -> Fiber.return (Error err)
```
ulugbekna, line 208: you could do `let+ code_action_results ...` and remove all `Fiber.return`s, I think

ulugbekna (commented):
a couple of quick notes

ocaml-lsp-server/src/inferred_intf.ml
```diff
@@ -0,0 +1,85 @@
...
+let language_id_of_fname s =
+  match Filename.extension s with
+  | ".mli" -> "ocaml.interface"
+  | ".ml" -> "ocaml"
+  | ".rei"
+  | ".re" ->
+    "reason"
+  | ".mll" -> "ocaml.ocamllex"
+  | ".mly" -> "ocaml.menhir"
+  | ext -> failwith ("Unknown extension " ^ ext)
```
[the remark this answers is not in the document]
tmattio: I wanted to use this, but it wouldn't differentiate between `ocaml.interface` and `ocaml`

ocaml-lsp-server/src/inferred_intf.ml
tmattio: It does not seem to be in `Import`, but I can add it if that makes sense

ocaml-lsp-server/src/inferred_intf.ml
tmattio: No, that's debugging logs, thanks!

ocaml-lsp-server/src/ocaml_lsp_server.ml
tmattio: Indeed, thanks

ocaml-lsp-server/src/position.ml
tmattio: I removed it

ocaml-lsp-server/src/inferred_intf.ml
tmattio: Done!

ocaml-lsp-server/src/inferred_intf.ml
tmattio: Reverted this part of the PR

rgrinberg:
> PR to handle the automatic generation of interfaces (most likely with the changes in ocamllsp/switchImplIntf), we can continue the discussion there.

Before you spend time on this, consider the `workspace/applyEdit` approach. We should really avoid editor specific approaches as much as possible. There's so many things left to work on that will benefit all editors.

ocaml-lsp-server/src/inferred_intf.ml
rgrinberg: Yes, please add it. Feel free to use whatever modules you want from stdune. The only reason why we don't do `open Stdune` is because some modules shadow modules from compiler-libs (`Env` for example).

ocaml-lsp-server/src/inferred_intf.ml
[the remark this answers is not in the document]
rgrinberg: > I wanted to use this, but it wouldn't differentiate between ocaml.interface and ocaml

That would be a bug with `to_language_id` then.

On a related note, I'm not sure if `ocaml.interface` works on non vscode lsp clients. Let me confirm on emacs.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -163,23 +167,45 @@ let on_initialize rpc (ip : Lsp.Types.InitializeParams.t) =
   let state = { state with init = Initialized ip.capabilities } in
   Fiber.return @@ Ok (initialize_info, state)
 
+let result_of_list l =
+  let rec aux acc = function
+    | [] -> Ok (acc |> List.rev)
+    | Error err :: _rest -> Error err
+    | Ok el :: rest -> aux (el :: acc) rest
+  in
+  aux [] l
```
rgrinberg, line 176: This looks like `Result.List.all` to me.

dune-project
```diff
@@ -53,6 +53,7 @@ possible and does not make any assumptions about IO.
   dot-merlin-reader
   (csexp (>= 1.2.3))
   (result (>= 1.5))
+  (reason :with-test)
```
rgrinberg, line 56: Can we drop reason from the tests? I recently removed it because they don't publish it regularly on opam: https://github.com/ocaml/ocaml-lsp/issues/247

If you'd like to exercise reason functionality, I'd suggest to create a mock `refmt` binary. I think it's fine to just drop it though.

ocaml-lsp-server/src/inferred_intf.ml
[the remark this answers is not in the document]
ulugbekna: If we are sure that we do need specific lang id, I would propose to move lang id handling to a separate module `Language_id` that would properly handle `Syntax` and `Kind`. I guess we could do that in another PR.

dune-project
tmattio: I'm not sure how I feel about supporting reason, but having no tests for it. If I understand correctly, release on Opam is the issue, but there's nothing blocking us from using reason in development, right?

ocaml-lsp-server/src/ocaml_lsp_server.ml
tmattio: Right, thanks, I'm not familiar with Stdune yet 🙂

dune-project
rgrinberg: > If I understand correctly, release on Opam is the issue, but there's nothing blocking us from using reason in development, right?

There are issues with setting it up in CI however.

> I'm not sure how I feel about supporting reason, but having no tests for it.

We can add tests for it, but we just need to mock `refmt`. As long as we match the cli, the tests using the mock binary would be just as valuable.

tmattio: Ok, I'll remove the tests from this PR and will bring them back in another one - don't want to delay the merge any longer if that's the only blocker 🙂

rgrinberg:
Looks good. Just needs a CHANGES entry

tmattio:
> Looks good. Just needs a CHANGES entry

It's there: https://github.com/ocaml/ocaml-lsp/pull/308/files#diff-d975bf659606195d2165918f93e1cf680ef68ea3c9cab994f033705fea8238b2R5

ocaml-lsp-server/src/inferred_intf.ml
tmattio: I added it 👍

PR #314, opened 2020-11-24
Add ocamllsp/inferIntf as a custom request

tmattio:
I've been experimenting with `workspace/applyEdit`.

I tried two options:

- Sending the request when receiving a `didOpen` for an interface that does not yet exist. Unfortunately, it does not work because VSCode does not send `didOpen` notifications for file previews. And if we save the file, then there's no way to know if the file is new or if it already exists, in which case we should not try to initialize it.

- Sending a notification when switching from an implementation to an interface that does not exist. Unfortunately, the `applyEdit` request does not seem to be processed by VSCode. Moreover, the implementation file might not be in the store, so we would need to open it manually in this case, which seem hacky.

After experimenting with it, I still think a custom request is our best option there. Since we already provide the interface inference feature through a code action that can be triggered manually, the fact that this feature will only be available on VSCode seems acceptable to me.

I'm open to other suggestions on how we could provide it with normal LSP capabilities though, let me know if you have any other idea.

### DIFF
--- a/ocaml-lsp-server/docs/ocamllsp/inferIntf-spec.md
+++ b/ocaml-lsp-server/docs/ocamllsp/inferIntf-spec.md
@@ -7,6 +7,9 @@ The document URI in the request has to be open before sending a the request.
 
 If the file cannot be found in the document store, an error will be returned.
 
+Warning: this custom request is meant to be consumed by `ocaml-vscode-platform` exclusively,
+it can be removed any time and should not be relied on.
+
 ##### Client capability
 
 nothing that should be noted
--- a/ocaml-lsp-server/docs/ocamllsp/inferIntf-spec.md
+++ b/ocaml-lsp-server/docs/ocamllsp/inferIntf-spec.md
@@ -1,0 +1,27 @@
+#### Infer Interface Request
+
+Infer Interface Request is sent from the client to the server to get the infered
+interface for a given module implementation.
+
+The document URI in the request has to be open before sending a the request.
+
+If the file cannot be found in the document store, an error will be returned.
+
+##### Client capability
+
+nothing that should be noted
+
+##### Server capability
+
+property name: `handleInferIntf`
+property type: `boolean`
+
+##### Request
+
+- method: `ocamllsp/inferIntf`
+- params: `DocumentUri` (see [`DocumentUri`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#uri) in LSP specification)
+
+##### Response
+
+- result: String
+- error: code and message set in case an exception happens during the processing of the request.
--- a/ocaml-lsp-server/src/inference.ml
+++ b/ocaml-lsp-server/src/inference.ml
@@ -1,0 +1,68 @@
+open Import
+
+let infer_intf_for_impl doc =
+  match Document.kind doc with
+  | Intf ->
+    Fiber.return
+      (Error
+         (Invalid_argument "the provided document is not an implementation."))
+  | Impl ->
+    Document.with_pipeline doc (fun pipeline ->
+        let typer = Mpipeline.typer_result pipeline in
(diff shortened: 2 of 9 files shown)


rgrinberg:
> The document URI in the request has to be open before sending a the request.

We should worry about fixing this first.

I'm not particularly worried about adding this feature to be frank. Okay, we saved a keystroke for a single editor, unless we run into some opaque edge cases. It's not very convincing for the amount of code it adds.

> I'm open to other suggestions on how we could provide it with normal LSP capabilities though, let me know if you have any other idea.

Luckily, the lsp standard is a live protocol to which we can propose amendments. I'd much rather pursue that avenue first.

tmattio:
> Okay, we saved a keystroke for a single editor, unless we run into some opaque edge cases

It's not just a keystroke. The code actions are only requested when there is content on the document, so when switching to an empty interface, the user has to start typing something to trigger the code action, then select the content, because it needs to be replaced, then insert the inferred interface. It's a terrible user experience IMO, especially for a feature that seems to be important for users.

> Luckily, the lsp standard is a live protocol to which we can propose amendments. I'd much rather pursue that avenue first.

It's not incompatible: we could make an RFC to LSP, and support the feature with a custom request in the meantime.

rgrinberg:
> The code actions are only requested when there is content on the document, so when switching to an empty interface, the user has to start typing something to trigger the code action, then select the content, because it needs to be replaced, then insert the inferred interface. It's a terrible user experience IMO, especially for a feature that seems to be important for users.

I see the distinction now. In emacs, I'm able to trigger any code action without typing. It appears that in vscode, the bulb only turns on when you start typing. That's too bad.

> I've been experimenting with workspace/applyEdit.

It truly is a shame that we can't do the insertion with this command. The user has the ability to reject any changes proposed by this request, which is exactly what we want.

Alright, we can revive this PR. But seriously, this entire thing is a tiny little feature hardly justifies so much work and discussion. There's already a way to implement print inferred interface files in dune, so it was never that all necessary. It was just a little warm up task before proceeding to implement much more important features.

ocaml-lsp-server/docs/ocamllsp/inferIntf-spec.md
```diff
@@ -0,0 +1,27 @@
+#### Infer Interface Request
+
+Infer Interface Request is sent from the client to the server to get the infered
```
rgrinberg, line 3: The docs are good, but let's make sure to mention the most important thing: these custom requests aren't to be relied on and can be removed without warning.

ocaml-lsp-server/src/inference.ml
```diff
@@ -0,0 +1,68 @@
+open Import
+
+let infer_intf_for_impl doc =
+  match Document.kind doc with
+  | Intf ->
+    Fiber.return
+      (Error
+         (Invalid_argument "the provided document is not an implementation."))
```
rgrinberg, line 8: We don't use `Invalid_argument` in our code base. If it's a programming error, then it should be `Code_error.E`, if it's something for the user, it should be `Jsonrpc.Response.Error.t`

tmattio:
I believe I addressed all of the reviews and concerns. Merging this now.

rgrinberg:
This PR is still using failwith. We don't use this function in this codebase.

tmattio:
Ah, I thought it was ok since we use it at several places (including `ocaml-lsp-server`): https://github.com/ocaml/ocaml-lsp/search?q=failwith

But I'll update that.

rgrinberg:
Would be good to fix those cases as well. We have code errors for programming errors and jsonrpc errors for all user visible errors.

tmattio:
Ok, I'll make a PR to remove all of them (when possible) then

rgrinberg:
Most of them aren't as urgent because they aren't even part of the rpc server

PR #317, opened 2020-11-28
Upgrade to ocamlformat.0.16.0

tmattio:


rgrinberg:
I update ocamlformat for vscode, ocamllsp, dune all at once usually.

PR #318, opened 2020-11-28
Improve Github Actions workflow

tmattio:
Builds on top of #296 (I didn't have the rights to push to your remote @smorimoto) with:

- Updated Makefile
- Install dependencies from lock files in CI
- Fix build on 4.11.1
- Enable build on all OCaml version and platforms
- Add caching in CI to spare some build minutes

I'm curious as to why the tests were run only on 4.11.1, is there any reason for this @rgrinberg?
From what I see, the tests fail for every combination except 4.11.1 on Linux and macOS. We should probably fix that.


Makefile
```diff
@@ -1,59 +1,61 @@
...
 
-opam-release:
-	dune-release distrib --skip-build --skip-lint --skip-tests --include-submodules
+.PHONY: release
+release: all ## Release on Opam
+	opam exec -- dune-release distrib --skip-build --skip-lint --skip-tests --include-submodules
 	# See https://github.com/ocamllabs/dune-release/issues/206
 	DUNE_RELEASE_DELEGATE=github-dune-release-delegate dune-release publish distrib --verbose
-	dune-release opam pkg
-	dune-release opam submit
```
rgrinberg, line 59: I have `dune-release` in a custom switch so this doesn't work well for me unfortunately. What's the issue with running make like this: `$ opam exec -- make`? It seems to be just as convenient.

rgrinberg:
> I'm curious as to why the tests were run only on 4.11.1, is there any reason for this

We don't have first class support for older versions of OCaml. It's quite likely that we'll drop support for them in the very near future following merlin.

Makefile
tmattio: More typing? 😄  More seriously, I don't have a strong opinion here, I'll just revert this.

tmattio:
> We don't have first class support for older versions of OCaml. It's quite likely that we'll drop support for them in the very near future following merlin.

What's the latest OCaml version we are supporting? Should I remove the others from the CI then?

rgrinberg:
> What's the latest OCaml version we are supporting? 

4.11

> Should I remove the others from the CI then?

Let's keep them for now and officially drop support a little later.

Makefile
rgrinberg: `$ opam exec -- bash` solves that :)

One thing that we can do to make it work for both of us is to add this line:
```
-include Makefile.dev
```

And then add some code to `Makefile.dev` to turn on the switch. I assume it's possible to set the switch by either customizing the shell command or by reading the environment variables and re-exporting them through make (might be a little difficult)

tmattio:
@rgrinberg I removed the `opam exec --` from the Makefile

rgrinberg (approved):

rgrinberg:
Hmm, I don't I reviewed this carefully enough. Looks like this commit comes with a regression:

```
(cd _build/default && /home/runner/.opam/4.11.1/bin/ocamlformat --intf ocaml-lsp-server/src/compl.mli) > _build/default/ocaml-lsp-server/src/.formatted/compl.mli
469
ocamlformat: Error while parsing .ocamlformat:
470
             For option "version": expecting "0.16.0" but got "0.15.1"
```

Also, this commit erroneously introduces tests on various old versions of OCaml. There's no need to run tests on them.

Issue #224, opened 2020-07-10
Handle Requests Concurrently


rgrinberg:
At the moment, the server handles requests sequentially. That is, the server must finish processing a request before proceeding to the next one. We'd like to improve this behavior to allow processing requests concurrently. In particular, we'd like requests that require merlin not to block requests that do not require merlin from being processed and vice versa.

rgrinberg:
Update: the jsonrpc server now supports the required infrastructure, we just need to push it down to the lsp implementation.

PR #323, opened 2020-11-30
Implement basic cancellation

rgrinberg:
Now `$/cancelRequest` is used to release positions in the merlin queue

In the future, we'll need something a little more robust. Such as:

- Run cancellers inside Fiber_detached
- Allow to hook finalizers as well as cancellers

### DIFF
--- a/lsp-fiber/src/rpc.ml
+++ b/lsp-fiber/src/rpc.ml
@@ -24,6 +24,34 @@ module Reply = struct
     | Later k -> Jsonrpc_fiber.Reply.later (fun send -> k (fun r -> send (f r)))
 end
 
+module Cancel = struct
+  type state =
+    | Pending of { mutable callbacks : (unit -> unit Fiber.t) list }
+    | Finished
+
+  type t = state ref
+
+  let var = Fiber.Var.create ()
+
+  let register f =
+    Fiber.Var.get var
+    |> Option.iter ~f:(fun t ->
+           match !t with
+           | Pending p -> p.callbacks <- f :: p.callbacks
+           | Finished -> ())
+
+  let create () = ref (Pending { callbacks = [] })
+
+  let destroy t = t := Finished
+
+  let cancel t =
+    match !t with
+    | Finished -> Fiber.return ()
+    | Pending { callbacks } ->
+      t := Finished;
+      Fiber.parallel_iter callbacks ~f:(fun f -> f ())
+end
+
 module State = struct
   type t =
     | Waiting_for_init
@@ -70,6 +98,8 @@ module type S = sig
     _ t -> 'resp out_request -> ('resp, Response.Error.t) result Fiber.t
 
   val notification : _ t -> out_notification -> unit Fiber.t
+
+  val on_cancel : (unit -> unit Fiber.t) -> unit
 end
 
 module type Request_intf = sig
@@ -119,6 +149,8 @@ struct
     ; (* Filled when the server is initialied *)
       initialized : Initialize.t Fiber.Ivar.t
     ; mutable req_id : int
+    ; pending : (Jsonrpc.Id.t, Cancel.t) Table.t
+    ; detached : Fiber_detached.t
     }
 
   and 'state on_request =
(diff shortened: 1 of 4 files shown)


lsp-fiber/src/rpc.ml
```diff
@@ -152,10 +181,16 @@ struct
...
+        let cancel = Cancel.create () in
+        Table.set t.pending req.id cancel;
+        let+ response =
+          Fiber.Var.set Cancel.var cancel (fun () ->
+              h_on_request.on_request t r)
+        in
         match response with
         | Error e -> (Reply.now (Response.error req.id e), state)
         | Ok (response, state) ->
+          Cancel.destroy cancel;
```
tmattio, line 193: You need to destroy it in case of errors as well, right?

ocaml-lsp-server/src/document.ml
```diff
@@ -83,10 +83,23 @@ let timer t = t.timer
 
 let source doc = Mpipeline.raw_source doc.pipeline
 
+let await task =
+  let open Fiber.O in
+  let () = Cancel.register (fun () -> Scheduler.cancel_task task) in
```
tmattio, line 88: I would expect `register` to be called for all of the requests, why only in `Document.with_pipeline`?

lsp-fiber/src/rpc.mli
```diff
@@ -1,6 +1,12 @@
 open! Import
 open Jsonrpc
 
+module Cancel : sig
```
tmattio, line 4: If the `Cancel` module was hidden and we provided functions like `val register_cancel : 'a t -> (unit -> unit Fiber.t) -> unit`, we wouldn't need to have mutable and ref in the `Cancel`, would make it easier to reason about the flow.

tmattio (commented):
LGTM overall. The combination of `ref`, `mutable`, and `Fiber.var.t` make it a bit hard to reason about the code. Perhaps that could be solved if we didn't expose the `Cancel` module (see comment)? 

Could we add a test for this also?

lsp-fiber/src/rpc.ml
rgrinberg: Good call. And I also forgot to remove the key from the table. Should be fixed.

ocaml-lsp-server/src/document.ml
rgrinberg: There's nothing else that we can cancel. All of our cancellation points are essentially fixed at where we call `Ivar.read`, and we use ivar read for exactly two things: async & timers. Timers don't belong to any requests, so there's only async left.

We could make it possible to cancel formatting requests I suppose, but it seems more trouble than its worth.

lsp-fiber/src/rpc.mli
rgrinberg: What's `'a t` here?

rgrinberg:
I just realized that we need #224 for cancellation to actually work. After all, we need to process cancellation while other requests are running.

lsp-fiber/src/rpc.mli
tmattio: `Rpc.S.t`, so that we could inline the API of `Cancel` in `Rpc.S`.

ocaml-lsp-server/src/document.ml
tmattio: IMO we should try to make it general and work for all requests, but if you think that's too much trouble, let's wait until we need to register cancellations for other requests 🙂

lsp-fiber/src/rpc.mli
rgrinberg: Ah, but that doesn't work as that `'a t` isn't request specific. We'd need to introduce something like `Context.t` on the jsonrpc side. That's something that I've been itching to do, but didn't get around to.

I admit that using fiber local storage isn't very attractive, but it's the easiest way to proceed without large refactoring. I will get rid of the unnecessary `Cancel` public module and expose a single function.

rgrinberg:
I just realized that our API is too limited to actually exercise this feature. For now, this PR is good as it least doesn't break anything.

Issue #312, opened 2020-11-20
Signature Help


rgrinberg:
This ticket tracks support for the `textDocument/signatureHelp` request.

I've nominated @mnxn to work on this feature. Listed below are the various components of this feature. There's no need to finish all of this before we ship something however.

## Components

- [ ] Functions
- [ ] Records
- [ ] Constructors
- [ ] Methods
- [ ] Trigger characters. I'm assuming this set should be good: `[' '; '~', ':', '('; '{']`
- [ ] Active Parameters.
- [ ] Active Signature. This should be useful when we need nested signatures e.g. when typing `foo { field ; |`, we want help for the function and the record.
- [ ] Documentation: signatures
- [ ] Documentation: parameters
- [ ] Labels. Let's stick to supporting offsets only.

PR #324, opened 2020-12-01
Implement signature help request

mnxn:
Partially completes #312 

Depends on https://github.com/rgrinberg/merlin/pull/8

Currently supports showing signature help when an expression is detected as having a function type or is being applied. So it will show signature help for these two scenarios:

```ocaml
let f x = x
let _  = f |
```
> f : 'a -> 'a

```ocaml
let f x = x
let _  = f 1|
```
> f : ___int___ -> int

The second one will underline the specific parameter that is currently being applied. Operators also show signature help.

---
**Notes:**

- I haven't figured what the best way to provide function/parameter documentation is. The function could just return positions for function documentation, but I feel like there should be a better way to do it.

- Showing signature help for operators or after entering a space after a function-type value might be annoying to users. Perhaps the functionality should be configurable? I couldn't find a built-in VSCode setting to disable signature help.

- The current implementation seems to depend on having correctly typed arguments. This doesn't show any signature help:
```ocaml
let _ = string_of_int true|
```

- Providing too many arguments just shows the function type without any parameters highlighted, despite the type error.
```ocaml
let _ = string_of_int 1 2|
```
> string_of_int : int -> string

---

**Images:**
Space after function-typed value:
image

Providing a function argument:
image

Anonymous function:
image

Operator:
image


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -341,6 +346,46 @@ let hover (state : State.t) { HoverParams.textDocument = { uri }; position } =
...
+let signature_help (state : State.t)
+    { SignatureHelpParams.textDocument = { uri }; position; context = _ } =
+  let store = state.store in
+  let uri = Uri.t_of_yojson (`String uri) in
+  let open Fiber.Result.O in
+  let* doc = Fiber.return (Document_store.get store uri) in
+  let pos = Position.logical position in
+  let open Fiber.O in
+  let+ result =
+    Document.with_pipeline doc (fun pipeline ->
```
rgrinberg, line 358: You can just use `with_pipeline_exn` here. Exceptions are already converted to jsonrpc errors.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -341,6 +346,46 @@ let hover (state : State.t) { HoverParams.textDocument = { uri }; position } =
...
+        let context = Merlin_analysis.Completion.application_signature node in
+        match context with
+        | `Application { fun_name; signature; param_offset } ->
+          let fun_name = Option.value ~default:"_" fun_name in
+          let prefix = sprintf " %s : " fun_name in
+          let offset = String.length prefix in
+          let parameters =
+            match param_offset with
+            | Some (param_start, param_end) ->
+              [ ParameterInformation.create
```
rgrinberg, line 371: How comes there's always only one parameter? Shouldn't we show all of them here and pick the right one with `activeParameter`?

mnxn:
The first non-labelled parameter is now underlined right after a function-typed value:

image


Typing a `~` will underline the `f:('a -> 'b)` instead.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -34,6 +34,11 @@ let initialize_info : InitializeResult.t =
     CompletionOptions.create ~triggerCharacters:[ "."; "#" ]
       ~resolveProvider:true ()
   in
+  let signatureHelpProvider =
+    SignatureHelpOptions.create
+      ~triggerCharacters:[ " "; "~"; "?"; ":"; "("; "["; "{" ]
```
rgrinberg, line 39: `{` is for records in the future right? Why `[` though?

rgrinberg (approved):
Looks good. I'll merge both this and the merlin PR they are both ready.

rgrinberg:
Worth adding a CHANGES entry as well.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -36,7 +36,7 @@ let initialize_info : InitializeResult.t =
   in
   let signatureHelpProvider =
     SignatureHelpOptions.create
-      ~triggerCharacters:[ " "; "~"; "?"; ":"; "("; "["; "{" ]
+      ~triggerCharacters:[ " "; "~"; "?"; ":"; "("; "{" ]
```
rgrinberg, line 39: I'd remove `{` too. As we don't support records yet.

ocaml-lsp-server/src/ocaml_lsp_server.ml
mnxn: I thought I noticed an interaction with the function signature help, but I checked again and it doesn't make a difference. Removed.

rgrinberg:
Awesome work. Merged manually.

mnxn:
> Awesome work. Merged manually.

Thanks, I was having trouble figuring out why Github was detecting conflicts...

tmattio:
Thanks for the great work on this @mnxn, the feature is amazing 🎉 

I noticed two things when testing:

- The highlighted argument seems to always be the first one

image
(I added a space after the first `0`)

- The highlighted argument does not switch to the first labeled with only `~`, it needs the name of the labeled argument as well:

image
then
image

mnxn:
> The highlighted argument seems to always be the first one

The highlighted argument won't always be the first, but it won't detect that you've moved onto the next parameter until you start a new expression. I'm not sure it's possible to detect you've moved on to the next function argument with the expression tree that Merlin provides. It's harder to do in a function-application-by-juxtaposition language like OCaml since it lacks the comma argument delimiters that C-like languages have.


> The highlighted argument does not switch to the first labeled with only ~, it needs the name of the labeled argument as well:

The language server actually isn't providing an `activeParameter` index in this example, but VSCode is defaulting to the first one. We could fix this by _always_ returning an `activeParameter` based on the `triggerCharacter` (`~`, `?`, or other).

mnxn:
> * The highlighted argument does not switch to the first labeled with only `~`, it needs the name of the labeled argument as well:

@tmattio: your second issue should be fixed in https://github.com/ocaml/ocaml-lsp/pull/328

PR #325, opened 2020-12-01
Revert "Improve Github Actions workflow"

rgrinberg:
Reverts ocaml/ocaml-lsp#318


rgrinberg:
Actually, the format issue was still easy to fix. But we should still get rid of all the pointless test jobs.

PR #327, opened 2020-12-02
Replace failwith with Code_error or Jsonrpc.Response.Error

tmattio:


ocaml-lsp-server/src/inference.ml
```diff
@@ -2,7 +2,9 @@ open Import
 
 let infer_intf_for_impl doc =
   match Document.kind doc with
-  | Intf -> failwith "the provided document is not an implementation."
+  | Intf ->
+    Code_error.raise
+      "expected an implementation document, got an interface instead" []
```
rgrinberg, line 7: Is it possible for the user to trigger this error?

ocaml-lsp-server/src/document.ml
```diff
@@ -13,7 +13,9 @@ module Kind = struct
     | ".mli"
     | ".rei" ->
       Intf
-    | ext -> failwith ("Unknown extension " ^ ext)
+    | ext ->
+      Code_error.raise "unsupported file extension"
```
rgrinberg, line 17: Is it possible for the user to trigger this error?

rgrinberg (approved):
Looks good. On the account that it's not possible for the user to trigger the code errors.

ocaml-lsp-server/src/inference.ml
tmattio: The input argument is a `Document.t`, and from what I can see the only way to get that is to get the doc from the store. So a user should not be able to trigger this unless there's a bug.

ocaml-lsp-server/src/document.ml
tmattio: The input argument is a `Document.t`, and from what I can see the only way to get that is to get the doc from the store. So a user should not be able to trigger this unless there's a bug.

tmattio: That's not true, I wasn't looking at the right place.

rgrinberg (approved):

ocaml-lsp-server/src/document.ml
```diff
@@ -13,7 +13,11 @@ module Kind = struct
     | ".mli"
     | ".rei" ->
       Intf
-    | ext -> failwith ("Unknown extension " ^ ext)
+    | ext ->
+      Jsonrpc.Response.Error.raise
+        (Jsonrpc.Response.Error.make ~code:InvalidRequest
+           ~message:(Printf.sprintf "unsupported file extension %S" ext)
```
rgrinberg, line 19: Technically `ext` should occur inside the error's payload

PR #328, opened 2020-12-03
Use completion prefix to determine active signature help parameter

mnxn:
Depends on https://github.com/rgrinberg/merlin/pull/9

Passes the result of `Compl.prefix_of_position` to merlin so it can determine which labelled parameter is active.

Original issue pointed out in this comment: https://github.com/ocaml/ocaml-lsp/pull/324#issuecomment-737107231

Examples: 

image

image

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -353,20 +353,26 @@ let signature_help (state : State.t)
   let open Fiber.Result.O in
   let* doc = Fiber.return (Document_store.get store uri) in
   let pos = Position.logical position in
+  let prefix =
+    Compl.prefix_of_position (Document.source doc) pos ~short_path:false
+  in
   Document.with_pipeline_exn doc (fun pipeline ->
       let typer = Mpipeline.typer_result pipeline in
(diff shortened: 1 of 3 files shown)


rgrinberg:
Looks good. Can we have a test that demonstrates in behavior?

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -353,20 +353,26 @@ let signature_help (state : State.t)
   let open Fiber.Result.O in
   let* doc = Fiber.return (Document_store.get store uri) in
   let pos = Position.logical position in
+  let prefix =
+    Compl.prefix_of_position (Document.source doc) pos ~short_path:false
```
rgrinberg, line 357: Why `~short_path:false`?

mnxn: It doesn't make a different either way. Should I make it `true` instead?

rgrinberg: I'd leave a comment explaining that it doesn't make a difference. `short_path:true` is a little faster.

rgrinberg (approved):
Feel free to merge when you're done.

PR #329, opened 2020-12-03
Makefile & CI tweaks

rgrinberg:
* test should run all tests.
* all test logic should live in the makefile rather than the ci script
* fix typo in test target (test-template)
* add promote/check targets
* do not set dune root. it should respect workspace files above the repo dir

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ $(TEST_E2E_DIR)/node_modules:
 
 .PHONY: all
 all:
-	dune build --root . @install
+	dune build @all
 
 .PHONY: dev
 dev: ## Setup a development environment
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -19,40 +19,49 @@ dev: ## Setup a development environment
 
 .PHONY: build
 build: ## Build the project, including non installable libraries and executables
-	dune build --root .
+	dune build @all
```
tmattio, line 22: Is there a difference between `dune build` and `dune build @all`?

Makefile
```diff
@@ -19,40 +19,49 @@ dev: ## Setup a development environment
...
+.PHONY: promote
+	dune promote
+
+.PHONY: check
+	dune build @check
+
+.PHONY: test-e2e
 test-e2e: $(TEST_E2E_DIR)/node_modules all ## Run the template integration tests
-	cd $(TEST_E2E_DIR) && dune exec -- yarn test
+	dune build @install && cd $(TEST_E2E_DIR) && dune exec -- yarn test
```
tmattio, line 44: The `all` target (on which this target depends) already runs `dune build @install` so I don't think it's needed here

```suggestion
	cd $(TEST_E2E_DIR) && dune exec -- yarn test
```

tmattio (approved):

Makefile
rgrinberg: The former builds the default target (`@all` since 2.0). The latter builds the `@all` target. I guess we should build the default target.

Note that what we had before is `$ dune build .` which means build all the targets in `.`

Makefile
rgrinberg: Ah yes. But it's weird that `all` actually builds `@install`. I changed `all` to build `@all` and this target should depend on `@install`. Running the tests should only require the installed targets.

PR #330, opened 2020-12-03
Concurrent request handling

rgrinberg:
This PR makes ocamllsp server serve requests concurrently. Almost all requests
are serialized by using merlin, but it does give us the opportunity to read
cancellation notifications.

### DIFF
--- a/lsp-fiber/src/rpc.ml
+++ b/lsp-fiber/src/rpc.ml
@@ -2,6 +2,28 @@ open Import
 open Jsonrpc
 module Session = Jsonrpc_fiber.Make (Fiber_io)
 
+module Reply = struct
+  type 'r t =
+    | Now of ('r, Jsonrpc.Response.Error.t) result
+    | Later of
+        (   (('r, Jsonrpc.Response.Error.t) result -> unit Fiber.t)
+         -> unit Fiber.t)
+
+  let now r = Now r
+
+  let later f = Later f
+
+  let to_jsonrpc t id to_json : Jsonrpc_fiber.Reply.t =
+    let f x =
+      match x with
+      | Error e -> Jsonrpc.Response.error id e
+      | Ok x -> Jsonrpc.Response.ok id (to_json x)
+    in
+    match t with
+    | Now r -> Jsonrpc_fiber.Reply.now (f r)
+    | Later k -> Jsonrpc_fiber.Reply.later (fun send -> k (fun r -> send (f r)))
+end
+
 module State = struct
   type t =
     | Waiting_for_init
@@ -25,8 +47,7 @@ module type S = sig
 
     type 'state on_request =
       { on_request :
-          'a.    'state session -> 'a in_request
-          -> ('a * 'state, Response.Error.t) result Fiber.t
+          'a. 'state session -> 'a in_request -> ('a Reply.t * 'state) Fiber.t
       }
 
     type 'state t
@@ -81,8 +102,6 @@ end)
 (In_request : Request_intf)
 (In_notification : Notification_intf) =
 struct
-  open Jsonrpc_fiber
-
   type 'a out_request = 'a Out_request.t
 
   type 'a in_request = 'a In_request.t
@@ -104,8 +123,7 @@ struct
 
   and 'state on_request =
     { on_request :
-        'a.    'state t -> 'a in_request
-        -> ('a * 'state, Response.Error.t) result Fiber.t
+        'a. 'state t -> 'a in_request -> ('a Reply.t * 'state) Fiber.t
     }
 
   and 'state handler =
@@ -116,8 +134,7 @@ struct
   module Handler = struct
     type nonrec 'state on_request = 'state on_request =
       { on_request :
-          'a.    'state t -> 'a in_request
-          -> ('a * 'state, Response.Error.t) result Fiber.t
+          'a. 'state t -> 'a in_request -> ('a Reply.t * 'state) Fiber.t
       }
 
     type nonrec 'state t = 'state handler =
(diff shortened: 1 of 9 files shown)


lsp-fiber/src/rpc.mli
```diff
@@ -1,6 +1,16 @@
...
 open Jsonrpc
 
+module Reply : sig
+  type 'resp t
+
+  val now : ('r, Jsonrpc.Response.Error.t) result -> 'r t
+
+  val later :
+       ((('r, Jsonrpc.Response.Error.t) result -> unit Fiber.t) -> unit Fiber.t)
+    -> 'r t
```
RosalesJ, line 11: I _think_ I understand `Reply.now`–it's a value that's just returned immediately–but `Reply.later` is a little tough for me to grok. Your local definition of `later` in `ocaml_lsp_server.ml` was a bit enlightening.
```ocaml
Reply.later (fun k ->
    let open Fiber.O in
    let* resp = f state req in
    k resp)
```
I can kind of see that `f` here is "the thing that you want to be done asynchronously", but I'm having trouble pinning down what `k` is and who provides it. I'm guessing that's the secret to how this change makes concurrency possible.

Also `Rpc.Reply` seems similar to `Jsonrpc_fiber.Reply` in its form and handling, I presume it hands off the concurrency concerns down to `Fiber` in a similar way to how `Rpc.Reply` hands them down to `Jsonrpc_fiber.Reply`. Any clarification would be much appreciated 😁

RosalesJ (commented):
I've been perusing this area of the code lately in preparation for attempting to tackle this one. Turns out I would have been out of my depth so I'm glad to see you picked it up.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -512,18 +509,15 @@ let selection_range (state : State.t)
...
-          let open Fiber.O in
-          let+ shapes = Document.dispatch_exn doc command in
-          selection_range_of_shapes x shapes)
-    in
-    Ok (List.filter_opt ranges)
+  let open Fiber.O in
+  let+ ranges =
+    Fiber.sequential_map positions ~f:(fun x ->
+        let command = Query_protocol.Shape (Position.logical x) in
+        let open Fiber.O in
```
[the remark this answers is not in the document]
rgrinberg: It's not. Will get rid of it.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -784,28 +822,6 @@ let on_notification server (notification : Client_notification.t) :
...
-             ~message:"Multiple exceptions" ())
-    in
-    (* TODO: what to do with merlin notifications? *)
-    Logger.with_notifications (ref []) @@ fun () ->
-    Merlin_utils.File_id.with_cache @@ f
-  in
-  let on_request server req =
-    prepare_and_run Jsonrpc.Response.Error.of_exn @@ fun () ->
-    on_request server req
-  in
```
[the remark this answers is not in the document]
rgrinberg: We should be handling this at the jsonrpc layer, so this shouldn't be necessary.

lsp-fiber/src/rpc.mli
rgrinberg: > I can kind of see that f here is "the thing that you want to be done asynchronously", but I'm having trouble pinning down what k is and who provides it. I'm guessing that's the secret to how this change makes concurrency possible.

The secret is in jsonrpc_fiber. To summarize, we spawn a handler fiber to serve the current request. We give the handler's  fiber a continuation `k` that it may use to send the response. Concurrently, we go on to serve the next request.

> Also Rpc.Reply seems similar to Jsonrpc_fiber.Reply in its form and handling, I presume it hands off the concurrency concerns down to Fiber in a similar way to how Rpc.Reply hands them down to Jsonrpc_fiber.Reply

That's exactly it. Rpc.Rply is translated to the corresponding Jsonrpc_fiber.Reply, which goes on to serve the requests concurrently. Note that `now` isn't strictly necessary, as it's fine to send all responses with `later`. I might simplify that later.

rgrinberg:
> I've been perusing this area of the code lately in preparation for attempting to tackle this one. Turns out I would have been out of my depth so I'm glad to see you picked it up.

Np. I realized that I actually needed this to implement cancellation.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -553,140 +547,176 @@ let definition_query (state : State.t) uri position merlin_request =
...
-  | Client_request.WorkspaceSymbol _ -> Fiber.return @@ Ok (None, state)
+    match Document_store.get_opt store uri with
+    | None -> now None
+    | Some doc -> now (Some (Msource.text (Document.source doc))) )
+  | Client_request.DebugEcho params -> now params
+  | Client_request.TextDocumentColor _ -> now []
+  | Client_request.TextDocumentColorPresentation _ -> now []
+  | Client_request.TextDocumentHover req -> later hover req
+  | Client_request.TextDocumentReferences req -> later references req
+  | Client_request.TextDocumentCodeLensResolve codeLens -> now codeLens
```
tmattio, line 618: What's the reason for not running codeLens concurrently?

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -553,140 +547,176 @@ let definition_query (state : State.t) uri position merlin_request =
   let open Fiber.O in
   let+ result = Document.dispatch_exn doc command in
   let result = location_of_merlin_loc uri result in
-  Ok (result, state)
+  Ok result
+
+let highlight (state : State.t)
```
tmattio, line 552: Might be nice to have the same pattern for all of the requests, e.g. extract `prepare_rename`. But outside of the scope of this PR.

lsp-fiber/src/rpc.ml
```diff
@@ -2,6 +2,28 @@ open Import
 open Jsonrpc
 module Session = Jsonrpc_fiber.Make (Fiber_io)
 
+module Reply = struct
+  type 'r t =
+    | Now of ('r, Jsonrpc.Response.Error.t) result
```
tmattio, line 7: Would be nice to extract `('r, Jsonrpc.Response.Error.t) result` in its own type: it comes back regularly, so it would certainly improve readability.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -553,140 +547,176 @@ let definition_query (state : State.t) uri position merlin_request =
...
-    let* doc = Fiber.return (Document_store.get store uri) in
-    Formatter.run rpc doc
-  | Client_request.TextDocumentOnTypeFormatting _ ->
-    Fiber.return @@ Ok (None, state)
-  | Client_request.SelectionRange req -> selection_range state req
+    later
+      (fun _ () ->
+        let open Fiber.Result.O in
+        let uri = Uri.t_of_yojson (`String uri) in
+        let* doc = Fiber.return (Document_store.get store uri) in
```
tmattio, line 707: Now that the requests are run concurrently, do we need to lock the access to the store?

tmattio (approved):

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: This is `TextDocumentCodeLensResolve` rather code lens. We don't do anything interesting on this request anyway (yet)

lsp-fiber/src/rpc.ml
rgrinberg: It can later be moved to `Import.Or_error`

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: Currently, it doesn't impact correctness. The code that accesses the document store never yield. It would certainly be nice to add regardless, but I'm still thinking of the correct way to do it. Independent of this PR, we have lots of harmless imperfections of the form:

```
  let* doc = Fiber.return (Document_store.get store uri) in
  let client_capabilities = client_capabilities state in
  let open Fiber.O in
  let+ symbols = Document_symbol.run client_capabilities doc uri in
```

We are possibly using an out of date document by the time we acquire merlin. We need to delay getting the document until merlin is ready, rather than doing it eagerly. I'm not yet sure of a good api for this.

lsp-fiber/src/rpc.ml
rgrinberg: Actually, I'm thinking of dropping these result types altogether. We already raise in many places, might as well stick to one way of doing things.

PR #331, opened 2020-12-03
Show function documentation for signatureHelp

mnxn:
Depends on https://github.com/rgrinberg/merlin/pull/10

Uses the position information from merlin to query for the documentation of the function being applied. Like hover, it can return either plaintext or markdown-formatted documentation.

Example: 
image

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -378,12 +378,7 @@ let signature_help (state : State.t)
         let typer = Mpipeline.typer_result pipeline in
         let pos = Mpipeline.get_lexing_pos pipeline pos in
         let node = Mtyper.node_at typer pos in
-        let signature =
-          Merlin_analysis.Signature_help.application_signature node ~prefix
-        in
-        match signature with
-        | `Application a -> Some a
-        | `Unknown -> None)
+        Merlin_analysis.Signature_help.application_signature node ~prefix)
   in
   match application_signature with
   | None ->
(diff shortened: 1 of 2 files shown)


rgrinberg (approved):

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -353,41 +365,61 @@ let signature_help (state : State.t)
...
+        let node = Mtyper.node_at typer pos in
+        let signature =
+          Merlin_analysis.Signature_help.application_signature node ~prefix
         in
-        Ok (help, state)
-      | `Unknown ->
-        let help = SignatureHelp.create ~signatures:[] () in
-        Ok (help, state))
+        match signature with
+        | `Application a -> Some a
```
rgrinberg, line 385: I wonder if it makes sense for the merlin to just return an option type. In any case, we can change this we have signature help for things other than functions.

ocaml-lsp-server/test/e2e/__tests__/textDocument-signatureHelp.ts
```diff
@@ -186,20 +208,22 @@ describe_opt("textDocument/completion", () => {
...
+      let _ = create ?ra
     `);
 
-    let items = await querySignatureHelp(Types.Position.create(0, 26));
+    let items = await querySignatureHelp(Types.Position.create(2, 18));
     expect(items).toMatchObject({
       signatures: [
         {
-          label: "Hashtbl.create : ?random:bool -> int -> ('a, 'b) Hashtbl.t",
+          label: "create : ?random:bool -> int -> ('a, 'b) Hashtbl.t",
```
rgrinberg, line 220: Why did the name change here? `Hashtbl.create` seems like a better label.

mnxn: > I see you've changed the position.

Yep, I made the `let create = Hashtbl.create` binding so those tests wouldn't read standard library documentation comments, which could change between releases.

PR #345, opened 2020-12-10
Add some tests for barrier

rgrinberg:
We're not handling errors as well as we should. It should not be possible to
await/signal after close.


tmattio (approved):
Since you're changing `create` to return a reference, it might be useful to add mutexes in `await` and `signal`, but I'm not sure how critical it is.

rgrinberg:
Good call. That also reminds me that we need a mutex for await as well.

PR #348, opened 2020-12-17
Run unit tests only on latest OCaml version

### LINKED ISSUE #337: Test only on 1 version of OCaml
It's too difficult to make expect tests work on multiple versions of OCaml. Let's not bother.

tmattio:
Fixes #337


rgrinberg:
Looks good. I pushed a commit that fixes the formatting. So this should be rebased.

tmattio:
There are a few CI failures, but they are unrelated to this change, I think we can fix them in another one

PR #350, opened 2020-12-17
Support new languageclient format

tmattio:
It seems that `vscode-languageclient.7.0.0` changed the way parameters are send to the LSP servers for custom requests and now send a list of string when sending a string. This broke out custom requests. This PR handles the new format send by VSCode client.

### DIFF
--- a/ocaml-lsp-server/src/custom_requests/req_infer_intf.ml
+++ b/ocaml-lsp-server/src/custom_requests/req_infer_intf.ml
@@ -6,7 +6,8 @@ let meth = "ocamllsp/inferIntf"
 
 let on_request ~(params : Json.t option) (state : State.t) =
   match params with
-  | Some (`String (_ : DocumentUri.t) as json_uri) -> (
+  | Some (`String (_ : DocumentUri.t) as json_uri)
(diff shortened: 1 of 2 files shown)


rgrinberg:
This needs a documentation update as well.

By the way, in the future we can't just change the spec of the impl-intf switch request. This request is part of the public api and we don't want to break editors that rely on it simply because vscode changed something. It's probably fine here as I don't think any of the editors integrated it yet, but we should keep this in mind in the future.

tmattio:
> This needs a documentation update as well.

Actually, I don't think the documentation needs to change: we're doing everything correctly on the VSCode side and send a single string. Somehow, VSCode decides to send a list instead, and we handle that, but the intended API remains unchanged: the user should send a string.

> By the way, in the future we can't just change the spec of the impl-intf switch request.

That's fair. Just to be clear: this change is not a breaking change: the previous API remains unchanged, we simply handle the case where the client sends a list with a single string.

rgrinberg:
Looks like my interpretation of the spec was incorrect and `params` may only be an array or an object.

In any case, we can fix that later once all the clients have update to the latest vscode.

PR #351, opened 2020-12-18
Comply to jsonrpc spec

rgrinberg:
Only objects or arrays are allowed as parameters. We reflect this in the
new `Structured.t` type for this field.

This drops support for older versions of the client. So we can only merge this
once everyone upgrades.

### DIFF
--- a/jsonrpc/src/jsonrpc.ml
+++ b/jsonrpc/src/jsonrpc.ml
@@ -38,10 +38,24 @@ module Constant = struct
 end
 
 module Message = struct
+  module Structured = struct
+    type t =
+      [ `Assoc of (string * Json.t) list
+      | `List of Json.t list
+      ]
+
+    let of_json = function
+      | `Assoc xs -> `Assoc xs
+      | `List xs -> `List xs
+      | json -> Json.error "invalid structured value" json
+
+    let to_json t = (t :> Json.t)
+  end
+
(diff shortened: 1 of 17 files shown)


rgrinberg:
The tests are failing now, but I assume that should be fixed when I update them to use the latest client.

jsonrpc/src/jsonrpc.ml
```diff
@@ -38,10 +38,24 @@ module Constant = struct
 end
 
 module Message = struct
+  module Structured = struct
```
tmattio, line 41: How about `Params` instead to use the same terminology as VSCode?

tmattio (approved):

jsonrpc/src/jsonrpc.ml
rgrinberg: The name `Structured` comes from the jsonrpc [spec](https://www.jsonrpc.org/specification#parameter_structures), so it's more descriptive of the value used.

tmattio:
@rgrinberg I think we can merge this, unless there's anything blocking

rgrinberg:
Feel free to merge if you manage to fix the test suite.

tmattio:
I fixed the failing tests. The CI is still failing because of the usual errors, but that's unrelated to this PR, so I'm merging now.

PR #354, opened 2020-12-23
support eliom file extensions

jrochel:
Can anyone tell me why this doesn't work?
The problem: In my editor, when I edit an `.eliomi` file it is still treated as an `Impl` and not a `Intf` file. Does my editor have to inform ocaml-lsp about the fact that the file should be treated as an interface?

### DIFF
--- a/ocaml-lsp-server/src/document.ml
+++ b/ocaml-lsp-server/src/document.ml
@@ -8,9 +8,11 @@ module Kind = struct
   let of_fname p =
     match Filename.extension p with
     | ".ml"
+    | ".eliom"
     | ".re" ->
       Impl
     | ".mli"
+    | ".eliomi"
     | ".rei" ->
       Intf
     | ext ->
@@ -44,6 +46,8 @@ module Syntax = struct
 
   let of_fname s =
     match Filename.extension s with
+    | ".eliomi"
+    | ".eliom"
     | ".mli"
     | ".ml" ->
       Ocaml
@@ -187,13 +191,14 @@ let get_impl_intf_counterparts uri =
            ~message:"provided file URI (param) doesn't follow URI spec" ())
   in
   let fname = Filename.basename fpath in
-  let ml, mli, re, rei, mll, mly = ("ml", "mli", "re", "rei", "mll", "mly") in
+  let ml, mli, eliom, eliomi, re, rei, mll, mly =
+    ("ml", "mli", "eliom", "eliomi", "re", "rei", "mll", "mly") in
   let exts_to_switch_to =
     match Syntax.of_fname fname with
     | Ocaml -> (
       match Kind.of_fname fname with
-      | Intf -> [ ml; mly; mll; re ]
-      | Impl -> [ mli; mly; mll; rei ] )
+      | Intf -> [ ml; mly; mll; eliom; re ]
+      | Impl -> [ mli; mly; mll; eliomi; rei ] )
     | Reason -> (
       match Kind.of_fname fname with
       | Intf -> [ re; ml ]
--- a/ocaml-lsp-server/src/inference.ml
+++ b/ocaml-lsp-server/src/inference.ml
@@ -21,8 +21,10 @@ let infer_intf_for_impl doc =
 
 let language_id_of_fname s =
   match Filename.extension s with
-  | ".mli" -> "ocaml.interface"
-  | ".ml" -> "ocaml"
+  | ".mli"
+  | ".eliomi" -> "ocaml.interface"
+  | ".ml"
+  | ".eliom" -> "ocaml"
   | ".rei"
   | ".re" ->
     "reason"


tmattio:
You can reuse the `ocaml` and `ocaml.interface` language ids I think, since they are the same syntax:

```diff
diff --git a/ocaml-lsp-server/src/document.ml b/ocaml-lsp-server/src/document.ml
index 2f3245c..fd1441f 100644
--- a/ocaml-lsp-server/src/document.ml
+++ b/ocaml-lsp-server/src/document.ml
@@ -39,8 +39,6 @@ module Syntax = struct
   let all =
     [ ("ocaml.interface", Ocaml)
     ; ("ocaml", Ocaml)
-    ; ("eliom.interface", Ocaml)
-    ; ("eliom", Ocaml)
     ; ("reason", Reason)
     ; ("ocaml.ocamllex", Ocamllex)
     ; ("ocaml.menhir", Menhir)
diff --git a/ocaml-lsp-server/src/inference.ml b/ocaml-lsp-server/src/inference.ml
index 98aa1c9..cc1c150 100644
--- a/ocaml-lsp-server/src/inference.ml
+++ b/ocaml-lsp-server/src/inference.ml
@@ -21,10 +21,12 @@ let infer_intf_for_impl doc =

 let language_id_of_fname s =
   match Filename.extension s with
-  | ".mli" -> "ocaml.interface"
-  | ".ml" -> "ocaml"
-  | ".eliomi" -> "eliom.interface"
-  | ".eliom" -> "ocaml"
+  | ".mli"
+  | ".eliomi" ->
+    "ocaml.interface"
+  | ".ml"
+  | ".eliom" ->
+    "ocaml"
   | ".rei"
   | ".re" ->
     "reason"
```

With the diff above and this PR: https://github.com/ocamllabs/vscode-ocaml-platform/pull/487, `eliom` and `eliomi` files are treated correctly.

It doesn't work without the VSCode PR: the editor won't query the LSP server for every file, it has to know which file should be managed by the server first.

jrochel:
Thanks!

tmattio (approved):
LGTM will merge soon if there are no concerns from other maintainers.

mnxn (approved):

jrochel:
I'm having trouble though, configuring my editor (vim) to send the correct information to ocaml-lsp-server. Everything continues to be treated as Impl files. It's probably the same problem as for VSCode. Could anyone give me high-level description of what the LSP client needs to send to the server in order to correctly handle eliomi files?

tmattio:
Here are the LSP traces for the `textDocument/didOpen` request:

```
[Trace - 11:42:56 AM] Sending notification 'textDocument/didOpen'.
Params: {
    "textDocument": {
        "uri": "file:///Users/tmattio/Workspace/tarides/test-vscode/lib/utils.eliomi",
        "languageId": "ocaml.interface",
        "version": 1,
        "text": "val greet : 'a -> string\n"
    }
}

[Trace - 11:42:59 AM] Sending notification 'textDocument/didOpen'.
Params: {
    "textDocument": {
        "uri": "file:///Users/tmattio/Workspace/tarides/test-vscode/lib/utils.eliom",
        "languageId": "ocaml",
        "version": 1,
        "text": "let greet s = print_endline (\"Hello \" ^ s)\n"
    }
}
```

The important part is the `languageId` field.

jrochel:
Thanks a lot !

jrochel:
@tmattio 
I have the impression that what you said is not quite right.
Even when sending ``"languageId": "ocaml.interface"` my `.eliomi` files are still treated as `Impl` files. When I edit some `.mli` file the file is correctly treated as an `Intf` file, while in the logs I see no occurrence of `ocaml.interface`. Also, in the ocaml-lsp source code the mapping from `"ocaml.interface"` is to `OCaml` and not to `(OCaml, Intf)`. I suspect that there is some magic somewhere that uses the file extension of the `path` to determine whether the file is an `Intf` or an `Impl`, but I don't understand why it only works for `.ml`/`.mli` but not `.eliom`/`.eliomi`.

ocaml-lsp-server/src/document.ml
```diff
@@ -187,13 +191,14 @@ let get_impl_intf_counterparts uri =
...
+  let ml, mli, eliom, eliomi, re, rei, mll, mly =
+    ("ml", "mli", "eliom", "eliomi", "re", "rei", "mll", "mly") in
   let exts_to_switch_to =
     match Syntax.of_fname fname with
     | Ocaml -> (
       match Kind.of_fname fname with
-      | Intf -> [ ml; mly; mll; re ]
-      | Impl -> [ mli; mly; mll; rei ] )
+      | Intf -> [ ml; mly; mll; eliom; re ]
+      | Impl -> [ mli; mly; mll; eliomi; rei ] )
```
rgrinberg, line 201: Pretty sure this is wrong and it's not possible to have a `foo.ml` and `foo.eliomi` pair. It should only be possible to switch from `eliom` to `eliomi` and vice versa. @drup can you confirm?

rgrinberg:
@jrochel that's because we should only be using `ocaml.interface` when the client is vscode. That's actually a separate bug and isn't limited to eliom in any way.

ocaml-lsp-server/src/document.ml
Drup: In theory, it's fine to have any kind of paring. I'm not sure any current build system supports it though.

rgrinberg: So the current makefile/ocamlbuild build systems only support `eliom`/`eliomi`. I suppose we should reflect this until eliom support is added to dune and this can be revised.

Issue #353, opened 2020-12-22
feature request: support eliom file extensions


jrochel:
This would be ".eliom" for OCaml and ".eliomi" for Ocaml interface.
The patch is trivial, but I still can't seem to get it to work.
I hope someone who knows this source code will be able to implement this within 5 seconds.

jordanrmerrick:
Do you mean you want to apply OCaml syntax to `.eliom` and `.eliomi` files? If so, a current workaround can be found [here](https://discuss.ocaml.org/t/why-eliom-has-its-own-file-name-extensions-eliom/1717).

jrochel:
Indeed, that's what I want and what I'm already doing. However, the recognition of whether the current file is an interface file or and implementation file is done in ocaml-lsp-server, if I'm not mistaken.

I did add the file extensions ".eliom" and ".eliomi" to `ocaml-lsp-server/src/document.ml` and to `ocaml-lsp-server/src/inference.ml`, but that didn't seem to do the trick.

jrochel:
https://github.com/ocaml/ocaml-lsp/pull/354

jrochel:
The PR has been merged, but I still cannot get the recognition of .eliomi files to work. 

> Even when sending "languageId": "ocaml.interface" my .eliomi files are still treated as Impl files. When I edit some .mli file the file is correctly treated as an Intf file, while in the logs I see no occurrence of ocaml.interface, but only ever "languageId":"ocaml". Also, in the ocaml-lsp source code the mapping from "ocaml.interface" is to OCaml and not to (OCaml, Intf). I suspect that there is some magic somewhere that uses the file extension of the path to determine whether the file is an Intf or an Impl, but I don't understand why it only works for .ml/.mli but not .eliom/.eliomi.

PR #359, opened 2020-12-30
Support Eliom as a separate syntax

tmattio:
Related to https://github.com/ocaml/ocaml-lsp/issues/353

@jrochel The language IDs are now `ocaml.eliom` and `ocaml.eliom.interface`, not sure this is solving the problem you have, since the previous setup should have been working fine, but let me know!

### DIFF
--- a/ocaml-lsp-server/src/document.ml
+++ b/ocaml-lsp-server/src/document.ml
@@ -29,12 +29,14 @@ module Syntax = struct
     | Reason
     | Ocamllex
     | Menhir
+    | Eliom
 
   let human_name = function
     | Ocaml -> "OCaml"
     | Reason -> "Reason"
     | Ocamllex -> "OCamllex"
     | Menhir -> "Menhir/ocamlyacc"
+    | Eliom -> "Eliom"
 
   let all =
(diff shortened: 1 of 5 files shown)


rgrinberg:
@tmattio this PR slipped my inbox. Is it still relevant? The code looks fine after a brief glance.

smorimoto:
Yes, it's still relevant.

tmattio:
I'll test this again. Was also hoping for a confirmation that this solved the issue by @jrochel, but let's merge once I've confirmed it works

jrochel:
@tmattio The current master version works fine for me, but since #353 I have switched from vim to neovim. Testing with vim things seem to already work fine too.

tmattio:
Great to hear it's working! I think we still want to have a separate syntax name for Eliom, which this PR brings, but it should not change the current behavior (at least not in VSCode with https://github.com/ocamllabs/vscode-ocaml-platform/pull/492 in, but support for the `eliom` syntax might have to be added to vim's client)

rgrinberg:
@tmattio for clients such as vim or emacs, we can guess the languageId based on the extension. That should be more manageable than keeping the clients up to date.

PR #377, opened 2021-02-25
Unlock 4.12.x

smorimoto:
Before merging this, we need to update the Merlin fork.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -2,9 +2,11 @@ name: Build and Test
 
 on:
   push:
-    branches: [ master ]
+    branches:
+      - master
   pull_request:
-    branches: [ master ]
+    branches:
+      - master
 
 jobs:
   build-and-test:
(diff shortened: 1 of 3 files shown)


smorimoto:
@voodoos It seems that something is missing:
https://github.com/ocaml/ocaml-lsp/pull/377/checks?check_run_id=2037932247#step:9:236

voodoos:
This is probably due to recent changes in how we package Merlin: we release separate packages for different OCaml versions:

Branch `master` if for OCaml 4.12 and released as version `4.1-412`.
Branch `411` if for OCaml 4.11 and released as version `4.1-411`.
Branch `3.4` if for all versions <= 4.11 and released as version `3.4.2`.

That means Merlin does not perform build-time source selection any-more (except on legacy branch `3.4`) and this should probably be done on OCaml-LSP's side now. (I think this has been discussed by trefis and @rgrinberg, not sure what is the definitive plan for OCaml-LSP though).

For now OCam-LSP's vendored Merlin is based on the `3.4` branch which does not build under 4.12.

smorimoto:
@voodoos I see. Thanks for the detailed explanation.
@rgrinberg Do you already have a clear spec for this?

PR #386, opened 2021-03-16
OCamlformat RPC

voodoos:
This PR add support for formatting of types in popups. Types provided by Merlin are not always nicely formatted and module signatures often wrap because it exceeds the maximum number of columns of a VSCode popup (64 characters on my setup). The idea is to use OCamlformat to format these types, but starting a new process would be to heavy for that usage.

In a joint effort with the OCamlformat team this PR acheive type formatting by spawning an "OCamlformat RPC server" and then communicating with it using a simple protocol. The protocol is vendored for now (in the file `ocamlformat_rpc_lib.ml` but will be made available as a separate and common package.

In case of failure to start the server, negotiate a version or receive a response we log some information and fallback to the non-formatted type.

## Todo
- [x] Wait for OCaml-RPC lib to be published as a separate package and remove the vendored copy.
- [x] Change the program name to the definitive name of the OCamllsp-RPC server when known. (for now it is a hard path to my copy of OCamlformat).

## Related works
- This [OCamlformat PR](https://github.com/ocaml-ppx/ocamlformat/pull/1586) tracks the progress on the RPC library and the server implementation.
- This [mdx PR](https://github.com/realworldocaml/mdx/pull/316) tracks the progress on implementing a client similar to the one of this PR to format code blocks in mdx.

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -997,135 +997,87 @@ let start () =
     Fdecl.get server
   in
   let dune = ref None in
-  Fiber.fork_and_join_unit
-    (fun () -> Fiber.Pool.run detached)
-    (fun () ->
-      let run_dune () =
-        let* (init_params : InitializeParams.t) = Server.initialized server in
-        let progress =
-          Progress.create init_params.capabilities
-            ~report_progress:(fun progress ->
-              Server.notification server
-                (Server_notification.WorkDoneProgress progress))
-            ~create_task:(fun task ->
-              Server.request server (Server_request.WorkDoneProgressCreate task))
-        in
-        let dune =
-          let dune' = Dune.create diagnostics progress in
-          dune := Some dune';
-          dune'
-        in
-        let* state = Dune.run dune in
-        let message =
-          match state with
-          | Error Binary_not_found ->
-            Some "Dune must be installed for project functionality"
-          | Error Out_of_date ->
-            Some
-              "Dune is out of date. Install dune >= 3.0 for project \
-               functionality"
-          | Ok () -> None
-        in
-        match message with
-        | None -> Fiber.return ()
-        (* We disable the warnings because dune 3.0 isn't available yet *)
-        | Some _ when true -> Fiber.return ()
-        | Some message ->
-          let* (_ : InitializeParams.t) = Server.initialized server in
-          let state = Server.state server in
-          task_if_running state ~f:(fun () ->
-              let log = LogMessageParams.create ~type_:Warning ~message in
-              Server.notification server (Server_notification.LogMessage log))
-      in
-      let* () =
-        Fiber.parallel_iter
-          ~f:(fun f -> f ())
-          [ (fun () ->
-              if enable_dune_rpc then
-                run_dune ()
-              else
-                Fiber.return ())
-          ; (fun () ->
-              let* () = Server.start server in
-              (* When server exits we also stop ocamlformat rpc *)
-              Ocamlformat_rpc.stop ocamlformat_rpc)
-          ; (fun () ->
-              let logger ~type_ ~message () =
-                task_if_running (Server.state server) ~f:(fun () ->
-                    let log = LogMessageParams.create ~type_ ~message in
-                    Server.notification server
-                      (Server_notification.LogMessage log))
-              in
-              let* state = Ocamlformat_rpc.run ~logger ocamlformat_rpc in
-              let message =
-                match state with
-                | Error `Binary_not_found ->
-                  Some
-                    "OCamlformat_rpc is missing, displayed types might not be \
-                     properly formatted. "
-                | Ok () -> None
-              in
-              match message with
-              | None -> Fiber.return ()
-              | Some message ->
-                let* (_ : InitializeParams.t) = Server.initialized server in
-                let state = Server.state server in
-                task_if_running state ~f:(fun () ->
-                    let log = ShowMessageParams.create ~type_:Info ~message in
-                    Server.notification server
-                      (Server_notification.ShowMessage log)))
-          ; (fun () ->
-              let* (init_params : InitializeParams.t) =
-                Server.initialized server
-              in
-              let progress =
-                Progress.create init_params.capabilities
-                  ~report_progress:(fun progress ->
-                    Server.notification server
-                      (Server_notification.WorkDoneProgress progress))
-                  ~create_task:(fun task ->
-                    Server.request server
-                      (Server_request.WorkDoneProgressCreate task))
-              in
-              let dune =
-                let dune' = Dune.create diagnostics progress in
-                dune := Some dune';
-                dune'
-              in
-              let* state = Dune.run dune in
-              let message =
-                match state with
-                | Error Binary_not_found ->
-                  Some "Dune must be installed for project functionality"
-                | Error Out_of_date ->
-                  Some
-                    "Dune is out of date. Install dune >= 3.0 for project \
-                     functionality"
-                | Ok () -> None
-              in
-              match message with
-              | None -> Fiber.return ()
-              (* We disable the warnings because dune 3.0 isn't available yet *)
-              | Some _ when true -> Fiber.return ()
-              | Some message ->
-                let* (_ : InitializeParams.t) = Server.initialized server in
-                let state = Server.state server in
-                task_if_running state ~f:(fun () ->
-                    let log = LogMessageParams.create ~type_:Warning ~message in
-                    Server.notification server
-                      (Server_notification.LogMessage log)))
-          ]
-      in
-      let* () =
+  let run_dune () =
+    let* (init_params : InitializeParams.t) = Server.initialized server in
+    let progress =
+      Progress.create init_params.capabilities
+        ~report_progress:(fun progress ->
+          Server.notification server
+            (Server_notification.WorkDoneProgress progress))
+        ~create_task:(fun task ->
+          Server.request server (Server_request.WorkDoneProgressCreate task))
+    in
+    let dune =
+      let dune' = Dune.create diagnostics progress in
+      dune := Some dune';
+      dune'
+    in
+    let* state = Dune.run dune in
+    let message =
+      match state with
+      | Error Binary_not_found ->
+        Some "Dune must be installed for project functionality"
+      | Error Out_of_date ->
+        Some
+          "Dune is out of date. Install dune >= 3.0 for project functionality"
+      | Ok () -> None
+    in
+    match message with
+    | None -> Fiber.return ()
+    (* We disable the warnings because dune 3.0 isn't available yet *)
+    | Some _ when true -> Fiber.return ()
+    | Some message ->
+      let* (_ : InitializeParams.t) = Server.initialized server in
+      let state = Server.state server in
+      task_if_running state ~f:(fun () ->
+          let log = LogMessageParams.create ~type_:Warning ~message in
+          Server.notification server (Server_notification.LogMessage log))
+  in
+  Fiber.parallel_iter
+    ~f:(fun f -> f ())
+    [ (fun () ->
+        if enable_dune_rpc then
+          run_dune ()
+        else
+          Fiber.return ())
+    ; (fun () -> Fiber.Pool.run detached)
+    ; (fun () ->
+        let* () = Server.start server in
+        (* When server exits we also stop ocamlformat rpc *)
         Fiber.parallel_iter
           ~f:(fun f -> f ())
           [ (fun () -> Document_store.close store)
           ; (fun () -> Fiber.Pool.stop detached)
-          ]
-      in
-      match !dune with
-      | None -> Fiber.return ()
-      | Some dune -> Dune.stop dune)
+          ; (fun () -> Ocamlformat_rpc.stop ocamlformat_rpc)
+          ; (fun () ->
+              match !dune with
+              | None -> Fiber.return ()
+              | Some dune -> Dune.stop dune)
+          ])
+    ; (fun () ->
+        let logger ~type_ ~message () =
+          task_if_running (Server.state server) ~f:(fun () ->
+              let log = LogMessageParams.create ~type_ ~message in
+              Server.notification server (Server_notification.LogMessage log))
+        in
+        let* state = Ocamlformat_rpc.run ~logger ocamlformat_rpc in
+        let message =
+          match state with
+          | Error `Binary_not_found ->
+            Some
+              "OCamlformat_rpc is missing, displayed types might not be \
+               properly formatted. "
+          | Ok () -> None
+        in
+        match message with
+        | None -> Fiber.return ()
+        | Some message ->
+          let* (_ : InitializeParams.t) = Server.initialized server in
+          let state = Server.state server in
+          task_if_running state ~f:(fun () ->
+              let log = ShowMessageParams.create ~type_:Info ~message in
+              Server.notification server (Server_notification.ShowMessage log)))
+    ]
 
 let run () =
   Unix.putenv "__MERLIN_MASTER_PID" (string_of_int (Unix.getpid ()));
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -346,8 +346,8 @@ let query_type doc pos =
     None
   | (location, `String value, _) :: _ -> Some (location, value)
 
-let hover (state : State.t) { HoverParams.textDocument = { uri }; position; _ }
-    =
+let hover server (state : State.t)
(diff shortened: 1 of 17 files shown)


ocaml-lsp-server/src/ocamlformat_rpc_lib.ml
```diff
@@ -0,0 +1,220 @@
+(**************************************************************************)
+(*                                                                        *)
+(*                              OCamlFormat                               *)
+(*                                                                        *)
+(*            Copyright (c) Facebook, Inc. and its affiliates.            *)
+(*                                                                        *)
+(*      This source code is licensed under the MIT license found in       *)
+(*      the LICENSE file in the root directory of this source tree.       *)
+(*                                                                        *)
+(**************************************************************************)
```
rgrinberg, line 10: If this is vendored code, it should live in the vendored directory.

ocaml-lsp-server/src/ocamlformat_rpc.ml
```diff
@@ -0,0 +1,69 @@
...
+  ; output : out_channel
+  ; client : Ocamlformat_rpc_lib.client
+  }
+
+type state =
+  | Uninitialized
+  | Running of process
+  | Errored
+
+let running_process : state ref = ref Uninitialized
```
rgrinberg, line 16: This state should exist inside the LSP server instance.

ocaml-lsp-server/src/ocamlformat_rpc.ml
```diff
@@ -0,0 +1,69 @@
...
+let start () =
+  let prog =
+    "/Users/ulysse/tmp/ocamlformat/_build/default/bin/ocamlformat-rpc/main.exe"
+  in
+  let argv = [| "main.exe" |] in
+  let input, output = Unix.open_process_args prog argv in
+  let pid = Unix.process_pid (input, output) in
+  let open Result.O in
+  Ocamlformat_rpc_lib.pick_client ~pid input output supported_versions
+  >>| (fun client ->
```
rgrinberg, line 29: `let+`

ocaml-lsp-server/src/ocamlformat_rpc_lib.ml
voodoos: It should disappear before we merge the PR. But I will move it to the vendor dir if it doesn't at the end.

ocaml-lsp-server/src/ocamlformat_rpc.ml
voodoos: You mean that it should be a part of `State.t` ?

rgrinberg: Yes.

ocaml-lsp-server/src/ocamlformat_rpc_lib.ml
rgrinberg: Sounds good.

ocaml-lsp-server/src/ocamlformat_rpc_lib.ml
```diff
@@ -0,0 +1,220 @@
...
+      | `Version v when v = latest -> get_client_exn ~pid input output v
+      | `Version v -> (
+        match others with
+        | h :: _ when v = h -> get_client_exn ~pid input output v
+        | _ -> aux others )
+      | `Unknown -> aux others
+      | `Halt ->
+        Error
+          (`Msg
+            "OCamlformat-RPC did not respond. Check that a compatible version \
```
rgrinberg, line 205: We should probably mentioned what minimum version is required.

voodoos: cc @gpetiot

gpetiot: done

gpetiot:
`ocamlformat-rpc-lib` is now available on opam!

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -336,7 +336,20 @@ let hover (state : State.t) { HoverParams.textDocument = { uri }; position; _ }
...
+    and+ typ =
+      (* We ask Ocamlformat to format this type *)
+      let+ result = Ocamlformat_rpc.query_type state.ocamlformat_rpc typ in
+      match result with
+      | Ok v -> v
+      | Error `No_process -> typ
+      | Error (`Msg message) ->
+        (* We log OCamlformat errors and display the unformated type *)
+        Log.log ~section:"ocamlformat" (fun () ->
+            Log.msg "An error occured while querying ocamlformat"
```
rgrinberg, line 349: Note that this log is mostly used for debugging. Users will not see these messages unless they open the debug pane. If you want users to see this error, use the `ShowMessage` notification.

ocaml-lsp-server/src/ocamlformat_rpc.ml
```diff
@@ -0,0 +1,172 @@
...
+     editors that use the server. *)
+  match
+    Scheduler.async io_thread (fun () ->
+        match
+          Ocamlformat_rpc_lib.config
+            [ ("module-item-spacing", "compact"); ("margin", "64") ]
+            client
+        with
+        | Error (`Msg msg) ->
+          Log.log ~section:"ocamlformat" (fun () ->
```
rgrinberg, line 48: I'd move this back to the main thread. it's better to write log messages from a single thread if possible.

ocaml-lsp-server/src/ocamlformat_rpc.ml
```diff
@@ -0,0 +1,172 @@
...
+    Fiber.return @@ Error `No_process
+
+let get_process t =
+  let open Fiber.O in
+  match !t with
+  | Stopped -> start t
+  | Running ({ pid; io_thread; _ } as p) -> (
+    (* Is the process still running ? *)
+    match
+      Scheduler.async io_thread (fun () -> Unix.waitpid [ WNOHANG ] pid |> fst)
```
rgrinberg, line 127: You should use `Scheduler.wait_for_process` here. We do the waiting for all processes in a special thread.

ocaml-lsp-server/src/ocamlformat_rpc.mli
```diff
@@ -0,0 +1,7 @@
+type t
+
+val create : unit -> t
+
+val stop : t -> unit Fiber.t
+
+val query_type : t -> string -> (string, [> `Msg of string ]) result Fiber.t
```
rgrinberg, line 7: `string` should be labelled.

ocaml-lsp-server/src/ocamlformat_rpc.ml
```diff
@@ -0,0 +1,172 @@
+open Import
+
+type process =
+  { pid : int
```
rgrinberg, line 4: I'd prefer `Stdune.Pid.t` here.

ocaml-lsp-server/src/ocamlformat_rpc.ml
```diff
@@ -0,0 +1,172 @@
...
+        match
+          Ocamlformat_rpc_lib.config
+            [ ("module-item-spacing", "compact"); ("margin", "64") ]
+            client
+        with
+        | Error (`Msg msg) ->
+          Log.log ~section:"ocamlformat" (fun () ->
+              Log.msg "An error occured while configuring ocamlformat"
+                [ ("msg", `String msg) ])
+        | Ok () -> ())
```
rgrinberg, line 51: This short clause should be first.

ocaml-lsp-server/src/ocamlformat_rpc.ml
```diff
@@ -0,0 +1,172 @@
...
+      let process = { pid; input; output; io_thread; client } in
+      let+ () = configure process in
+      Log.log ~section:"ocamlformat" (fun () ->
+          Log.msg "Ocamlformat-RPC server started" [ ("pid", `Int pid) ]);
+      t := Running process;
+      Ok process)
+
+let start ?io_thread t =
+  let open Fiber.O in
+  try
```
rgrinberg, line 103: This `try` is suspicious. It will not catch exceptions raised by fibers. You're probably looking for `Fiber.map_reduce_errors`.

ocaml-lsp-server/src/ocamlformat_rpc.ml
```diff
@@ -0,0 +1,172 @@
...
+      let stdout_i, stdout_o = Unix.pipe () in
+      let pid = Unix.create_process bin argv stdin_i stdout_o Unix.stderr in
+      (pid, stdout_i, stdin_o)
+    in
+    let open Fiber.O in
+    let input = Unix.in_channel_of_descr stdout in
+    let output = Unix.out_channel_of_descr stdin in
+    let* io_thread =
+      (* Don't create a new thread if there is already one. *)
+      (* This can happen if the Ocamlformat server crashed. *)
```
rgrinberg, line 80: I'd prefer not to reuse the threads between processes unless there's a concrete performance improvement that is totally necessary. The issue is that we might be stuck arbitrarily long waiting for existing tasks to complete. If we create a new thread, we'll always get new responses promptly.

ocaml-lsp-server/src/ocamlformat_rpc.ml
```diff
@@ -0,0 +1,172 @@
...
+  match Bin.which prog with
+  | None ->
+    t := Binary_not_found;
+    Fiber.return @@ Error `No_process
+  | Some bin_path -> (
+    let bin = Fpath.to_string bin_path in
+    let argv = [| bin |] in
+    let pid, stdout, stdin =
+      let stdin_i, stdin_o = Unix.pipe () in
+      let stdout_i, stdout_o = Unix.pipe () in
```
rgrinberg, line 71: I don't see these fd's being closed anywhere. That seems wrong.

ocaml-lsp-server/src/ocamlformat_rpc.ml
```diff
@@ -0,0 +1,172 @@
...
+  let* p = get_process t in
+  match p with
+  | Error `No_process -> Fiber.return @@ Error (`Msg "No process")
+  | Error (`Msg msg) -> Fiber.return @@ Error (`Msg msg)
+  | Ok process -> (
+    match
+      Scheduler.async process.io_thread (fun () ->
+          Ocamlformat_rpc_lib.format typ process.client)
+    with
+    | Error `Stopped -> Fiber.return @@ Error (`Msg "Thread stopped")
```
rgrinberg, line 154: Is this error message going to be visible to the user? If so,

* Under what circumstances should the user expect this?
* The error message should be more descriptive as to what is happening.

Same for the `"No process"` message.

ocaml-lsp-server/src/ocamlformat_rpc.ml
voodoos: I tried, but it raise the following exception if I use it twice:
```
[Trace - 10:37:25 AM] Received response 'textDocument/hover - (10)' in 2ms. Request failed: uncaught exception (-32603).
Error data: {
    "exn": "File \"fiber-unix/src/scheduler.ml\", line 418, characters 28-34: Assertion failed",
    "backtrace": 
"Raised at Fiber_unix__Scheduler.Process_watcher.Process_table.add in file \"fiber-unix/src/scheduler.ml\", line 418, characters 28-40
Called from Fiber_unix__Scheduler.Process_watcher.register in file \"fiber-unix/src/scheduler.ml\", line 441, characters 4-31
Called from Fiber_unix__Scheduler.wait_for_process in file \"fiber-unix/src/scheduler.ml\", line 558, characters 2-71
Called from Fiber.Execution_context.run_jobs in file \"submodules/dune/src/fiber/fiber.ml\", line 204, characters 8-13\n"
}```

It looks like `Swheduler.waitpid` always tries to add the process to the watchlist `Table` and that fails when it is already in the table.

Comment (voodoos) [reply ocaml-lsp-server/src/ocamlformat_rpc.ml:127]: Hum I think I see it: it also uses `Unix.waitpid` which is blocking. So in fact I should not try to use it like the non-blocking `Unix.waitpid [ WNOHANG ]`.

I should not call `Scheduler.wait_for_process` each time I want to query ocf to check that it is still running.
I should use `Scheduler.wait_for_process` as a watcher that will do something if ocf stops running.

Is that right ?

File: ocaml-lsp-server/src/ocamlformat_rpc.mli (hunk 8 above)
Comment (voodoos) [reply ocaml-lsp-server/src/ocamlformat_rpc.mli]: I used a type alias to also make the return type clearer:

`val query_type : t -> type_ -> (type_, [> `Msg of string ]) result Fiber.t`

**Comment by voodoos** [comment]
About error messages, I think most of them will stay as debug messages.
The only user-facing message should be an information  diagnostic showing when OCaml-LSP cannot find the `ocamlformat-rpc` binary and the user should install it for full functionality.

OCamlformat crashes and errors should be kept in the debug log: there are useful for devs to debug issues but useless to the users and unrelated to their programs.

File: ocaml-lsp-server/src/ocamlformat_rpc.ml (hunk 7 above)
Comment (rgrinberg) [reply ocaml-lsp-server/src/ocamlformat_rpc.ml:129]: dune.ml should give you an example of how it should work. Basically, when your RPC is in the running state, it should be roughly this:

```ocaml
Fiber.fork_and_join_unit
  (fun () -> (* code to interact with ocamlformat rpc *))
  (fun () -> Scheduler.wait_for_process pid)
```

Additionally, there should be a `Ocamlformat_rpc.run` function that will return a fiber that should complete when RPC is completely finished and everything is cleaned up.

Let me know if that's not clear and we can walk through this in a call.

voodoos:
I used the same model as for the Dune RPC.
I start the OCamllsp_rpc server and wait for it to finish in `Ocaml_lsp_server.start`:

```ocaml
Fiber.parallel_iter
          ~f:(fun f -> f ())
          [ (fun () -> Server.start server)
          ; (fun () ->
              let* _state = Ocamlformat_rpc.run ocamlformat_rpc in
              (* TODO ulysse Handle errors*)
              Fiber.return ())
          ; (fun () -> ...  Dune.run dune  ... )
          ]
```

However I have a problem to stop it: you appear to stop the DRPC client after this first fiber returns:
```ocaml
      match !dune with
      | None -> Fiber.return ()
      | Some dune -> Dune.stop dune)
```

But I don't understand why the first fiber would return before the process is stopped because it is waiting for the process to stop to complete ? In any case I had that problem with OCamlformat server and decided to stop it explicitely when the Server does:

```ocaml
Fiber.parallel_iter
          ~f:(fun f -> f ())
          [ (fun () ->
              let* () = Server.start server in
              (* When server exits we also stop ocamlformat rpc *)
              Ocamlformat_rpc.stop ocamlformat_rpc)
          ; (fun () ->
              let* _state = Ocamlformat_rpc.run ocamlformat_rpc in
              (* TODO ulysse Handle errors*)
              Fiber.return ())
          ; (fun () -> ...  Dune.run dune  ... )
          ]
```

This way the second item in the parallel iter does complete because it watches the process exit and the process is exited by the stopping of the server in the first item of the parallel iter.

Is that the right thing to do or did I miss something essential ?

(Also when I use Dune master the ocamllsp process does not exits when asking a reload of the server in vscode, process are multiplying , which does not happen with a version of Dune without the rpc server)

rgrinberg:
> Is that the right thing to do or did I miss something essential ?

You aren't missing anything.  I already had this fixed in a branch that I merged yesterday.

voodoos:
> > Is that the right thing to do or did I miss something essential ?
> 
> You aren't missing anything. I already had this fixed in a branch that I merged yesterday.

Great, I will have one last pass and it will be ready for a (hopefully) final review.

voodoos:
I refactored the module a bit.

I also added a user notification (`ShowNotification`) showing when the `ocamlformat-rpc` binary is not found.
However I do not show other errors, they are logged only if debug is enabled.

Some types will fail to be formatted by ocamlformat: Merlin prints types which may be disambiguated with a stamp like in `t/1 -> t/2` and this is not valid syntax.
In this case a non-formatted type will be showed to the user as before.

@gpetiot is working on supporting this syntax in ocamlformat.

rgrinberg:
Looks much better now! There are still some opaque error messages that need fixing (or removing)

voodoos:
I reworked the login.

Also, is it expected that tests fail ? They do not happen on master, but I don't understand how the failures could be related to my changes.

rgrinberg:
> running the tests spawns a huge number of hard-to-kill ocamlformat-rpc instances, I will have to investigate that. Tests also fail on CI because ocamlformat-rpc is not installed.

That's definitely problematic. I forgot to mention that we should delay launching the process until we actually request it to format something.

voodoos:
I decided to use two `Ivars` in ad5e84792384e2c12bedc722ba4a0a0a37e976a9 to only start of the process when the first query is done and then perform that query as soon as the process started.

rgrinberg:
Looks good. I modified the code to us two ivars instead of the mvar. I'm more confident with an ivar that we won't double write when we shouldn't.

PR #389, opened 2021-03-21
Remove ocaml-language-server

ShalokShalom:
Project is archived and last commit happened Nov 2019, so no competitor anymore, and no valid alternative.


smorimoto:
This list is not a list of alternative implementations, but a just comparison with other implementations as the heading says. Therefore, there is no need to remove it.

PR #397, opened 2021-03-31
Add code action annotate

### LINKED ISSUE #396: Feature request: add "annotate value with type" code action
I'd like to have a code action that would turn something like `let foo = 1 in` into `let (foo: int) = 1 in` when called on `foo`.

Khady:
New code action `annotate`, that adds a type next to the value on which the action is called. The goal is to be able to turn `let x = x + 1` into `let (x : int) = x + 1`.

I'm missing a few things before the PR is ready.

1. [x] I don't know how to get in the file what is in the original range of text
1. [x] I need to get the name of the type only and not its detail, for example in this snippet I should get `let f (x : t) = (x : t)` and not `let f (x : t) = (x : type t = Foo of int | Bar of bool)`
  ```
type t = Foo of int | Bar of bool

let f (x : t) = x
  ```

fix #396

### DIFF
--- a/ocaml-lsp-server/src/code_actions/action_annotate.ml
+++ b/ocaml-lsp-server/src/code_actions/action_annotate.ml
@@ -31,7 +31,15 @@ let code_action doc (params : CodeActionParams.t) =
     Query_protocol.Type_enclosing (None, start, None)
   in
   let open Fiber.O in
-  let+ res = Document.dispatch doc command in
+  let+ res =
+    Document.with_pipeline doc (fun pipeline ->
+        let config = Mpipeline.final_config pipeline in
+        let config =
+          { config with query = { config.query with verbosity = 0 } }
+        in
+        let pipeline = Mpipeline.make config (Document.source doc) in
+        Query_commands.dispatch pipeline command)
+  in
   match res with
   | Ok []
   | Ok ((_, `Index _, _) :: _) ->
--- a/ocaml-lsp-server/src/code_actions/action_annotate.ml
+++ b/ocaml-lsp-server/src/code_actions/action_annotate.ml
@@ -2,6 +2,17 @@ open Import
 
 let action_kind = "annotate"
 
+let check_context doc pos_start =
+  Document.with_pipeline doc (fun pipeline ->
+      let pos_start = Mpipeline.get_lexing_pos pipeline pos_start in
+      let typer = Mpipeline.typer_result pipeline in
+      let browse = Mbrowse.of_typedtree (Mtyper.get_typedtree typer) in
+      match Mbrowse.enclosing pos_start [ browse ] with
+      | (_, (Expression _ | Pattern _)) :: _ -> `Valid
+      | _ :: _
+      | [] ->
+        `Invalid)
+
 let get_source_text doc (loc : Loc.t) =
   let open Option.O in
   let source = Document.source doc in
@@ -25,25 +36,28 @@ let code_action_of_type_enclosing uri doc (loc, typ) =
     ~isPreferred:false ()
 
 let code_action doc (params : CodeActionParams.t) =
-  let uri = Uri.t_of_yojson (`String params.textDocument.uri) in
-  let command =
-    let start = Position.logical params.range.start in
-    Query_protocol.Type_enclosing (None, start, None)
-  in
   let open Fiber.O in
-  let+ res =
-    Document.with_pipeline doc (fun pipeline ->
-        let config = Mpipeline.final_config pipeline in
-        let config =
-          { config with query = { config.query with verbosity = 0 } }
-        in
-        let pipeline = Mpipeline.make config (Document.source doc) in
-        Query_commands.dispatch pipeline command)
-  in
-  match res with
-  | Ok []
-  | Ok ((_, `Index _, _) :: _) ->
-    Ok None
-  | Ok ((location, `String value, _) :: _) ->
-    Ok (code_action_of_type_enclosing uri doc (location, value))
-  | Error e -> Error (Jsonrpc.Response.Error.of_exn e)
+  let uri = Uri.t_of_yojson (`String params.textDocument.uri) in
+  let pos_start = Position.logical params.range.start in
+  let* context = check_context doc pos_start in
+  match context with
+  | Error e -> Fiber.return (Error (Jsonrpc.Response.Error.of_exn e))
+  | Ok `Invalid -> Fiber.return (Ok None)
+  | Ok `Valid -> (
+    let command = Query_protocol.Type_enclosing (None, pos_start, None) in
+    let+ res =
+      Document.with_pipeline doc (fun pipeline ->
+          let config = Mpipeline.final_config pipeline in
+          let config =
+            { config with query = { config.query with verbosity = 0 } }
+          in
+          let pipeline = Mpipeline.make config (Document.source doc) in
+          Query_commands.dispatch pipeline command)
+    in
+    match res with
+    | Ok []
+    | Ok ((_, `Index _, _) :: _) ->
+      Ok None
+    | Ok ((location, `String value, _) :: _) ->
+      Ok (code_action_of_type_enclosing uri doc (location, value))
+    | Error e -> Error (Jsonrpc.Response.Error.of_exn e))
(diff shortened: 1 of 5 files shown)


ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
```diff
@@ -66,6 +66,30 @@ let f (x : t) = x
         kind: "destruct",
         title: "Destruct",
       },
+      {
+        edit: {
+          changes: {
+            "file:///test.ml": [
+              {
+                newText: "(x : type t = Foo of int | Bar of bool)",
```
Khady, line 74: here is the place where the type is not displayed as it should

rgrinberg: Yes, you should only enable this code actions on expressions.

ulugbekna: Would querying the type of an expression using mconfig verbosity = 0 (see `ocaml_lsp_server/src/document.ml:136`) help here?

Then we go from 

```ocaml
let (not_supported : ('a, Jsonrpc.Response.Error.t) result

type ('a, 'b) result = Ok of 'a | Error of 'b) =
  Error
    (make_error ~code:InternalError ~message:"Request not supported yet!" ())
```

to correct

```ocaml
let (not_supported : ('a, Jsonrpc.Response.Error.t) result) =
  Error
    (make_error ~code:InternalError ~message:"Request not supported yet!" ())
```

Khady: Thanks, I found a way to reduce the verbosity and it seems to help

ocaml-lsp-server/src/code_actions/action_annotate.ml
```diff
@@ -0,0 +1,49 @@
...
+let code_action doc (params : CodeActionParams.t) =
+  let uri = Uri.t_of_yojson (`String params.textDocument.uri) in
+  let command =
+    let start = Position.logical params.range.start in
+    Query_protocol.Type_enclosing (None, start, None)
+  in
+  let open Fiber.O in
+  let+ res =
+    Document.with_pipeline doc (fun pipeline ->
+        let config = Mpipeline.final_config pipeline in
```
Khady, line 36: I don't know if I am using the right function to get the config

ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
```diff
@@ -143,4 +167,169 @@ let f (x : t) = x
...
+type t = Foo of int | Bar of bool
+
+let f x = Foo x
+`,
+      "file:///test.ml",
+    );
+    let start = Types.Position.create(2, 6);
+    let end = Types.Position.create(2, 7);
+    let actions = await codeAction("file:///test.ml", start, end);
+    expect(actions).toMatchObject([
```
rgrinberg, line 183: Can you filter this list to only show the code action you're interested in?

ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
```diff
@@ -143,4 +167,169 @@ let f (x : t) = x
...
+                    character: 15,
+                    line: 3,
+                  },
+                },
+              },
+            ],
+          },
+        },
+        kind: "destruct",
+        title: "Destruct",
```
rgrinberg, line 307: Same here. We don't need to see destruct here.

rgrinberg:
What happens when you call this code action on this snippet (where `[ ]` is the cursor)?

```ocaml
type x =
   | Foo of int
   | [B]az of string
```

in other words, you try to execute the code action in a non expression context?

ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
Khady: Solved. The diff displayed is less actionable (everything is on a single line), but it's not dramatic

Khady:
I've added a test to show the behavior in a non expression context. For now it would change `Baz` into `(Baz : string -> x)`. I don't know how to check the context to forbid the annotation in this case.

rgrinberg:
>  I don't know how to check the context to forbid the annotation in this case.

You can refer to how destruct works. We make sure that it works only expressions. This might require some changes to our merlin fork, but that's fine.

Khady:
I haven't found where to do that. Could you give me a pointer?

While digging I found this [`type node`](https://github.com/rgrinberg/merlin/blob/f6954e953b4168e6a798a0255d6a2dfbd868a3c6/src/ocaml/merlin_specific/browse_raw.mli#L49-L98) which I believe is pretty close to what we would want to filter on. But didn't completely figured how to get it yet.

rgrinberg:
You can obtain it with:

```
    let browse = Mbrowse.of_typedtree (Mtyper.get_typedtree typer) in
    let nodes = Mbrowse.enclosing pos_start [browse] in
```

Then inspect the first node to see if its an expression.

Khady:
Thanks. I think that I got it working.

ocaml-lsp-server/src/code_actions/action_annotate.ml
```diff
@@ -0,0 +1,63 @@
...
+        `Invalid)
+
+let get_source_text doc (loc : Loc.t) =
+  let open Option.O in
+  let source = Document.source doc in
+  let* start = Position.of_lexical_position loc.loc_start in
+  let+ end_ = Position.of_lexical_position loc.loc_end in
+  let (`Offset start) = Msource.get_offset source (Position.logical start) in
+  let (`Offset end_) = Msource.get_offset source (Position.logical end_) in
+  String.sub (Msource.text source) ~pos:start ~len:(end_ - start)
```
rgrinberg, line 23: You can use `String.drop` here.

ocaml-lsp-server/src/code_actions/action_annotate.ml
```diff
@@ -0,0 +1,63 @@
+open Import
+
+let action_kind = "annotate"
+
+let check_context doc pos_start =
```
rgrinberg, line 5: I'd choose a name that's closer to the property being checked. `check_typeable_context` or something.

ocaml-lsp-server/src/code_actions/action_annotate.ml
```diff
@@ -0,0 +1,63 @@
...
+  let uri = Uri.t_of_yojson (`String params.textDocument.uri) in
+  let pos_start = Position.logical params.range.start in
+  let* context = check_context doc pos_start in
+  match context with
+  | Error e -> Fiber.return (Error (Jsonrpc.Response.Error.of_exn e))
+  | Ok `Invalid -> Fiber.return (Ok None)
+  | Ok `Valid -> (
+    let command = Query_protocol.Type_enclosing (None, pos_start, None) in
+    let+ res =
+      Document.with_pipeline doc (fun pipeline ->
```
rgrinberg, line 49: The execution of this code action makes two calls to `with_pipeline`. That is slightly inefficient, but there's a bit of a bigger problem - when you call `with_pipeline`, you're yielding until the merlin thread is ready. So it's possible that there might be other requests served in the meantime that will invalidate your check for expression context. It should be possible to refactor the code to only call `with_pipeline` once.

ocaml-lsp-server/src/code_actions/action_annotate.ml
Khady: I don't think so, drop is only dropping the beginning of the string while I want to drop at the beginning and at the end?

rgrinberg: Nvm, you're right.

rgrinberg:
Oh, and we need a CHANGES entry for this.

Khady:
Thanks for being so reactive. I moved to a single with_pipeline and added a line to CHANGES.

rgrinberg:
Thanks for your contribution.

Issue #379, opened 2021-02-27
Support for workspace/symbol request


leungbk:
The old `ocaml-language-server` written in TypeScript seems to support the `workspace/symbol` request. It would be nice if this server also supported it.

tatchi:
Happy to give this one a shot :) I don't have any experience with lsp related stuff so it might take me a while to figure out the different pieces and how they interact together. But let's try it 💪

PR #398, opened 2021-04-03
implement workspace/symbol

tatchi:
Opening this one as a draft to get early feedback :) So this is my attempt at implementing https://github.com/ocaml/ocaml-lsp/issues/379.

It uses [ocp-index](https://github.com/OCamlPro/ocp-index) which does pretty much all the work. That's nice because it avoids us to do all this work (reading the `typedtree` from `.cmt(i)` files), but has some drawbacks (hence why I'm opening this PR to get your feedback):

* Merlin is vendoring some part of `compiler-libs`, and `ocp-index` is using `compiler-libs` as well. There were some namespace clashing and therefore I had to `unwrap` some modules that Merlin is vendoring. Diff is here: https://github.com/rgrinberg/merlin/compare/lsp...tatchi:unwrap Note that I'm definitely not sure about the naming I've used   😅
* From my understanding, the `complete` function from the `ocp-index` lib that's used to find the matching symbols is "limited" in the sense that:

 1. It uses a `prefix-first` strategy. So if we search for a function named `add_file`, and the query contains `file`, it won't appear in the results.
  2. It's case-sensitive.
  3. It first looks up for the modules. That's only after we've typed the full module name followed by a `.` that we'll see the symbols from the module. In other words, we can't search for a symbol across modules.
  
  
If we want to improve that, we could:

* Use the `all` function instead of `complete`. This function returns all the symbols. We could then do the filtering based on the query by ourselves. That might be less efficient though since we'll always ask for all the symbols.
* Get rid of the `ocp-index` and re-implement things by ourselves?
  
I would love to have your input on that. Whether the current implementation with its limitations are acceptable or not? And if not, what would be the way to go?


ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,57 @@
...
+  in
+  let uri = loc.loc_start.pos_fname in
+  let range = Range.of_loc loc in
+  let location : Location.t = { uri; range } in
+  let name = LibIndex.Print.path info in
+  let kind = kind info.kind in
+  SymbolInformation.create ~name ~kind ~location ()
+
+let run ({ query } : WorkspaceSymbolParams.t) =
+  let _, build_dir = IndexMisc.project_root () in
```
tatchi, line 43: Any better way to get the `_build` directory path? From what I understand, this function walks up the tree starting from the current dir until it finds the build directory.

Also, what if we have multiple workspaces? I haven't tested that yet and wondering if we should do anything special?

rgrinberg: > Any better way to get the _build directory path? From what I understand, this function walks up the tree starting from the current dir until it finds the build directory.

This is ok for now. Eventually we'll have better integration with dune.

> Also, what if we have multiple workspaces? I haven't tested that yet and wondering if we should do anything special?

The user will need to run a server per workspace then. The model we should stick to is 1 lsp server per dune workspace.

rgrinberg:
I have to say that I'm not a fan of using ocp-index. The only good thing ocp-index provides (IMHO) is the efficient prefix search, but for us that isn't even the right behavior and we'd much prefer full substring search. Additionally, the compatibility problems it creates with merlin are annoying to say the least.

tatchi:
> I have to say that I'm not a fan of using ocp-index. The only good thing ocp-index provides (IMHO) is the efficient prefix search, but for us that isn't even the right behavior and we'd much prefer full substring search. Additionally, the compatibility problems it creates with merlin are annoying to say the least.

Thanks for the feedback. It will probably take me some time, but I can definitely try to implement it without using `ocp-index`. I guess it should be doable, now that I have a better understanding and that I have an implementation I can base myself on.

One thing I'm a little worried about: I see a [few code paths](https://github.com/OCamlPro/ocp-index/blob/master/libs/indexBuild.ml#L410) that depend on the OCaml version. Does that mean that the `typedtree` structure is different from one OCaml version to another? If so, what are the versions we would like to support?

rgrinberg:
> One thing I'm a little worried about: I see a few code paths that depend on the OCaml version. Does that mean that the typedtree structure is different from one OCaml version to another? If so, what are the versions we would like to support?

We only support the OCaml version that our particular version of merlin does. So that means 4.12 at the moment. That should simplify things.

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,261 @@
...
+    Array.fold_left
+      ~f:(fun (roots, found) f ->
+        match memsuffix f roots with
+        | None -> (roots, found)
+        | Some roots -> (roots, Some f))
+      ~init:(build_roots, None) files
+  in
+  match root with
+  | None -> None
+  | Some (("_obuild" | "_build") as dir) -> Some (path / dir)
```
rgrinberg, line 85: `_obuild`?

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,261 @@
...
+  | Types.Sig_modtype (id, _, _)
+  | Types.Sig_class (id, _, _, _)
+  | Types.Sig_class_type (id, _, _, _) ->
+    id
+
+let kind_of_sig_item (signature_item : Types.signature_item) : SymbolKind.t =
+  match signature_item with
+  | Types.Sig_value _ -> Function
+  | Types.Sig_type _ -> String
+  (* | Types.Sig_typext (_, _, Types.Text_exception, _) -> Constructor *)
```
rgrinberg, line 42: It does make sense to extract exception declaration.

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,261 @@
...
+  let name = Ident.name id in
+  let kind = kind_of_sig_item signature_item in
+  let range = Range.of_loc (loc |> to_loc) in
+  let uri = sourcefile |> Uri.of_path in
+  let location = Location.create ~uri ~range in
+  SymbolInformation.create ~name ~kind ~location ()
+
+let build_roots =
+  (* by increasing order of priority *)
+  [ "_darcs"; ".hg"; ".git"; "jengaroot.ml"; "omakeroot"; "_build"; "_obuild" ]
```
rgrinberg, line 62: Let's just stick to `_build`. There's no need for half hearted support for other build systems. If we have users interested in using other build systems, we'll think of a proper way to integrate.

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,261 @@
...
+  | Cmt of string
+  | Cmti of string
+
+let string_of_cm cm =
+  match cm with
+  | Cmt f
+  | Cmti f ->
+    f
+
+module Cm_Files = Map.Make (struct
```
rgrinberg, line 176: Our naming convention is `Cm_files`.

ocaml-lsp-server/src/workspace_symbol.ml
tatchi: Yes, that makes sense. This code was copied from `ocp-index` and I didn't want to modify it before I had something that works.

ocaml-lsp-server/src/workspace_symbol.ml
tatchi: That's covered by the line below (43):

```ocaml
Types.Sig_typext _ -> Constructor
```

`ocp-index` has an `Exception` variant, hence why they handle this case differently: https://github.com/OCamlPro/ocp-index/blob/master/libs/indexBuild.ml#L355

In our case, `SymbolKind.t` hasn't any `Exception` variant, so I thought I would always map `Types.Sig_typext` to `Constructor`. I'm definitely not sure about the mapping between the `Types.signature_item` and `SymbolKind.t` so that's for sure something we'll have to revisit at some point.

ocaml-lsp-server/src/state.ml
```diff
@@ -2,7 +2,7 @@ open Import
 
 type init =
   | Uninitialized
-  | Initialized of ClientCapabilities.t
+  | Initialized of InitializeParams.t
```
tatchi, line 5: Previously, we were trying to find the `_build` folder by walking up the tree based on the `cwd`. This caused an issue when I implemented the snapshot test because it was taking `cmt` files under the project's `_build` folder instead of the test folder (`__tests__/workspace_symbol`).

To overcome this issue, I had to [specify](https://github.com/ocaml/ocaml-lsp/pull/398/files#diff-c84ef09154bddc0a37a75301087282d6a921acfe31607b44cdde69b4ea38198bR13) the `rootUri` during the `initialize` request. Since on the server-side we didn't store this value, I had no way to use it. Hence why I now store the whole `InitializeParams.t` instead of just the `capabilities`.

Even without taking the tests into account, I think that it makes sense to keep the full `initializeParams`. Also because `roorUri` is deprecated and we'll need at some point to also use the `workspaceFolders` (introduced in 3.6.0)

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -555,6 +558,20 @@ let definition_query (state : State.t) uri position merlin_request =
...
 
+let workspace_symbol (state : State.t) (params : WorkspaceSymbolParams.t) =
+  let rootUri = (init_params state).rootUri in
+  let symbols = Workspace_symbol.run params rootUri in
+  match symbols with
+  | Ok symbols -> Fiber.return (Some symbols)
+  | Error e -> (
+    match e with
+    | Workspace_symbol.No_build_dir path ->
+      Jsonrpc.Response.Error.raise
```
tatchi, line 569: Is it the correct place to raise an error? Also, `InternalError` is probably not the good error code 🤔

tatchi: Actually, I'm not even sure we should raise an error in this case. I guess it will happen mostly because the user did not run the build command yet. Maybe we should display a notification instead to inform the user?

ocaml-lsp-server/test/e2e/__tests__/__snapshots__/workspace-symbol.test.ts.snap
```diff
@@ -0,0 +1,1485 @@
...
+        "end": Object {
+          "character": 23,
+          "line": 1,
+        },
+        "start": Object {
+          "character": 0,
+          "line": 1,
+        },
+      },
+      "uri": "file://<testWorkspacePath>/bin/.main.eobjs/dune__exe.ml-gen",
```
tatchi, line 18: These are the modules created from files. Do we want to keep them in the result? If so, we'll have to figure out how to get the correct `uri`

image

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,216 @@
...
+  in
+  List.fold_left
+    ~f:(fun files dir ->
+      Sys.readdir dir
+      |> Array.fold_left
+           ~f:(fun acc file ->
+             let modul, ext = split_filename file in
+             let path = dir ^ "/" ^ file in
+             match ext with
+             | "cmt" -> Cm_files.set acc modul (Cmt path)
```
tatchi, line 156: I noticed that some of the `cmt` files end up in both the `_build/default` and `_build/install` folder. That was resulting in some duplications when looking up for a symbol. That's why I store them in a Map whose keys are the modules' names. I thought it would be reasonable since, AFAIK, modules' names should be unique across the project.

If I'm not mistaken, files in `_build/install` are actually `symlinks`. So we could also just follow the `symlink` until we reach the real file, and store the `path` in e.g a Set. That's probably what we'll have to do anyway if we want to support `symbols` lookup across workspaces? Since in that case, I guess it would be valid to have some duplication of modules' names

rgrinberg: We should just look up cmt's in `_build/default`. That's not entirely accurate as the user can change the merlin context, but that's good enough for now.

ocaml-lsp-server/test/e2e/__tests__/__snapshots__/workspace-symbol.test.ts.snap
rgrinberg: I think any cmt file with `ml-gen` as the source extension should be excluded.

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: Internal error is definitely not right. A message asking the user to build the project first seems fine.

ocaml-lsp-server/test/e2e/__tests__/__snapshots__/workspace-symbol.test.ts.snap
```diff
@@ -0,0 +1,1485 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`workspace/symbol returns symbols from workspace 1`] = `
```
rgrinberg, line 3: This test should be cleaned up so the output is much smaller. Perhaps a different workspace? Or post process the result to only include the important stuff?

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,216 @@
...
+        | "cmt"
+        | "cmti" ->
+          read_cmt rootPath file
+        | _ -> [])
+    in
+    let all_symbols = List.map ~f:load_file all_files |> List.flatten in
+    match query with
+    | "" -> Ok all_symbols
+    | query ->
+      let reg_exp = Str.regexp_string query in
```
rgrinberg, line 201: Unfortunately, `Str` will not work for us (it's too hard to use correctly in a multi threaded context). Use `Re` instead.

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,216 @@
...
+          String.lowercase_ascii
+            (String.sub file ~pos:(i + 1) ~len:(len - i - 1))
+        in
+        match ext with
+        | "cmt"
+        | "cmti" ->
+          read_cmt rootPath file
+        | _ -> [])
+    in
+    let all_symbols = List.map ~f:load_file all_files |> List.flatten in
```
rgrinberg, line 197: `List.concat_map`.

ocaml-lsp-server/test/e2e/__tests__/__snapshots__/workspace-symbol.test.ts.snap
```diff
@@ -0,0 +1,1485 @@
...
+          "line": 98,
+        },
+        "start": Object {
+          "character": 0,
+          "line": 98,
+        },
+      },
+      "uri": "file://<testWorkspacePath>/lib/lib.ml",
+    },
+    "name": "append",
```
tatchi, line 581: That's what gets produced due to the `include List`. The `cmt.cmt_sourcefile` point to the file where we use the `include` (`lib/lib.ml`) but the location is the location where the function (`append` in this case) is defined (in the `List` module somewhere e.g in the `_opam` folder). Therefore, we end up at a wrong place in the file where we use the `include` when we click on the symbol. Would that be ok to ignore `symbols` from `include` for now?

tatchi:
@rgrinberg I made some adjustments based on your first review. I also created a snapshot test so we can have a better idea of what gets changed as we move on.

There are a few things I noticed and that I'm planning to tackle next:

* We can't search for symbols defined in a module
* As explained [here](https://github.com/ocaml/ocaml-lsp/pull/398#discussion_r624429823), `include` are a bit tricky and currently produce wrong results.
* We can't search across `workspaces`

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,216 @@
+open Import
+
+let to_loc (loc : Warnings.loc) : Loc.t =
+  { loc_start = loc.loc_start
```
rgrinberg, line 4: Why is this needed? I'm pretty sure that `Warnings.loc = Loc.t`

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,216 @@
...
+        else
+          acc)
+      ~init:(path :: acc) (Sys.readdir path)
+  in
+  subdirs [] dir
+
+let read_cmt root_uri cmt_file =
+  let cmt = Cmt_format.read_cmt cmt_file in
+  let sourcefile =
+    root_uri ^ "/" ^ Option.value ~default:"" cmt.cmt_sourcefile
```
rgrinberg, line 104: `Filename.concat` perhaps?

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,216 @@
...
+
+let find_dirfiles dirs =
+  let choose_file f1 f2 =
+    match (f1, f2) with
+    | (Cmt _ as f), _
+    | _, (Cmt _ as f) ->
+      f
+    | (Cmti _ as f), Cmti _ -> f
+  in
+  let split_filename file =
```
rgrinberg, line 135: `String.rsplit2` can be used to simplify this.

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,216 @@
...
+      in
+      (modul, ext)
+  in
+  List.fold_left
+    ~f:(fun files dir ->
+      Sys.readdir dir
+      |> Array.fold_left
+           ~f:(fun acc file ->
+             let modul, ext = split_filename file in
+             let path = dir ^ "/" ^ file in
```
rgrinberg, line 154: `Filename.concat`

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,216 @@
...
+             match ext with
+             | "cmt" -> Cm_files.set acc modul (Cmt path)
+             | "cmti" -> (
+               let current_file = Cm_files.find acc modul in
+               let cmti_file = Cmti path in
```
rgrinberg, line 159: Instead of looking cmti files, why don't we look in cmi files?

ocaml-lsp-server/src/workspace_symbol.ml
tatchi: Good catch. That's something I needed in the first implementation that was using `ocp-index`

ocaml-lsp-server/test/e2e/__tests__/__snapshots__/workspace-symbol.test.ts.snap
tatchi: I went for keeping `.ml(i)` files only (a682d739311a399c1b6313c284927cc8fd15aac8). Does that make sense?

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -172,14 +172,14 @@ let run ({ query; _ } : WorkspaceSymbolParams.t) (rootUri : Uri.t option) =
...
-      let reg_exp = Str.regexp_string query in
+      let reg_exp = Re.Str.regexp_string query in
       let symbols =
         all_symbols
         |> List.filter ~f:(fun symbol ->
                try
                  let () =
                    ignore
-                     (Str.search_forward reg_exp
+                     (Re.Str.search_forward reg_exp
```
rgrinberg, line 185: I believe `Re.Str` suffers from the same problems. You can just use the vanilla `Re` api instead.

ocaml-lsp-server/src/workspace_symbol.ml
tatchi: I think I did that because of [this comment](https://github.com/OCamlPro/ocp-index/blob/master/libs/indexBuild.ml#L1263). My understanding is that it should be the same as `cmi` but with more info. To be honest, the difference between all of them is not clear yet to me 😅

Anyway, I guess `cmi` files should be fine too. I'll make the change :)

rgrinberg:
> Anyway, I guess `cmi` files should be fine too. I'll make the change :)

My understanding is between cmi's and cmti's files, there's a lot of
redundancy and that we should favour cmi's whenever possible. I've seen
a couple of people expressing the desire to get rid of cmti's in the
future.  So I prefer to use cmi's whenever possible to stay on the
"safe" side.  However, if you find that the feature works better with
cmti's for whatever reason, feel free to keep it.

ocaml-lsp-server/test/e2e/__tests__/__snapshots__/workspace-symbol.test.ts.snap
rgrinberg: Seems fine to me.

ocaml-lsp-server/test/e2e/__tests__/__snapshots__/workspace-symbol.test.ts.snap
tatchi: Should have been addressed in https://github.com/ocaml/ocaml-lsp/pull/398/commits/8b31c489f2a152674479f3c2a768b2edd7d9d967

ocaml-lsp-server/test/e2e/__tests__/__snapshots__/workspace-symbol.test.ts.snap
tatchi: https://github.com/ocaml/ocaml-lsp/pull/398/commits/1e26f5707299c507865e063ce78184508b09e4a6 does that look better?

I also switched over to using inlineSnaptshot in https://github.com/ocaml/ocaml-lsp/pull/398/commits/1b8a17b51cf439fc6c68ca3777fe33308f11162b since that's what is being used elsewhere

rgrinberg:
@ulugbekna do you mind reviewing/testing this as well? I need to focus on the dune integration at the moment.

tatchi:
To give you an update on the PR, I should have addressed most of @rgrinberg's comments except for https://github.com/ocaml/ocaml-lsp/pull/398#discussion_r624423283 that I still need to implement.

I also would like to:

* Support multiple `workspaces`
* Improve symbols lookup:
  * Figure out why `stack_of_ints` appears many time in the tests: https://github.com/ocaml/ocaml-lsp/pull/398/files#diff-c84ef09154bddc0a37a75301087282d6a921acfe31607b44cdde69b4ea38198bR62
  * Only the symbols at the "top" level of the file are currently shown. Symbols defined in a `module`, `class` or `expression` for instance are not shown.
  
```ocaml
let a = 
  let b = 5 in (* b is not in the symbols result *)
  b

module A = struct
  let a = 5 (* a is not in the symbols result *)
end

```

I'll ping you once I've addressed these items :)

ulugbekna:
@rgrinberg ok, I can take over reviewing/testing

@tatchi sounds good, thanks!

ocaml-lsp-server/src/state.ml
rgrinberg: This sounds fine to me.

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,185 @@
...
+let is_directory dir =
+  try Sys.is_directory dir with
+  | Sys_error _ -> false
+
+let find_build_dir dir =
+  let build_dir = "_build" in
+  let is_build folder = Filename.basename folder = build_dir in
+  let dirs dir =
+    dir |> Sys.readdir |> Array.to_list
+    |> List.map ~f:(Filename.concat dir)
```
rgrinberg, line 63: List.filter_map

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,185 @@
...
+    f
+
+let extension file =
+  match String.rsplit2 ~on:'.' file with
+  | Some (_, ext) -> Some ext
+  | None -> None
+
+let symbols_from_cm_file root_uri cm_file =
+  let filename_from_path path =
+    match String.rsplit2 ~on:'/' path with
```
rgrinberg, line 95: Can Filename.basename be used instead?

ocaml-lsp-server/src/workspace_symbol.ml
rgrinberg: Even better: write `Array.find_map` and use that.

rgrinberg:
Apart from some nits, the PR looks good. Good work.

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -159,32 +148,28 @@ let find_cm_files dir =
...
+         let path = Uri.to_path uri in
+         let* build_dir = find_build_dir path in
+         let cm_files = find_cm_files build_dir in
+         let all_symbols =
+           List.concat_map ~f:(symbols_from_cm_file path) cm_files
+         in
+         match query with
+         | "" -> Ok all_symbols
+         | query ->
+           let match_query = Re.str query |> Re.compile |> Re.execp in
```
rgrinberg, line 169: it would be more efficient to compile the query only once.

rgrinberg:
@tatchi do you still have some changes planned? If not, then could you remove the "Draft" status?

tatchi:
> @tatchi do you still have some changes planned? If not, then could you remove the "Draft" status?

Sorry, I wanted to push one last commit which I just did now. I removed the draft status and rebased the PR :)

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -566,6 +569,61 @@ let definition_query server (state : State.t) uri position merlin_request =
...
+  let to_tuple t =
+    let fst, snd =
+      List.fold_left t
+        ~f:(fun (rs, es) x ->
+          match x with
+          | Error e -> (rs, e :: es)
+          | Ok r -> (r :: rs, es))
+        ~init:([], [])
+    in
+    (List.rev fst, List.rev snd)
```
tatchi, line 607: Does that make sense? The `Workspace_symbol.run` function returns a list of result. What I need is to get the list of `Ok symbols` on one side (to send the symbols to the client) and the list of `Error` on the other side (to send a notification to the client).

rgrinberg: It makes sense, but there's already a function to do this: List.partition_map

ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -0,0 +1,167 @@
...
+let string_of_cm cm =
+  match cm with
+  | Cmt f
+  | Cmi f ->
+    f
+
+let symbols_from_cm_file root_uri cm_file =
+  let filename_from_path path = Filename.basename path in
+  let filename = string_of_cm cm_file in
+  let cmt = Cmt_format.read_cmt filename in
```
rgrinberg, line 80: This seems wrong to me - it clearly cannot work for cmi files. Shouldn't it be `Cmi_format.read_cmi` in that case?

rgrinberg:
Hey @tatchi, I did a deeper review. Everything looks good and was easy to understand. Apart from stylistic changes, I made these bigger changes:

* Traverse directories and generate symbols in a single pass. The old approach would call readdir on the same directory twice. The new approach makes sure to do so only once.

* The traversal is now offloaded to a background thread. The main benefit is that this does not block other requests from being handled as we're traversing. This is done in a separate commit.

There's still one issue remaining with regards to Cmi_format vs. Cmt_format. Once that is addressed, this PR can be merged.

ocaml-lsp-server/src/fmt.ml
```diff
@@ -6,32 +6,6 @@ type command_result =
...
-    let* v = Fiber.Ivar.peek t.value in
-    match v with
-    | Some s -> Fiber.return s
-    | None ->
-      let* v = (Option.value_exn t.f) () in
-      let+ () = Fiber.Ivar.fill t.value v in
-      t.f <- None;
-      v
-end
-
```
tatchi, line 34: Was that change intentional? Looks like it produces a build error (I've the same locally): https://github.com/ocaml/ocaml-lsp/pull/398/checks

rgrinberg:
> Was that change intentional? Looks like it produces a build error (I've the same locally): https://github.com/ocaml/ocaml-lsp/pull/398/checks
I forgot to commit a couple of files. Fixed

ocaml-lsp-server/test/e2e/__tests__/__snapshots__/workspace-symbol.test.ts.snap
rgrinberg: Looks fine to me.

ocaml-lsp-server/src/workspace_symbol.ml
tatchi: I switched over to using `cmti` files instead of `cmi`. It appears that `cmti` [is what gets created for `mli` files](https://discord.com/channels/436568060288172042/436568060288172044/824343476900593704)

The advantage is that they can be read with the same `Cmt_format.read` function and thus return the same type.

rgrinberg:
I changed the implementation to use merlin's Outline functionality. This gives us two improvements:

* Symbol information is constructed consistently. E.g. we now get deprecated information
* We now get correct locations for all symbols. Previously, we'd use locations from the inferred interface of modules. The locations there are mostly wrong.

Thanks for all your efforts!

Issue #344, opened 2020-12-08
Wrong documentation on hover


ulugbekna:
On hover, I sometimes get documentation for a completely different thing than the identifier I hover over. The "different thing" is usually something that comes from stdlib such as `>` operator or as in the screenshot, where I hover over `err_pos_range` and get correct type but not doc.

![image](https://user-images.githubusercontent.com/16353531/101482455-94c92100-3978-11eb-896c-51c3304f7f9b.png)

(might be a merlin issue, but the last issue found with ocaml-lsp I raised in merlin repo turned out to be ocaml-lsp related, so creating the issue here)

tmattio:
Related to https://github.com/ocamllabs/vscode-ocaml-platform/issues/111

ulugbekna:
That issue could be closed since the problem cannot be fixed on the vscode
extension side.

ulugbekna:
I think I got a reproducible case.

When one types a module name `List`, example : 

```
(* file s.ml *) 
module M = struct 
  let u = () 
  let f = List.<cursor>
end
```

for some reason the documentation on hover at position `p` in file `s.ml` is fetched from the file where `List` is defined, when it should actually fetch the documentation from file `s.ml`. I have only noticed the issue with stdlib modules. 

In the video I need to push the module def down to reproduce the bug, because the `List` module documentation in `list.mli` occurs on lines ~20; hence, we see that the documentation is fetched for things at position `p` from `list.mli` instead of the currently open file.

https://user-images.githubusercontent.com/16353531/109388892-a7297a00-792b-11eb-955a-0b2d6b3f1286.mov

ulugbekna:
This is fixed on master by updating to the latest merlin 4.2

rgrinberg:
@ulugbekna can you make a regression test for this? I'll try and have a look.

rgrinberg:
I'm unable to reproduce this issue with the example above.

ulugbekna:
Sorry, I missed your first message. I'll try to come up with a test. I can
still reproduce it manually though.

Sir4ur0n:
I never know if this kind of message is useful, but just in case: I regularly encounter this problem too, and so do several of my teammates (to confirm this is not an isolated issue for OP).

rgrinberg:
What's blocking me from making progress on this issue is a (preferably small) reproducible test case. Next time you run into this issue, see if you can isolate it.

Rudi.

smelc:
[...]

PR #399, opened 2021-04-03
add support for refactor-open (qualify and unqualify) code action

ulugbekna:
Reservations I have:
- [x] I'm not sure code action names are easily understandable / friendly to newcomers

Blockers: 

- [x] Code actions are shown only if the cursor is on `open` statement, but they can be shown even if the code action won't do anything, e.g.,

```ocaml
open Unix (* both qualify and unqualify code actions will be shown *)
let u = () 
```
- [x] (this is merged into ocaml/merlin#master, we need to rebase vendored merlin) `refactor-open qualify` qualifies with full paths, e.g., for `open Import`, it does `Dune__exe.Import.foo...`.

To-do 
- [x] add tests
- [x] add an entry to the README, so that users know about these code actions

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,6 +48,34 @@
 
 - Change the formatting-on-save error notification to a warning notification (#472)
 
+- Code action to qualify ("put module name in identifiers") and unqualify
+  ("remove module name from identifiers") module names in identifiers (#399)
+
+  Starting from:
+  ```ocaml
+  open Unix
+
+  let times = Unix.times ()
+  let f x = x.Unix.tms_stime, x.Unix.tms_utime
+  ```
+
+  Calling "remove module name from identifiers" with the cursor on the open
+  statement will produce:
+  ```ocaml
+  open Unix
+
+  let times = times ()
+  let f x = x.tms_stime, x.tms_utime
+  ```
+
+  Calling "put module name in identifiers" will restore:
+  ```ocaml
+  open Unix
+
+  let times = Unix.times ()
+  let f x = x.Unix.tms_stime, x.Unix.tms_utime
+  ```
+
 # 1.7.0 (07/28/2021)
 
 ## Features
@@ -83,7 +111,6 @@
 ## Features
 
 - Code action to annotate a value with its type (#397)
-
 ## Fixes
 
 - Fix interface/implementation switching on Windows (#427)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -76,6 +76,12 @@
   let f x = x.Unix.tms_stime, x.Unix.tms_utime
   ```
 
+- Bug fix: do not show "random" documentation on hover
+  - fixed by [merlin#1364](https://github.com/ocaml/merlin/pull/1364)
+  - fixes duplicate:
+    - [ocaml-lsp#344](https://github.com/ocaml/ocaml-lsp/issues/344)
+    - [vscode-ocaml-platform#111](https://github.com/ocamllabs/vscode-ocaml-platform/issues/111)
+
 # 1.7.0 (07/28/2021)
 
 ## Features
--- a/ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
+++ b/ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
@@ -445,4 +445,160 @@ let x = _
       }
     `);
   });
+
+  type refactorOpenTestSpec = {
+    documentUri?: string;
+    documentText: string;
+    queryStartPos: Types.Position;
+    queryEndPos: Types.Position;
+    codeActionTitle: string;
+  };
+
+  // this removes some repetition in code for testing `refactor-open` code actions
+  // it specifically doesn't include `expect(...).toMatchInlineSnapshot` to be able to
+  // capture correct output (the snapshot) from jest automatically
+  // (similar to ppx_expect promotion with correct output)
+  async function testRefactorOpen({
+    documentUri,
+    documentText,
+    queryStartPos,
+    queryEndPos,
+    codeActionTitle,
+  }: refactorOpenTestSpec) {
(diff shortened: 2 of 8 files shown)


rgrinberg:
I don't mind the names. This is one of those things that's easy enough to learn by example.

> Code actions are shown only if the cursor is on open statement, but they can be shown even if the code action won't do anything, e.g.,

If the code action won't do anything, the edit it computes should be empty. Can we filter it out then?

ulugbekna:
> If the code action won't do anything, the edit it computes should be empty. Can we filter it out then?

Unfortunately, merlin always returns edits, even if those edits when applied do not change the document at all. 

--- 

Other notes:

1. The code action is quite useful for stdlib and in-file modules refactor-open qualify, as is seen in the tests. However, merlin doesn't use _short paths_ when qualifying modules. Example:

```ocaml
open Import
module Bin = Stdune.Bin

let _PATH =
  lazy
    (Bin.parse_path
       (Dune__exe.Import.Option.value ~default:"" (Dune__exe.Import.Unix_env.get Dune__exe.Import.Unix_env.initial "PATH")))
```

Meanwhile, do you happen to know if that's also the case with refactor-open in merlin's emacs mode?

I will try to fix this anyway.

2. Do you happen to know what's a good way to do several sequenced queries to merlin (vs `ocamlmerlin single`)? I need merlin to take into account ".merlin" and all source files (while `single` query only can take one)?

rgrinberg:
> Meanwhile, do you happen to know if that's also the case with refactor-open in merlin's emacs mode?

No idea. I don't use the emacs mode.

> Do you happen to know what's a good way to do several sequenced queries to merlin (vs ocamlmerlin single)? I need merlin to take into account ".merlin" and all source files (while single query only can take one)?

Is this for tests or something? We don't have .merlin files anymore.

Let's see if @voodoos can help us out.

ulugbekna:
A due update: 

1. I created PRs to the merlin repo for short-paths and useless edits problems (https://github.com/ocaml/merlin/pull/1313 and https://github.com/ocaml/merlin/pull/1314).

I would appreciate your take on whether we should merge this PR as is and wait for upstream merlin to merge fix-PRs, or merge those in ocaml-lsp merlin. Both are fine with me. 

2. 
> > Do you happen to know what's a good way to do several sequenced queries to merlin (vs ocamlmerlin single)? I need merlin to take into account ".merlin" and all source files (while single query only can take one)?
> Is this for tests or something? We don't have .merlin files anymore.

Yes, I needed to write some tests for merlin, but I figured this out.

rgrinberg:
> I would appreciate your take on whether we should merge this PR as is and wait for upstream merlin to merge fix-PRs, or merge those in ocaml-lsp merlin. Both are fine with me.

Do you think this feature is useful to users in its current form?

ulugbekna:
Let's aim to merge this for the release after the next one. I hope by then we'll get a feel whether ocaml/merlin is willing to merge.

ulugbekna:
@rgrinberg I think this PR is ready as soon as we switch to using the latest merlin master. I think that this feature has improved quite a bit in merlin and should be useful to ocaml-lsp users as well :-) 

I included b975615 in this PR as well: 
- what it does? when we update to the latest merlin, we automatically get the fix for #344, so I included the changes entry about it and updated the according test
- why here? we anyway rebase and merge, so including this here should be innocuous; let me know if you'd prefer it in a separate PR 

I can squash all but b975615 for cleaner history; lmk if you'd like that or you have other proposals of course.

ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
```diff
@@ -445,4 +445,233 @@ let x = _
...
+          },
+          "isPreferred": false,
+          "kind": "remove module name from identifiers",
+          "title": "Remove module name from identifiers",
+        },
+      ]
+    `);
+  });
+
+  it("refactor-open unqualify external library module", async () => {
```
rgrinberg, line 545: This relies on the fact that cmi's of the unix library are available even without the user listing `unix` as an explicit dependency. We don't want our tests to rely on this bug. A true test that uses an external library will tell the server that this library is available.

CHANGES.md
```diff
@@ -48,6 +48,12 @@
 
 - Change the formatting-on-save error notification to a warning notification (#472)
 
+- Bug fix: do not show "random" documentation on hover
```
rgrinberg, line 51: Isn't this already in the merlin change list? We usually do not replicate merlin's change list here.

rgrinberg (approved):
The PR looks good but we can't rely on the following special libraries in the tests: `unix`, `str`, `dynlink`.

CHANGES.md
ulugbekna: It is already in the merlin changes list, but since this bug has been reported several times in both ocaml-lsp & vscode-ocaml-platform, I thought it is useful for users to see it fixed. 

I think that merlin is ocaml-lsp's backend, but the end user shouldn't be lurking in merlin Changes to see what's new in ocaml-lsp -- maybe we could at least add a link to their Changes inside the ocaml-lsp-vendored merlin source.

ulugbekna: The change entry in merlin is "locate: reset global state from all entry points (#1364)", which is cryptic, isn't it?

rgrinberg: It is cryptic. But that can be mended on merlin's side. I'd prefer to see the links to the corresponding change list entries where the fix was done.

> I think that merlin is ocaml-lsp's backend, but the end user shouldn't be lurking in merlin Changes to see what's new in ocaml-lsp -- maybe we could at least add a link to their Changes inside the ocaml-lsp-vendored merlin source.

We could have a policy of replicating only the relevant merlin changes in our repo. The problem is that I often lack context as to what is relevant to lsp, and it is extra work. We can bring it up with the rest of the team to see what they think.

rgrinberg (approved):

PR #402, opened 2021-04-04
[Fiber] allow concurrent writes

rgrinberg:
Our use of them is harmless

### DIFF
--- a/vendor/fiber/fiber.ml
+++ b/vendor/fiber/fiber.ml
@@ -757,19 +757,10 @@ module Stream = struct
   end
 
   module Out = struct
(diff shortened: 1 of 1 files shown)


ulugbekna (approved):
Thanks for a quick fix! The crashes are gone, and this looks good.

(I don't know whether it would make future fiber updates easier in comparison to this PR if we introduced the patches by overriding `Stream.Out` in `import.ml` or creating concurrent-write version for `Stream.Out`)

rgrinberg:
I'd rather just fix the issue upstream. I think this check does more harm than good anyway.

PR #406, opened 2021-04-08
3.16 Support

rgrinberg:
Updating the spec, code generator, and the generated code.


rgrinberg:
@ulugbekna to review this PR, it's enough to make sure we like the new `types.mli`. The code generator does not matter so much.

ulugbekna (approved):
Great work, thanks!

PR #411, opened 2021-04-12
Print jsonrpc errors

rgrinberg:
Should make it simpler to debug issues for ourselves and other users of this
library.

### DIFF
--- a/jsonrpc/src/jsonrpc.ml
+++ b/jsonrpc/src/jsonrpc.ml
@@ -213,6 +213,13 @@ module Response = struct
 
     exception E of t
 
+    let () =
+      Printexc.register_printer (function
+        | E t ->
+          Some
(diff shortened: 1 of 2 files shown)


jsonrpc/src/jsonrpc.ml
```diff
@@ -213,6 +213,13 @@ module Response = struct
 
     exception E of t
 
+    let () =
+      Printexc.register_printer (function
+        | E t ->
+          Some
+            ("jsonrcp response error " ^ Json.to_pretty_string (yojson_of_t t))
```
ulugbekna, line 220:
```suggestion
            ("jsonrpc response error " ^ Json.to_pretty_string (yojson_of_t t))
```

ulugbekna (approved):

PR #412, opened 2021-04-12
Fix jsonrpc memory leak

rgrinberg:
A recursive call inside a fork is going to allocate a closure that will only be
collected by the GC once all the recursive calls are done. Since we are looping
over all incoming packets, that's essentially a memory leak. This PR removes the
calls to fork_and_join_unit that rely on recursion.

I was always aware of this leak, but never bothered to fix it since it's quite
tiny (1 closure per incoming packet). I was looking for ways to use it other
than a pool, but eventually gave up.

### DIFF
--- a/jsonrpc-fiber/src/jsonrpc_fiber.ml
+++ b/jsonrpc-fiber/src/jsonrpc_fiber.ml
@@ -148,6 +148,7 @@ struct
           Log.msg "sending response" [ ("response", Response.yojson_of_t resp) ]);
       Chan.send t.chan (Response resp)
     in
+    let later = Fiber.Pool.create () in
     let rec loop () =
       t.tick <- t.tick + 1;
       log t (fun () -> Log.msg "new tick" [ ("tick", `Int t.tick) ]);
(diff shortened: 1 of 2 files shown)


ulugbekna:
I just noticed that with #410 ubuntu tests were failing, and with this PR all tests are failing. Is this expected?

rgrinberg:
> I just noticed that with #410 ubuntu tests were failing

No idea what was happening there. I've yet to reproduce it, but I wouldn't be surprised if there's a bit of non-determinism in those tests since they rely on IO.

>  and with this PR all tests are failing. Is this expected?

#410 caused a subtle change in shutdown behavior in conjunction with this PR. I updated the tests.

ulugbekna (approved):
This looks good. 

Last remark: `Fiber.Pool.stop` doc says that the submitted tasks will need to execute, which is not what we want I guess, but I haven't noticed any problems.

rgrinberg:
> Fiber.Pool.stop doc says that the submitted tasks will need to execute, which is not what we want I guess, but I haven't noticed any problems.

Yes, I considered the behavior of dropping all pending tasks after `stop`, but then callers would have no way of knowing their task was rejected. So I just went for this instead.

PR #414, opened 2021-04-12
Improve logging and set trace level

rgrinberg:


rgrinberg:
@giltho do you have time to review this one?

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -839,17 +823,23 @@ let on_notification server (notification : Client_notification.t) :
     Fiber.return state
   | Unknown_notification req -> (
     match req.method_ with
-    | "$/setTraceNotification"
+    | "$/setTraceNotification" -> (
```
giltho, line 826: Why isn't setTrace parsed as a known method ?
I mean it might not be necessary to create a whole module just for that..

ocaml-lsp-server/src/state.ml
```diff
@@ -11,4 +11,5 @@ type t =
   ; scheduler : Scheduler.t
   ; detached : Fiber.Pool.t
   ; configuration : Configuration.t
+  ; trace : TraceValue.t
```
giltho, line 14: The trace value is set, but not really used yet. It should maybe be used to update the `!level` function ?

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -824,7 +808,7 @@ let on_notification server (notification : Client_notification.t) :
       let+ () = send_diagnostics server doc in
       state)
   | CancelRequest _ ->
-    log ~title:Logger.Title.Warning "ignoring cancellation";
+    Log.log ~section:"debug" (fun () -> Log.msg "ignoring cancellation" []);
```
giltho, line 811: I believe that log function should be used to write a logTrace notification.
But then it's a different choice, it means that you'll have to read the logs in something like vscode, it won't be written in a file. The good side would be to leverage the LSP inspector, but right now it is down https://github.com/microsoft/language-server-protocol/issues/1089
It's also easier to read the logs in real time.

But then, writing to stderr or a file makes sure that the info will be displayed even if the connection between the server and the editor is somehow broken

giltho (commented):
I always learn about at least one feature of OCaml when reading your code it's always really interesting !

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: I use Emacs, and I can already see the stderr for the server. Is this not possible for vscode?

ocaml-lsp-server/src/state.ml
rgrinberg: > The trace value is set, but not really used yet.

Yeah, I'm just not sure kind of messages need to be "traced". 

> It should maybe be used to update the !level function ?

That's one way to do it. Should trace receive non lsp related messages?

giltho:
I'm going to answer both of your questions here @rgrinberg 

Re VSCode, when the trace is not off, VSCode prints every req/notif in a special output window:
image
That is in verbose mode where messages and their content are displayed, in messages mode, you'll get something like that
image

LogTrace is a specific message that has no meaning but is used to add information to this trace, debug information or things like that. The idea is that you can then open the trace the lsp inspector and see everything in order.
By default, the log trace is off, so that the server doesn't spend ressources building and sending these messages if not needed. So logTrace should be used for anything, LSP-related or not, that has to do with debugging information.

The issue is that, until 3.16, that mechanism was implemented for VSCode and only half-specified (there was a logTrace level, but no meaning for it), so right now it is possible that very few editors other than VSCode implement it correctly...

I think that to see what's happening in stderr in vscode, you have to look in the browser's console (since it's in electron).

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: Mostly because it's not part of the standard. I suppose this shouldn't really matter to use and we should make it easy for users to make use of this common method.

giltho: Ah, sorry I should have caught that.
It was added in 3.16 and named `$/setTrace` (not `$/setTraceNotification`) : https://microsoft.github.io/language-server-protocol/specification#setTrace

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -154,29 +154,13 @@ let send_diagnostics ?diagnostics rpc doc =
...
-        | Debug -> (Log, Printf.sprintf "debug: %s" text)
-        | Notify -> (Log, Printf.sprintf "notify: %s" text)
-        | Custom s -> (Log, Printf.sprintf "%s: %s" s text)
-        | LocalDebug -> failwith "impossible"
-      in
-      let message = Printf.sprintf "[%s] %s" section text in
-      let notif = Server_notification.LogMessage { message; type_ } in
-      let (_ : _ Fiber.t) =
-        let state : State.t = Server.state rpc in
-        task_if_running state ~f:(fun () -> Server.notification rpc notif)
```
rgrinberg, line 173: By the way, this code was just broken.  It was ignoring a fiber value so I don't know how this was ever logged.

rgrinberg:
@giltho Thanks, I understand now. Unfortunately, I just realized that the way we forward traces to the server doesn't even work. We need the log consumer to return a fiber and this requires `Log.msg` to fiber valued. We need some sort of an escape hatch to enqueue logs messages and then run a fiber that will consume the queue in the background. That will take a little more work than I planned. This PR is still good to go since the current code is broken anyway.

PR #416, opened 2021-04-13
Support async waiting for process

rgrinberg:
This allows us to run ocamlformat asynchronously. This is not very useful right
now, but will be once we run ocamlformat in server mode and use dune rpc.

### DIFF
--- a/fiber-unix/src/scheduler.ml
+++ b/fiber-unix/src/scheduler.ml
@@ -95,6 +95,15 @@ end
 
 module Timer_id = Id.Make ()
 
+type process =
+  { pid : Pid.t
+  ; ivar : Unix.process_status Fiber.Ivar.t
+  }
+
+type process_state =
+  | Running of process
+  | Zombie of Unix.process_status
+
 type t =
   { mutable events_pending : int
   ; events : event Queue.t
@@ -111,6 +120,7 @@ type t =
   ; mutable waker : Thread.t
   ; (* TODO Replace with Removable_queue *)
     timers : (Timer_id.t, active_timer ref) Table.t
+  ; process_watcher : process_watcher Lazy.t
   }
 
 and event =
@@ -139,6 +149,14 @@ and active_timer =
   { scheduled : float
   ; ivar : [ `Resolved | `Cancelled ] Fiber.Ivar.t
   ; parent : timer
+  }
+
+and process_watcher =
+  { mutex : Mutex.t
+  ; something_is_running : Condition.t
+  ; table : (Pid.t, process_state) Table.t
+  ; process_scheduler : t
+  ; mutable running_count : int
   }
 
 let add_events t = function
@@ -225,28 +243,6 @@ let wake_loop t =
   in
   loop None
 
-let create () =
-  let t =
-    { events_pending = 0
-    ; events = Queue.create ()
-    ; events_mutex = Mutex.create ()
-    ; time_mutex = Mutex.create ()
-    ; event_ready = Condition.create ()
-    ; earliest_next_mutex = Mutex.create ()
-    ; earliest_next = None
-    ; earliest_next_barrier = Barrier.create ()
-    ; threads = []
-    ; timers = Table.create (module Timer_id) 10
-    ; timers_available = Condition.create ()
-    ; timers_available_mutex = Mutex.create ()
-    ; time = Thread.self ()
-    ; waker = Thread.self ()
-    }
-  in
-  t.time <- Thread.create time_loop t;
-  t.waker <- Thread.create wake_loop t;
-  t
-
 let create_thread scheduler =
   let worker =
     let do_ (Pending (f, ivar)) =
@@ -314,10 +310,6 @@ let cancel_timers t =
           false));
   Fiber.parallel_iter !timers ~f:(fun ivar -> Fiber.Ivar.fill ivar `Cancelled)
 
-let cleanup t =
-  Barrier.close t.earliest_next_barrier;
-  List.iter t.threads ~f:stop
-
 type run_error =
   | Never
   | Abort_requested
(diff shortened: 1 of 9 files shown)


fiber-unix/src/scheduler.ml
```diff
@@ -446,3 +418,174 @@ let cancel_timer (timer : timer) =
...
+        add_events t.process_scheduler [ Job_completed (status, job.ivar) ]
+      | Some (Running _) -> assert false
+
+    let remove t ~pid status =
+      match Table.find t.table pid with
+      | None -> Table.set t.table pid (Zombie status)
+      | Some (Running job) ->
+        t.running_count <- t.running_count - 1;
+        Table.remove t.table pid;
+        add_events t.process_scheduler [ Job_completed (status, job.ivar) ]
```
tmattio, line 457: Does the `Job_completed` event do anything special? If not we should probably kill the process when calling `remove`

fiber-unix/src/scheduler.ml
```diff
@@ -446,3 +418,174 @@ let cancel_timer (timer : timer) =
...
+    while not (wait_nonblocking_win32 t) do
+      Mutex.unlock t.mutex;
+      Thread.delay 0.001;
+      Mutex.lock t.mutex
+    done
+
+  let wait_unix t =
+    Mutex.unlock t.mutex;
+    let pid, status = Unix.wait () in
+    Mutex.lock t.mutex;
```
tmattio, line 507: Shouldn't we release the mutex after the `Process.remove` call?

fiber-unix/src/scheduler.ml
rgrinberg: Note that we kill remove on the process after we did `Unix.wait` on it. So it's already dead.

fiber-unix/src/scheduler.ml
rgrinberg: Perhaps it's a bit confusing but this is how it's done in dune. The next call is always `Condition.wait` which will unlock the mutex.

rgrinberg:
@tmattio this is ready

tmattio:
There are some pieces I didn't review thoroughly, will go through it tomorrow/Monday

ocaml-lsp-server/src/fmt.ml
```diff
@@ -1,40 +1,44 @@
...
+let run_command scheduler thread bin stdin_value args : command_result Fiber.t =
+  let open Fiber.O in
+  let stdin_i, stdin_o = Unix.pipe ~cloexec:true () in
+  let stdout_i, stdout_o = Unix.pipe ~cloexec:true () in
+  let stderr_i, stderr_o = Unix.pipe ~cloexec:true () in
+  let pid =
+    let args = Array.of_list (bin :: args) in
+    Unix.create_process bin args stdin_i stdout_o stderr_o |> Stdune.Pid.of_int
+  in
+  Unix.close stdin_i;
```
tmattio, line 18: Shouldn't the file descriptors be closed only when the process exits?

ocaml-lsp-server/src/fmt.ml
```diff
@@ -1,40 +1,44 @@
...
+    Scheduler.async_exn thread (fun () ->
+        let out_chan = Unix.out_channel_of_descr stdin_o in
+        output_string out_chan stdin_value;
+        flush out_chan;
+        close_out out_chan;
+        let stdout_in = Unix.in_channel_of_descr stdout_i in
+        let stderr_in = Unix.in_channel_of_descr stderr_i in
+        let stdout = Stdune.Io.read_all stdout_in in
+        close_in_noerr stdout_in;
+        let stderr = Stdune.Io.read_all stderr_in in
```
tmattio, line 31: IIUC, `Stdune.Io.read_all stdout_in` will block until `EOF` caused by the process exit. Then `Stdune.Io.read_all stderr_in` will read the content buffered in the content. This means that the buffer has to be large enough to store all of the stderr for the execution of a command. Maybe that's always true in practice (although I double it is on older Linux versions), but maybe to be safe, we could read `stdout` and `stderr` concurrently?

ocaml-lsp-server/src/fmt.ml
```diff
@@ -1,40 +1,44 @@
...
+        close_in_noerr stderr_in;
+        (stdout, stderr))
+  in
+  let+ status, (stdout, stderr) =
+    Fiber.fork_and_join
+      (fun () -> Scheduler.wait_for_process scheduler pid)
+      (fun () ->
+        let+ res = Scheduler.await_no_cancel task in
```
tmattio, line 39: Perhaps `await_no_cancel` could be modified to return the exit status? If figure you want it often enough and it would save the cost of spawning an additional thread.

ocaml-lsp-server/src/fmt.ml
```diff
@@ -87,17 +91,22 @@ let formatter doc =
...
+let run scheduler thread doc : (string, error) result Fiber.t =
+  let res =
+    let open Result.O in
+    let* formatter = formatter doc in
+    let args = args formatter in
+    let+ binary = binary formatter in
+    (binary, args, Document.source doc |> Msource.text)
+  in
+  match res with
+  | Error e -> Fiber.return (Error e)
```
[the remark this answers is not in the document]
rgrinberg: Yes, I wrote this code before 418. It was the straw that broke the camel's back.

ocaml-lsp-server/src/fmt.ml
rgrinberg: It's indeed more correct to do it concurrently. That's hot it was in the old code and it worked. I'll push a change to make it more robust. It will be necessary for the ocamlformat rpc anyway.

ocaml-lsp-server/src/fmt.ml
rgrinberg: > Perhaps await_no_cancel could be modified to return the exit status

It cannot be. All the waiting for processes to finish is done in a dedicated thread in the scheduler.

> If figure you want it often enough and it would save the cost of spawning an additional thread.

Note that neither of these fibers create a thread. The first fiber asks the process watcher thread to watch for the pid, while the second fiber is waiting for a task enqueued in an existing thread to finish.

ocaml-lsp-server/src/fmt.ml
rgrinberg: I don't think so, because we are our only closing the end of the pipe that we aren't intending to use.

ocaml-lsp-server/src/fmt.ml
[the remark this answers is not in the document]
rgrinberg: Tried doing it after, it makes the ocamlformat tests stall

ocaml-lsp-server/src/fmt.ml
rgrinberg: Okay, I fixed it to do all IO concurrently. This requires 3 threads, but our total thread count is still tiny.

rgrinberg:
@tmattio ping. I need this to go ahead with integrating dune.

tmattio (approved):

tmattio:
Sorry, forgot to approve, none of my comments were blockers :)

PR #418, opened 2021-04-16
Improve error handling

rgrinberg:
First of all I removed all the result types in LSP. The code much cleaner
without all this extra noise, and I apologize to the other contributors for
cargo culting this silly style.

I fixed how `Jsonrpc.Response.Error.E` are handled. They are now passed as responses
instead of being wrapped as internal errors.

Finally, exceptions are now reported to clients as soon as possible. This has
the advantage that the client will see errors sooner, but the disadvantage that
we aren't reporting all errors. I think it's a good trade off.

### DIFF
--- a/ocaml-lsp-server/src/code_actions/action_destruct.ml
+++ b/ocaml-lsp-server/src/code_actions/action_destruct.ml
@@ -26,9 +26,12 @@ let code_action doc (params : CodeActionParams.t) =
     match res with
     | Ok res -> Some (code_action_of_case_analysis uri res)
     | Error
-        ( Merlin_analysis.Destruct.Wrong_parent _ | Query_commands.No_nodes
-        | Merlin_analysis.Destruct.Not_allowed _
-        | Merlin_analysis.Destruct.Useless_refine
-        | Merlin_analysis.Destruct.Nothing_to_do ) ->
+        { exn =
+            ( Merlin_analysis.Destruct.Wrong_parent _ | Query_commands.No_nodes
+            | Merlin_analysis.Destruct.Not_allowed _
+            | Merlin_analysis.Destruct.Useless_refine
+            | Merlin_analysis.Destruct.Nothing_to_do )
+        ; backtrace = _
+        } ->
       None
-    | Error exn -> raise exn)
+    | Error exn -> Exn_with_backtrace.reraise exn)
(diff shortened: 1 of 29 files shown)


tmattio:
I really like the proposed changes for `ocaml-lsp-server`, however, I'm wondering about raising exceptions in `lsp` and other libraries, and in particular, how this will impact projects like `linol`? 

My rule of thumb is to prefer monadic error handling in libraries, but that might be misguided..

cc @c-cube

c-cube:
Linol should catch exceptions and turn them into lsp errors, at least for client requests. If not that's a bug :).

I personally don't mind exceptions in such projects because there's not much recovery you can do anyway. Reply with an error and move on, as long as the lsp server stays up.

rgrinberg:
Linol doesn't use lsp apart from type definitions, do this has change has zero impact on it anyway.

tmattio (approved):

ocaml-lsp-server/src/code_actions/action_destruct.ml
```diff
@@ -24,11 +24,11 @@ let code_action doc (params : CodeActionParams.t) =
...
+    | Ok res -> Some (code_action_of_case_analysis uri res)
     | Error
         ( Merlin_analysis.Destruct.Wrong_parent _ | Query_commands.No_nodes
         | Merlin_analysis.Destruct.Not_allowed _
         | Merlin_analysis.Destruct.Useless_refine
         | Merlin_analysis.Destruct.Nothing_to_do ) ->
-      Ok None
-    | Error exn -> Error (Jsonrpc.Response.Error.of_exn exn))
+      None
+    | Error exn -> raise exn)
```
ulugbekna, line 34: Will this raised exn have a backtrace without using `Exn_with_backtrace.capture`?

ulugbekna (approved):
I'm not sure I'm fond of relying on exceptions but the code is indeed cleaner :-)

ocaml-lsp-server/src/code_actions/action_destruct.ml
rgrinberg: Good call. I changed this and other instances to maintain backtraces systematically.

rgrinberg:
The last PR now gives us proper backtraces everywhere.

PR #419, opened 2021-04-17
Use ocaml/setup-ocaml@v2

smorimoto:
This PR introduces setup-ocaml v2, which greatly simplifies the workflow and makes it as fast as possible in a reasonable way using common caching features as well as things like dune caching.


.github/workflows/build-and-test.yml
```diff
@@ -1,91 +1,59 @@
...
-      - name: Check formatting
-        if: runner.os == 'Linux'
-        run: opam exec -- make fmt
+      - run: opam exec -- make test-ocaml
+        timeout-minutes: 3
 
-      - name: Run OCaml tests
-        run: opam exec -- make test-ocaml
+      - run: opam exec -- make test-e2e
+        timeout-minutes: 3
```
smorimoto, line 57: Set a timeout because of test instability on Windows and automatically fail the job if the issue occurs.

lsp-fiber/test/lsp_fiber_test.ml
```diff
@@ -209,7 +209,7 @@ let%expect_test "ent to end run of lsp tests" =
   [%expect.unreachable]
   [@@expect.uncaught_exn
     {|
-  ("Abort: Ppx_yojson_conv_lib__Yojson_conv.Of_yojson_error(_, 870828711)")
+  ("Abort: Jsonrpc: json conversion failed: invalid structured value")
```
rgrinberg, line 212: I fixed this bug. So please rebase to remove this change.

rgrinberg:
The changes to CI look good, but how come the lock files are removed?

.github/workflows/build-and-test.yml
```diff
@@ -1,91 +1,59 @@
...
-        working-directory: ocaml-lsp-server/test/e2e
-        run: |
-          yarn --frozen-lockfile
+      - run: opam install . --with-test
 
-      - name: Build
-        run: opam exec -- make all
+      - run: opam exec -- make all
 
-      - name: Check formatting
```
rgrinberg, line 83: Which action is doing the formatting check now?

smorimoto: It does at the last, but auto promote is currently enabled, so it's actually being formatted instead of checked. Somehow, the check-fmt task seems to have been removed at some point.

smorimoto:
> The changes to CI look good, but how come the lock files are removed?

The lock files were originally added to ensure consistency of dependencies in the CI, but they are not properly updated and only make things more complicated. With v2, we have a caching strategy that does not require lock files, so we can get rid of them.

lsp-fiber/test/lsp_fiber_test.ml
smorimoto: Done!

rgrinberg:
I see. @tmattio did you use the lock files in development?

tmattio:
> I see. @tmattio did you use the lock files in development?

Not really, I added them specifically for the CI. It's generally a good idea to have lock files, but I agree that they won't really be useful/convenient until they are generated by default by Opam.

.github/workflows/build-and-test.yml
rgrinberg: Even if auto promote is enabled, the error code should still indicate failure if there was a diff I think. In any case, it's better to verify that formatting errors fail the CI

smorimoto:
The formatting check worked as expected, at least the last time I made a mistake.

.github/workflows/build-and-test.yml
```diff
@@ -1,91 +1,72 @@
...
+        if: ${{ matrix.os == 'ubuntu-latest' && always() }}
+
+      - name: Run build
+        run: opam exec -- make all
 
-      - name: Run OCaml tests
+      - name: Run the unit tests
         run: opam exec -- make test-ocaml
+        if: always()
+        continue-on-error: ${{ matrix.os == 'windows-latest' }}
```
tmattio, line 65: I'm not sure this is a good idea: the Windows tests are broken, we shouldn't hide it from the CI

smorimoto: Well, we can always mark it as a failure, but it doesn't seem to be inherently different from marking it as passing.

tmattio: > doesn't seem to be inherently different from marking it as passing.
Because we don't fix the tests?

That's bad but hiding it makes it even worse 😅

.github/workflows/build-and-test.yml
smorimoto: If "always" is enabled, the job is finally marked as failed after running to the last step, so the current configuration seems quite reasonable.

.github/workflows/build-and-test.yml
smorimoto: I don't have a strong opinion, but I don't really like to give the impression that "it doesn't work on Windows!" by marking it as a failure.

smorimoto:
Okay, this is now ready to merge.

ocaml-lsp-server/src/dune
```diff
@@ -5,7 +5,6 @@
  (libraries cmdliner dune-build-info fiber fiber_unix jsonrpc lsp lsp_fiber
    merlin.analysis merlin.kernel merlin.merlin_utils merlin.parsing
    merlin.query_commands merlin.query_protocol merlin.specific merlin.typing
-   merlin.utils octavius omd ppx_yojson_conv_lib result stdune yojson
-   dune_rpc))
+   merlin.utils octavius omd ppx_yojson_conv_lib result stdune yojson))
```
rgrinberg, line 8: Was this change accidental? We do in fact need to depend on dune_rpc

smorimoto: Yes, I dropped that commit.

smorimoto:
Okay, we can merge this now. actions-ml has been completely deprecated, and should be merged this as soon as possible.

tmattio (approved):
Still marking Windows failure as success, but I'll add a note to myself to fix the tests 🙂

smorimoto:
I just fixed errors caused by EOL and now marked errors as error

rgrinberg (approved):

smorimoto:
I don't have merge right here, can someone merge?

PR #424, opened 2021-04-21
Implement a tracer

### LINKED ISSUE #388: Collect request statistics
To improve lsp's performance, we need to have a better understanding which requests are slow, and what is slowing them down. To begin doing this, we need a stats collection mechanism in lsp. We should use dune's `chrome_trace` library to record events for the last N requests served. We could consider adding some custom requests to dump this information somewhere that the user can submit to us.

ulugbekna:
#### Events traced:
- currently traces only receiving a request and sending a response on `lsp-fiber` layer
- only the last N events are stored (N = 1000 and is not configurable _for now_)
- what other events to trace? 
    - notifications? sent AND received? 
    - keep initialization and shutdown requests separately? (they are important to understand who client is and the system's lifetime)

#### Getting output of tracing:
- Currently the PR implements a custom request "ocamllsp/dumpTrace" that returns the last N events in the system
- There is a custom request implement in vscode extension. For now, the extension cannot dump the trace into a file, but a user should be able to do that

#### Library layers
I am also trying to separate out layers (ocaml-lsp, lsp-fiber, and jsonrpc) to see maybe at which step we spend time. 

#### Example
_How `chrome://tracing` looks like for 1000 events trace:_

image

To-do:
- [ ] store request/response as json's (`Chrome_trace.Json.t` is much limited compared to `Yojson.Safe.t`, so we cannot use json request/response directly as event `args`, we need to convert first)
- [ ] I think we need to `Chrome_trace.Event.Timestamp.now () - base time` for events
- [ ] We need better classification of requests, e.g., by files? 
- [ ] add tracing to other layers 
- [ ] add tracing for various events
- [ ] add tests? 
- [ ] things that come up along the way...

I'm slightly hesitant about this implementation. I hope it makes sense.


tracer/tracer.mli
```diff
@@ -0,0 +1,30 @@
...
+
+val trace_layers :
+  ?app:unit -> ?lsp:unit -> ?jsonrpc:unit -> unit -> trace_layers
+
+val create : max_n_last_events:int -> trace_layers -> t
+
+val trace_lsp_recv_req :
+     ?name:string
+  -> t option
+  -> id:[ `String of string | `Int of int ]
```
rgrinberg, line 17: You can use `Jsonrpc.Id.t` directly here and elsewhere.

tracer/tracer.mli
```diff
@@ -0,0 +1,30 @@
+type t
+
+type trace_layers =
+  { app : bool
+  ; lsp : bool
+  ; jsonrpc : bool
+  }
+
+val trace_layers :
+  ?app:unit -> ?lsp:unit -> ?jsonrpc:unit -> unit -> trace_layers
```
rgrinberg, line 10: I'd pass the flags as booleans.

ocaml-lsp-server/src/custom_requests/req_dump_trace.ml
```diff
@@ -0,0 +1,15 @@
+open Import
+
+let meth = "ocamllsp/dumpTrace"
+
+let on_request ~params:_ state =
```
rgrinberg, line 5: How do you use this request? Do you paste the result from the response somewhere and then open chrome?

tracer/tracer.mli
ulugbekna: I am afraid of creating a cycle between `tracer` and `jsonrpc`. I want to trace events in jrpc with the tracer, if that makes sense.

rgrinberg: `jsonrpc` has only type definitions, so I don't see how it could ever depend back on tracer.

ocaml-lsp-server/src/custom_requests/req_dump_trace.ml
ulugbekna: I created a [PR](https://github.com/ocamllabs/vscode-ocaml-platform/pull/588) in vscode-ocaml-platform, which implements a command `Dump Trace` that can be called from the command palette. You should then be able to get the response from output pane either for ocaml-lsp or vscode-ocaml-platform -- both should print that.

Aside: I would generally welcome suggestions on when to enable tracing, how to set N, any suggestions on making tracing nicer...

tracer/dlist.mli
```diff
@@ -0,0 +1,21 @@
+(** Doubly linked list with all operations constant time *)
```
rgrinberg, line 1: I would have just created a ring of fixed size backed by an array. However if this already works, there's no need to change it.

tracer/tracer.mli
ulugbekna: The function sets the layer to true if the argument with that name is passed. 

I can change the flags to booleans and make the function to take true/false into account, ie user could pass `~app:false` explicity if they don't want tracing.

rgrinberg: Yes, that's what I prefer.

tracer/tracer.mli
ulugbekna: ah right! will fix where appropriate

ocaml-lsp-server/src/custom_requests/req_dump_trace.ml
rgrinberg: I see. If there's time, I'd experiment with doing it in an editor agnostic way. E.g. send a request to the editor to open a  temp file with the tracing info.

tracer/tracer.ml
```diff
@@ -0,0 +1,107 @@
...
+  let trace tracer =
+    let name = Option.value name ~default:meth in
+    let common_fields = req_resp_event_common_fields ~name in
+    let id = event_id_of_jsonrpc_id id in
+    let args =
+      if should_capture_req tracer ~me:`Lsp then
+        []
+      else
+        let params = Option.value params ~default:(`Assoc []) in
+        [ ("request", `String (Yojson.Safe.pretty_to_string params)) ]
```
rgrinberg, line 83: Is it required to stringify the json? We can't just pass the json payload directly?

ulugbekna: This is one of the todos. Yojson.Safe.t can't be coerced to Chrome_trace.Json.t, so as a workaround for now, I used a stringified json. I'll need to convert from one json type to another manually to have a proper json in the event args

rgrinberg:
> store request/response as json's (Chrome_trace.Json.t is much limited compared to Yojson.Safe.t, so we cannot use json request/response directly as event args, we need to convert first)

You can just remove `Chrome_trace.Json.t` in our fork and use `Yojson` directly. It was a pain to use yojson in dune because of cppo, but we already use it here.

> I think we need to Chrome_trace.Event.Timestamp.now () - base time for events

What kind of events do you have in mind? As far as I see, you're using Async everywhere. This should do the duration bookmarking automatically for you.

ulugbekna:
> You can just remove Chrome_trace.Json.t in our fork and use Yojson directly. 

Sure, will do. 

> What kind of events do you have in mind? As far as I see, you're using Async everywhere. This should do the duration bookmarking automatically for you.

You're right. I wasn't first sure whether timestamps are "normalized" so that first event occurs near "0ms" in chrome://tracing, but they should be.

PR #425, opened 2021-04-22
fix crash when trace is set to messages

tatchi:
Should fix https://github.com/ocamllabs/vscode-ocaml-platform/issues/587

### DIFF
--- a/lsp/bin/specifications/specification-3-16.md
+++ b/lsp/bin/specifications/specification-3-16.md
@@ -1653,7 +1653,7 @@ A `TraceValue` represents the level of verbosity with which the server systemati
 The initial trace value is set by the client at initialization and can be modified later using the [$/setTrace](#setTrace) notification.
 
 ```typescript
-export type TraceValue = 'off' | 'message' | 'verbose';
+export type TraceValue = 'off' | 'messages' | 'verbose';
 ```
 
 ### Actual Protocol
--- a/lsp/src/types.ml
+++ b/lsp/src/types.ml
@@ -26796,24 +26796,24 @@ end
 module TraceValue = struct
   type t =
     [ `Off
-    | `Message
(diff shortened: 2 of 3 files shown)


tatchi:
I can confirm that with this change, traces are correctly printed in the VScodes's output window

image

rgrinberg:
[According to the spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#traceValue), `"message"` is correct.

tatchi:
> [According to the spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#traceValue), `"message"` is correct.

Interesting! Looks like that was a change introduced in the current version (3.16) since the previous spec (3.15) uses `messages`: https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#initialize

rgrinberg:
Perhaps we need to upgrade our libraries in vscode?

tatchi:
> Perhaps we need to upgrade our libraries in vscode?

That's what I was looking into indeed. We use [version 7 of the `vscode-languageclient`](https://github.com/ocamllabs/vscode-ocaml-platform/blob/master/package.json#L843) which appears [to be the latest](https://www.npmjs.com/package/vscode-languageclient)

Also looking at the lsp code, `messages` looks to be correct: https://github.com/microsoft/vscode-languageserver-node/blob/main/jsonrpc/src/common/connection.ts#L205

That's also what typescript uses:

Screenshot 2021-04-22 at 07 25 49

That's probably an error in their doc?

rgrinberg:
Okay, can you raise an issue in the spec repo about this please?

giltho:
[That's on me](https://github.com/microsoft/language-server-protocol/pull/953/files) sorry...

I'll write the PR to fix that

PR #427, opened 2021-04-26
Fix extracting path from URI

rgrinberg:
see if this works @mnxn

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ lock: ## Generate the lock files
 
 .PHONY: test
 test-ocaml: ## Run the unit tests
-	dune build @lsp-fiber/runtest @fiber-unix/runtest @jsonrpc-fiber/runtest
+	dune build @lsp/test/runtest @lsp-fiber/runtest @fiber-unix/runtest @jsonrpc-fiber/runtest
 
 .PHONY: promote
 promote:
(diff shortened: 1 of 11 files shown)


rgrinberg:
I made the URI parsing a little better and removed the code that caused this bug altogether.

mnxn:
I think there are too many slashes at the beginning now:
```
[Trace - 1:46:20 PM] Sending request 'ocamllsp/switchImplIntf - (5)'.
Params: [
    "file:///c:/Users/Max/Projects/vscode-ocaml-platform/src/ocaml_lsp.mli"
]


[Trace - 1:46:20 PM] Received response 'ocamllsp/switchImplIntf - (5)' in 1ms.
Result: [
    "file:////c%3A/Users/Max/Projects/vscode-ocaml-platform/src/ocaml_lsp.ml"
]
```

Developer console:
```
[UriError]: If a URI does not contain an authority component, then the path cannot begin with two slash characters ("//")
```

mnxn:
@rgrinberg: I think there might be a problem with this patch too. 

My projects are filled with "Unbound module" errors and there are no errors/exception in the log or developer console.
I originally thought it was a problem with my projects, but release 1.5 doesn't have this problem.

The first commit of this PR (https://github.com/ocaml/ocaml-lsp/pull/427/commits/527d931cba7e0368eeca5353586ceb9c8d1013e7) had neither problem.

PR #429, opened 2021-04-28
Enable short paths by default

rgrinberg:
And also fix the ensuing bug in pretty printing of signatures


voodoos:
You might want to remove the workaround "fixing" the bug, we have now a more robust fix in https://github.com/ocaml/merlin/pull/1334

rgrinberg:
The last commit changes the code to use an environment consistent with how `ocamlc -i` works.

tmattio (approved):
I'm not familiar enough with Merlin to give an enlightened review, but this LGTM from the regression test

voodoos (approved):
LGTM too

PR #431, opened 2021-05-02
Add nix based dev env

rgrinberg:


rgrinberg:
`nix-shell` should now open a shell with everything that you need to run tests (after running yarn to fetch deps).

tmattio:
I'd be interested to know the motivation for this PR as well as Dune's one. I'm guessing this is to address shortcomings in an opam-based workflow and there's probably things we can learn from it, and maybe open issues on Opam's repo, I'm sure the maintainers would be happy to receive some feedback 🙂

rgrinberg:

> I'd be interested to know the motivation for this PR as well as Dune's
> one. I'm guessing this is to address shortcomings in an opam-based
> workflow and there's probably things we can learn from it, and maybe
> open issues on Opam's repo, I'm sure the maintainers would be happy to
> receive some feedback 🙂 

Yes, there are shortcomings in the opam workflow and it's highly
unlikely that they can be addressed. The constraints and resources
behind opam and nix are totally different.

Here's a short list of problems that nix solves for me:

* Manages the dependencies of non OCaml packages. Quite important for
  complex packages like dune or lsp.

* Deduplicates binaries across projects. I no longer need to install a
  separate ocamlformat/lsp/ocp-indent for every single development
  environment.  Nix deduplicates everything with its "store".

* The nix paradigm gives a full programming language to solve package
  management problems. Yes, it's a crappy untyped DSL, but it's so much
  better than gluing things together with the opam binary and shell
  script. Try overriding metadata in one of your packages for example.

Nix has its own problems, but it's such a powerful tool that almost any
problem can be worked around. I highly recommend to try it.

PR #433, opened 2021-05-03
Revert verbosity back to 0

rgrinberg:
This matches merlin's default behavior and prevents stack overflows in some
situations.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+# 1.6.1 (05/03/2020)
+
+## Fixes
+
+- Switch `verbosity` from 1 to 0. This is the same default that merlin uses.
+  The old value for verbosity (#433)
+
 # 1.6.0 (04/30/2020)
 
 ## Features
--- a/ocaml-lsp-server/src/document.ml
+++ b/ocaml-lsp-server/src/document.ml
@@ -136,7 +136,7 @@ let make_config uri =
   let mconfig =
     { mconfig with
       ocaml = { mconfig.ocaml with real_paths = false }
-    ; query = { mconfig.query with verbosity = 1; filename; directory }
+    ; query = { mconfig.query with filename; directory }
     }
   in
   Mconfig.get_external_config path mconfig


ulugbekna:
I understand usefulness of having verbosity set to 0, but I think this change needs to come along with a way to display a module's type in a useful way. 

Going from this 

image

to this 

image

would be a regression in hover functionality, don't you think? 

### What to do

1) Maybe we could detect the ast node in merlin and if it's a module, use verbosity = 1 
2) OR look at the ident being hovered over in ocaml-lsp and send a request with an apt verbosity? 

I think we should aim to have the next release to have both this PR and a fix for modules. What do you think?

rgrinberg:
> would be a regression in hover functionality, don't you think?

I think the unrecoverable crash is worse than the loss of functionality.

> I think we should aim to have the next release to have both this PR and a fix for modules. What do you think? 

Yes, please give it a try.

voodoos:
@ulugbekna :
> I think we should aim to have the next release to have both this PR and a fix for modules. What do you think?

Using the master version of OCaml-LSP is indeed less useful since it does not gives the signature of module.
Have you had time to look into that issue ?

ulugbekna:
@voodoos I completely forgot about this. I will look into it and report back.

ulugbekna:
@voodoos (cc @rgrinberg)

An interesting side note: you see, when we hover over a module pulled from a library, say, `Fiber` in ocaml-lsp or `Mvar` in `Fiber.Mvar.create`, we see the expanded type with the module's items, but if we hover over a module that's pulled from a local library or a same-directory file, you get `(module <modulename>)`. It gets even more interesting with `Stdlib.Hashtbl.create`, where `Stdlib` is expanded, but not `Hashtbl`. Do you happen to know how merlin treats differently "local" modules vs modules from other libs?

The simplest solution to the problem we have I see (without fixing the problem I described in the side note above): we reconstruct the (shortest) current identifier at the given position; here, by shortest I mean `Mvar` in `Fiber.M<|>var.create` where `<|>` is the cursor. If identifier starts with a capital letter, we send verbosity=1 request; otherwise, the default. Of course, in this solution we do not differentiate between a module and a value constructor, but 
1) how bad is that we use verbosity 1 vs 0 for value constructors? 
2) fixing this problem seems to dive quite deep into merlin source (I think merlin does this in `type_in_env` using `Context`?) 

Do you have better ideas? 

Other unrelated questions I have: 

I) can merlin offer auto-complete for polymorphic variants? 
II) Shouldn't [`Compl.prefix_of_position`](https://github.com/ocaml/ocaml-lsp/blob/0e49ad29935a15e524d6ea94b91ef466adf5dcf8/ocaml-lsp-server/src/compl.ml#L49-L75) allow a back tick in the prefix?

voodoos:
I guess some input from @trefis might be profitable here.

trefis:
> Maybe we could detect the ast node in merlin and if it's a module, use verbosity = 1

It'd be easier if merlin did that itself. We could add a `-verbosity smart` flag.

> OR look at the ident being hovered over in ocaml-lsp and send a request with an apt verbosity?

Not sure what exactly you mean. But the intended behavior for verbosity is that the editor would first set it to 0, and then if the user issues the same query at the same position (with no other query in between) increase the verbosity by 1.

>  Do you happen to know how merlin treats differently "local" modules vs modules from other libs?

I think you got your diagnostic slightly wrong.
Merlin will print you the type of the module, sometimes that type is a signature, sometimes it's just an alias to something else.

A higher value of verbosity asks merlin to traverse the alias.

trefis:
> can merlin offer auto-complete for polymorphic variants?

It used to work, at least in the context of function applications: if the current argument is supposed to be a polymorphic variant constructor, then completion would offer you the choice between the accepted ones.
Apparently that's broken, I don't know since when.

PR #434, opened 2021-05-03
textDocument/formatting: return diff of changes

actionshrimp:
Uses a simple diff algorithm to return a list of `TextEdit`s  for the delta from `ocamlformat`/`refmt`, rather than just returning a single `TextEdit` that replaces the whole file.

Why
===

My particular use case (from emacs with evil-mode):
- In my case I use emacs, and have it configured to format-on-save
- I noticed I had slow performance on some edits using `evil-mode`(/ spacemacs), which became almost unusable on large files, which I'd assumed was due to `tuareg`/`font-lock`. In particular, `O`  and `dd` would take a long time O(1s) to complete, which is pretty jarring.
- On closer investigation, I noticed performance was initially fine on these large files, but poor editing performance started after the first format of the buffer.
- The reason for this was `evil-mode` calling the `field-beginning` function as part of `evil-narrow-to-field` - calling `field-beginning` manually would be very fast before doing a format, then be the cause of the slowness afterwards.
- Intuitively this seems to be because `evil-mode` is using buffer information to figure out how indentation / other things (?) should work - sometimes this buffer information is based on recent nearby changes, and rewriting the whole file makes calculating this information less efficient.
- Another example of similar slowness from `field-beginning` around large-scale buffer changes is documented here: https://github.com/ProofGeneral/PG/issues/427#issuecomment-500612904

Generally:
- Admittedly, this might seem like a problem with emacs rather than `ocaml-lsp` itself, but intuitively it seems like this idea of the editor using change history to calculate things about the file might generalise to different editors, and `ocaml-lsp` is being a better LSP citizen by telling the editor to do less work 
- Other LSPs also calculate a diff, e.g. https://github.com/golang/tools/blob/a1fbb68135aa978db34613d32e259c0121a69e93/internal/lsp/diff/myers/diff.go

Implementation
===

I tried to find a simple algo to perform the diff that would work for both `ocamlformat` and `refmt`, and found `simple-diff` - this seemed to fit the bill, as I don't think that `textDocument/formatting` is ever used in batch and is more for formatting a single file at a time, so performance of the diff itself isn't critical. 

This seemed better than a more optimised diffing algorithm to avoid bringing in dependencies (e.g. with `patience_diff`) or depending on system binaries like `diff` being available, as they might not be available on Windows.

However, `simple-diff` seems to have a few bugs, so I inlined it and fixed them. I'd already gone down quite the rabbit hole at this point, so this seemed like the simplest solution!

--

I have tested this in emacs and it seems to make performance a lot better in large buffers after the first instance of `textDocument/formatting` - hopefully this is appealing in itself and is worth the additional complexity,

Very happy to rework the implementation if my assumptions there are off at all - particularly if I've overlooked an easy alternative for the algo, or if `textDocument/formatting` is designed to be used on a large number of files simultaneously (I'm not super familiar with LSP), in which case a more efficient diff algo would probably be beneficial.

### DIFF
--- a/ocaml-lsp-server/src/fmt.ml
+++ b/ocaml-lsp-server/src/fmt.ml
@@ -126,7 +126,175 @@ let exec state bin args stdin =
   | Unix.WEXITED 0 -> Result.Ok res.stdout
   | _ -> Result.Error (Unexpected_result { message = res.stderr })
 
-let run state doc : (string, error) result Fiber.t =
+module Simple_diff = struct
+  (* based on *)
+  (* https://github.com/paulgb/simplediff/blob/031dc772ca6795cfdfed27384a6b79e772213233/python/simplediff/__init__.py *)
+
+  type item = string
+
+  type diff =
+    | Deleted of item array
+    | Added of item array
+    | Equal of item array
+
+  let longest_subsequence old_lines new_lines =
+    let _, old_index_map =
+      old_lines
+      |> Array.fold_left ~init:(0, String.Map.empty) ~f:(fun (i, m) line ->
+             ( i + 1
+             , String.Map.update
+                 ~f:(function
+                   | None -> Some [ i ]
+                   | Some xs -> Some (i :: xs))
+                 m line ))
+    in
+    let overlap = ref Int.Map.empty in
+
+    let sub_start_old = ref 0 in
+    let sub_start_new = ref 0 in
+    let sub_length = ref 0 in
+
+    Array.iteri new_lines ~f:(fun inew v ->
+        let overlap' = ref Int.Map.empty in
+        let old_indices =
+          String.Map.find old_index_map v |> Option.value ~default:[]
+        in
+        old_indices
+        |> List.iter ~f:(fun iold ->
+               let o =
+                 1
+                 + (Int.Map.find !overlap (iold - 1) |> Option.value ~default:0)
+               in
+               overlap' := Int.Map.set !overlap' iold o;
+
+               if o > !sub_length then (
+                 sub_length := o;
+                 sub_start_old := iold - o + 1;
+                 sub_start_new := inew - o + 1
+               ));
+
+        overlap := !overlap');
+
+    (!sub_start_new, !sub_start_old, !sub_length)
+
+  let get_diff old_lines new_lines =
+    let rec get_diff' old_lines new_lines =
+      match (old_lines, new_lines) with
+      | [||], [||] -> []
+      | old_lines, [||] -> [ Deleted old_lines ]
+      | [||], new_lines -> [ Added new_lines ]
+      | _, _ ->
+        let sub_start_new, sub_start_old, sub_length =
+          longest_subsequence old_lines new_lines
+        in
+        if sub_length = 0 then
+          [ Deleted old_lines; Added new_lines ]
+        else
+          let old_lines_presubseq =
+            Array.sub ~pos:0 ~len:sub_start_old old_lines
+          in
+          let new_lines_presubseq =
+            Array.sub ~pos:0 ~len:sub_start_new new_lines
+          in
+          let old_lines_postsubseq =
+            let start_index = sub_start_old + sub_length in
+            let len = Array.length old_lines - start_index in
+            Array.sub ~pos:start_index ~len old_lines
+          in
+          let new_lines_postsubseq =
+            let start_index = sub_start_new + sub_length in
+            let len = Array.length new_lines - start_index in
+            Array.sub ~pos:start_index ~len new_lines
+          in
+          let unchanged_lines =
+            Array.sub ~pos:sub_start_new ~len:sub_length new_lines
+          in
+          get_diff' old_lines_presubseq new_lines_presubseq
+          @ [ Equal unchanged_lines ]
+          @ get_diff' old_lines_postsubseq new_lines_postsubseq
+    in
+    get_diff' (Array.of_list old_lines) (Array.of_list new_lines)
+end
+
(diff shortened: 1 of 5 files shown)


ocaml-lsp-server/src/fmt.ml
```diff
@@ -126,7 +126,182 @@ let exec state bin args stdin =
...
+let insert ~line_sep ~line added_lines =
+  let start = { Position.character = 0; line } in
+  let end_ = { Position.character = 0; line } in
+  let range = { Range.start; end_ } in
+  let newText =
+    (added_lines |> Array.to_list |> String.concat ~sep:line_sep) ^ line_sep
+  in
+  { TextEdit.newText; range }
+
+let replace ~line_sep ~line deleted_lines added_lines =
```
rgrinberg, line 138: `insert`, `replace`, and `delete` are quite similar. Can you combine them somehow?

```
type edit =
   | Add of ..
   | Replace of ..
   | Delete of ..
val text_edit : line:int -> line_sep:string -> edit -> TextEdit.t
```

ocaml-lsp-server/src/fmt.ml
```diff
@@ -126,7 +126,182 @@ let exec state bin args stdin =
...
+
+  type item = string
+
+  type diff =
+    | Deleted of item array
+    | Added of item array
+    | Equal of item array
+
+  module StringMap = Map.Make (Stdlib.String)
+  module IntMap = Map.Make (Stdlib.Int)
```
rgrinberg, line 171: `String.Map` and `Int.Map`. This might require editing `Import` to add the necessary imports.

ocaml-lsp-server/src/fmt.ml
```diff
@@ -126,7 +126,182 @@ let exec state bin args stdin =
...
+           (0, StringMap.empty)
+    in
+    let overlap = ref IntMap.empty in
+
+    let sub_start_old = ref 0 in
+    let sub_start_new = ref 0 in
+    let sub_length = ref 0 in
+
+    new_lines
+    |> Array.iteri (fun inew v ->
```
rgrinberg, line 193: You can `module Array = Array` to `import.ml` and use the labeled version.

ocaml-lsp-server/src/fmt.ml
```diff
@@ -126,7 +126,182 @@ let exec state bin args stdin =
...
+           in
+           old_indices
+           |> List.iter ~f:(fun iold ->
+                  let o =
+                    1
+                    + (IntMap.find_opt (iold - 1) !overlap
+                      |> Option.value ~default:0)
+                  in
+                  overlap' :=
+                    IntMap.update iold
```
rgrinberg, line 206: `Int.Map.set`

ocaml-lsp-server/src/fmt.ml
```diff
@@ -136,4 +311,8 @@ let run state doc : (string, error) result Fiber.t =
   in
   match res with
   | Error e -> Fiber.return (Error e)
-  | Ok (binary, args, contents) -> exec state binary args contents
+  | Ok (binary, args, contents) ->
+    exec state binary args contents
+    |> Fiber.map ~f:(function
+         | Ok formatted -> Ok (diff_edits contents formatted)
```
rgrinberg, line 317: `Result.map`.

ocaml-lsp-server/src/fmt.ml
```diff
@@ -126,7 +126,182 @@ let exec state bin args stdin =
...
+  in
+  let line, prev_deleted_lines, edits_rev =
+    Simple_diff.get_diff orig_lines formatted_lines
+    |> List.fold_left
+         ~init:(0, [||], [])
+         ~f:(fun (line, prev_deleted_lines, edits_rev) x ->
+           match x with
+           | Simple_diff.Deleted deleted_lines ->
+             ( line
+             , Array.concat [ prev_deleted_lines; deleted_lines ]
```
rgrinberg, line 271: `Array.append`.

rgrinberg:
Looks good - just a few minor comments. This was on my todo list for the longest time and I'm glad someone picked it up. For optimal experience in Emacs, we only need a way to integrate with https://github.com/raxod502/apheleia.

actionshrimp:
Fantastic, thanks for taking a look - have pushed an update with those updates.

rgrinberg:
Thanks. The PR is working great for me, though I don't notice much difference in performance. Possibly because I use eglot? I just made a few more stylistic changes to match the house style.

actionshrimp:
I only noticed the performance increase on files > 500 lines or so (and worse the larger they get). Also regular, non-evil-mode didn't seem to be impacted in the same way, not sure whether or not you use that?

This was with lsp-mode, yeah - perhaps lsp-mode is adding some spans behind the scenes that are causing the slowdown in my case. I didn't know about eglot so thanks for the heads-up, I'll have to try that out at some point!

rgrinberg:
> I only noticed the performance increase on files > 500 lines or so
> (and worse the larger they get). Also regular, non-evil-mode didn't
> seem to be impacted in the same way, not sure whether or not you use
> that?

I usually format types.ml in this repository without much delay (1<
second). I do use evil mode.

> This was with lsp-mode, yeah - perhaps lsp-mode is adding some spans
> behind the scenes that are causing the slowdown in my case. I didn't
> know about eglot so thanks for the heads-up, I'll have to try that out
> at some point! 

Quite possibly. I switched from lsp-mode to eglot because of all the
performance pathologies in lsp-mode. Eglot is more bare bones, but seems
to be more reliable.

PR #435, opened 2021-05-04
Start dune and detect version

rgrinberg:
We attempt to find dune and check that it's recent enough for RPC
functionality. No new functionality is added, this PR is mostly to avoid
conflicts for future PR's.

### DIFF
--- a/ocaml-lsp-server/src/dune.ml
+++ b/ocaml-lsp-server/src/dune.ml
@@ -1,0 +1,46 @@
+open Import
+
+type state =
+  | Binary_not_found
+  | Out_of_date
+  | Running
+
+type t =
+  { scheduler : Scheduler.t
+  ; state : state Fiber.Ivar.t
+  }
+
+let create scheduler = { scheduler; state = Fiber.Ivar.create () }
+
+let state t : state Fiber.t =
+  let open Fiber.O in
+  let* state = Fiber.Ivar.peek t.state in
+  match state with
+  | Some s -> Fiber.return s
+  | None -> (
+    match Bin.which "dune" with
+    | None ->
+      let res = Binary_not_found in
+      let+ () = Fiber.Ivar.fill t.state res in
+      res
+    | Some bin ->
+      let bin = Fpath.to_string bin in
+      let dev_null () = Unix.openfile "/dev/null" [ Unix.O_RDWR ] 0o666 in
+      let stdin = dev_null () in
+      let stdout = dev_null () in
+      let stderr = dev_null () in
+
+      let open Fiber.O in
+      let pid =
+        let args = Array.of_list [ bin; "rpc"; "--help=plain" ] in
+        Unix.create_process bin args stdin stdout stderr |> Stdune.Pid.of_int
+      in
+      let* status = Scheduler.wait_for_process t.scheduler pid in
+      let state =
+        match status with
+        (* TODO actually turn on *)
+        | Unix.WEXITED 0 -> Running
+        | _ -> Out_of_date
+      in
+      let+ () = Fiber.Ivar.fill t.state state in
+      state)
--- a/ocaml-lsp-server/src/dune.mli
+++ b/ocaml-lsp-server/src/dune.mli
@@ -1,0 +1,12 @@
+open Import
+
+type state =
+  | Binary_not_found
+  | Out_of_date
+  | Running
+
+type t
+
+val state : t -> state Fiber.t
+
+val create : Scheduler.t -> t
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -835,6 +835,7 @@ let start () =
   in
   let configuration = Configuration.default in
   let detached = Fiber.Pool.create () in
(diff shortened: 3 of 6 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -846,18 +847,44 @@ let start () =
...
+              | Out_of_date ->
+                Some
+                  "Dune is out of date. Install dune >= 3.0 for project \
+                   functionality"
+              | _ -> None
+            in
+            match message with
+            | None -> Fiber.return ()
+            (* We disable the warnings because dune 3.0 isn't available yet *)
+            | Some _ when true -> Fiber.return ()
```
tmattio, line 876: `Some _ when true`? Did you forget to remove this?

ocaml-lsp-server/src/dune.ml
```diff
@@ -0,0 +1,45 @@
+open Import
+
+type state =
+  | Absent
+  | Out_of_date
+  | Running
```
tmattio, line 6: What if it's there but is not running, or crashed?

ocaml-lsp-server/src/dune.ml
```diff
@@ -0,0 +1,45 @@
...
+  | Some s -> Fiber.return s
+  | None -> (
+    match Bin.which "dune" with
+    | None ->
+      let+ () = Fiber.Ivar.fill t.state Absent in
+      Absent
+    | Some bin ->
+      let bin = Fpath.to_string bin in
+      let dev_null () = Unix.openfile "/dev/null" [ Unix.O_RDWR ] 0o666 in
+      let stdin = dev_null () in
```
tmattio, line 28: Did you want to leave out the ability to send requests and receive responses from this PR?

rgrinberg:

> `Some _ when true`? Did you forget to remove this?

It's because there's no released version of dune yet, so the warning
would be pesky.

> What if it's there but is not running, or crashed?

Note that "running" refers to the status of the RPC client started by
LSP. So hopefully that should never crash. If dune isn't running, the
rpc client will notify us.

> Did you want to leave out the ability to send requests and receive
> responses from this PR?

Yes. Making actual requests is saved for another PR that will not be
merged in a while. This will require the unreleased client library, so I
don't want master to depend on it just yet.

tmattio:
Thanks for the replies, the intent of the PR is a bit clearer now and it looks good to me 🙂

tmattio (approved):

ocaml-lsp-server/src/dune.mli
```diff
@@ -0,0 +1,12 @@
+open Import
+
+type state =
+  | Absent
+  | Out_of_date
+  | Running
```
ulugbekna, line 6: 1. I'm finding name `Absent` slightly confusing, as to it's not very clear where it is absent from, especially in the presence of another variant `Running`. Maybe `Not_found` or `Not_found_on_system` (long but very clear)? 
2. Separate error cases from success cases by using `result`?

ocaml-lsp-server/src/dune.mli
```diff
@@ -0,0 +1,12 @@
...
+
+type t
+
+val state : t -> state Fiber.t
+
+val create : Scheduler.t -> t
```
ulugbekna, line 12: 3. Isn't it quite unexpected that the function `state` can start a process and mutate things? Yes, we return a fiber, but it still seems suspicious. Wdyt?
4. (minor) While use of `create` would signal that some mutation (launching a process) is happening (vs `make`, for example).

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -846,18 +847,44 @@ let start () =
...
+            | Some message ->
+              let* _ = Server.initialized server in
```
ulugbekna, line 878: Annotate type for `_`?

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -846,18 +847,44 @@ let start () =
...
+            let* state = Dune.state dune in
+            let message =
+              match state with
+              | Absent ->
+                Some "Dune must be installed for project functionality"
+              | Out_of_date ->
+                Some
+                  "Dune is out of date. Install dune >= 3.0 for project \
+                   functionality"
+              | _ -> None
```
ulugbekna, line 871: 5. could either expand `_` or separate success/error cases I mentioned in (2)?

ocaml-lsp-server/src/dune.mli
rgrinberg: 3. It cannot mutate things in a way that is observable to users of the module.
4. That's how I did it initially but it came with its own problems. Like now we need to wait for `create` to complete before we can finish initializing the server.

ocaml-lsp-server/src/dune.mli
rgrinberg: 1. `Not_found` sounds good to me.
2. It's worth doing if I use any of the functions in `Result`. I also wouldn't call these error states. Just states where the functionality is diminished.

PR #437, opened 2021-05-05
reproduce wrong-doc-on-hover bug from #344

ulugbekna:
Does what's said on the tin :p 

See #344 for issue description and when the bug occurs.

### DIFF
--- a/ocaml-lsp-server/test/e2e/__tests__/textDocument-hover.test.ts
+++ b/ocaml-lsp-server/test/e2e/__tests__/textDocument-hover.test.ts
@@ -352,4 +352,134 @@ let x : foo = 1
       }
     `);
   });
+
+  it("FIXME: reproduce [#344](https://github.com/ocaml/ocaml-lsp/issues/344)", async () => {
+    languageServer = await LanguageServer.startAndInitialize({
+      textDocument: {
+        hover: {
+          dynamicRegistration: true,
+          contentFormat: ["markdown", "plaintext"],
+        },
+        moniker: {},
+      },
+    });
+
+    await languageServer.sendNotification("textDocument/didOpen", {
+      textDocument: Types.TextDocumentItem.create(
+        "file:///test.ml",
+        "ocaml",
+        0,
+        // the empty space below is necessary to reproduce the bug
+        outdent`
+
+
+
+
+
+
+
+
+
+
+
+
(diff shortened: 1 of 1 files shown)


ocaml-lsp-server/test/e2e/__tests__/textDocument-hover.test.ts
```diff
@@ -352,4 +352,134 @@ let x : foo = 1
       }
     `);
   });
+
+  it("FIXME: reproduce [#344](https://github.com/ocaml/ocaml-lsp/issues/344)", async () => {
```
ulugbekna, line 356: note: `FIXME` here is to show that the test doesn't show the intended behavior

ocaml-lsp-server/test/e2e/__tests__/textDocument-hover.test.ts
```diff
@@ -352,4 +352,134 @@ let x : foo = 1
...
+        moniker: {},
+      },
+    });
+
+    await languageServer.sendNotification("textDocument/didOpen", {
+      textDocument: Types.TextDocumentItem.create(
+        "file:///test.ml",
+        "ocaml",
+        0,
+        // the empty space below is necessary to reproduce the bug
```
tmattio, line 372: It seems odd to me that we need the extra spaces to reproduce this bug. For instance in https://github.com/ocamllabs/vscode-ocaml-platform/issues/111, there doesn't seem to be a need for spaces to trigger the bug. Hopefully, we'll find that these are two manifestations of the same bug, but it could also be two different bugs..

ulugbekna: They should be the same bug; it's just merlin has some checks that the doc cannot be fetched for an identifier if the doc occurs later position-wise, which makes sense. Hence, I need to push down the identifier to match the first doc comment of the file that contains `List` module.

I mentioned about this [here](https://github.com/ocaml/ocaml-lsp/issues/344#issuecomment-787076463).

> In the video I need to push the module def down to reproduce the bug, because the List module documentation in list.mli occurs on lines ~20; hence, we see that the documentation is fetched for things at position p from list.mli instead of the currently open file.

rgrinberg (approved):

PR #438, opened 2021-05-05
Send diagnostics on file save

ulugbekna:
Send diagnostics on a file-save to avoid stale diagnostics.

Example of stale diagnostics: 

We update the interface of another file, but the diagnostics for the current file isn't updated automatically, only when a user makes an edit. Now a user can just save the file to get diagnostics.

----

Some notes:
- Another nicer solution would be to watch edits to all files and update diagnostics to open files that depend on the edited files. This can be implemented by using `ocamldep` I guess. I would be interested in doing this if that's what we want, or it seems like something dune rpc could be useful for. 
- I think this isn't too bad of a solution because the same is done by Scala Metals and perhaps some other language servers.
- Maybe we could avoid creating a new pipeline to get new errors, but we do so for file edits, so why not for a file save.

### DIFF
--- a/ocaml-lsp-server/src/document.ml
+++ b/ocaml-lsp-server/src/document.ml
@@ -165,11 +165,6 @@ let make timer merlin_thread tdoc =
   let+ pipeline = make_pipeline merlin_thread tdoc in
   { tdoc; pipeline; merlin = merlin_thread; timer }
 
-let doc_with_new_merlin_pipeline doc =
-  let open Fiber.O in
-  let+ pipeline = make_pipeline doc.merlin doc.tdoc in
-  { doc with pipeline }
-
 let update_text ?version doc changes =
   let tdoc =
     List.fold_left changes ~init:doc.tdoc ~f:(fun acc change ->
--- a/ocaml-lsp-server/src/document.ml
+++ b/ocaml-lsp-server/src/document.ml
@@ -165,6 +165,11 @@ let make timer merlin_thread tdoc =
   let+ pipeline = make_pipeline merlin_thread tdoc in
   { tdoc; pipeline; merlin = merlin_thread; timer }
 
+let doc_with_new_merlin_pipeline doc =
+  let open Fiber.O in
+  let+ pipeline = make_pipeline doc.merlin doc.tdoc in
+  { doc with pipeline }
+
 let update_text ?version doc changes =
   let tdoc =
     List.fold_left changes ~init:doc.tdoc ~f:(fun acc change ->
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -25,7 +25,7 @@ let initialize_info : InitializeResult.t =
     `TextDocumentSyncOptions
       (TextDocumentSyncOptions.create ~openClose:true
          ~change:TextDocumentSyncKind.Incremental ~willSave:false
-         ~willSaveWaitUntil:false ())
+         ~save:(`Bool true) ~willSaveWaitUntil:false ())
   in
   let codeLensProvider = CodeLensOptions.create ~resolveProvider:false () in
   let completionProvider =
@@ -801,7 +801,20 @@ let on_notification server (notification : Client_notification.t) :
        after receiving this notification *)
     let configuration = Configuration.update state.configuration req in
     Fiber.return { state with configuration }
-  | DidSaveTextDocument _
+  | DidSaveTextDocument { textDocument = { uri }; _ } -> (
+    let open Fiber.O in
+    let state = Server.state server in
+    match Document_store.get_opt state.store uri with
+    | None ->
+      ( Log.log ~section:"on receive DidSaveTextDocument" @@ fun () ->
+        Log.msg "saved document is not in the store" [] );
+      Fiber.return state
+    | Some doc ->
+      (* we need new pipeline to get updated errors *)
+      let* doc = Document.doc_with_new_merlin_pipeline doc in
+      Document_store.put store doc;
+      let+ () = send_diagnostics server doc in
+      state)
   | WillSaveTextDocument _
   | ChangeWorkspaceFolders _
   | Initialized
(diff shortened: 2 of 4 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -802,7 +801,20 @@ let on_notification server (notification : Client_notification.t) :
...
     let configuration = Configuration.update state.configuration req in
     Fiber.return { state with configuration }
-  | DidSaveTextDocument _
+  | DidSaveTextDocument { textDocument = { uri }; _ } -> (
+    let open Fiber.O in
+    let state = Server.state server in
+    match Document_store.get_opt state.store uri with
+    | None ->
+      ( Log.log ~section:"on receive DidSaveTextDocument" @@ fun () ->
+        Log.msg "saved document is not in the store" [] );
```
tmattio, line 810: Looks to me like an error?

ulugbekna: I wasn't confident enough to put an assert false, so logging.

ocaml-lsp-server/src/document.ml
```diff
@@ -165,6 +165,11 @@ let make timer merlin_thread tdoc =
   let+ pipeline = make_pipeline merlin_thread tdoc in
   { tdoc; pipeline; merlin = merlin_thread; timer }
 
+let doc_with_new_merlin_pipeline doc =
```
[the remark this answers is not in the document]
ulugbekna: Thanks for the review. 

I acknowledge that `with_pipeline`s coincide but 
1) `update_pipeline` doesn't sound like it returns a whole new doc, more like mutating things
2) `with_*` is a known idiom, which usually takes a thunk, which is not the case here, so one can figure things

Anyway, I'm happy to put another better name if people want.

ocaml-lsp-server/src/ocaml_lsp_server.ml
tmattio: My bad, I assumed `Log` had an `err` function, which is not the case.

ulugbekna:
No worries, I assumed the same 🙂

tmattio:
> Another nicer solution would be to watch edits to all files and update diagnostics to open files that depend on the edited files.

I would **love** to see this! It would make for a far superior user experience IMO

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -134,21 +134,20 @@ let send_diagnostics ?diagnostics rpc doc =
...
+            List.map errors ~f:(fun (error : Loc.error) ->
+                let loc = Loc.loc_of_report error in
+                let range = Range.of_loc loc in
+                let severity =
+                  match error.source with
+                  | Warning -> DiagnosticSeverity.Warning
+                  | _ -> DiagnosticSeverity.Error
+                in
+                let message =
+                  Loc.print_main Format.str_formatter error;
```
rgrinberg, line 147: This is wrong unfortunately. See #291

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -802,7 +801,20 @@ let on_notification server (notification : Client_notification.t) :
...
+    let open Fiber.O in
+    let state = Server.state server in
+    match Document_store.get_opt state.store uri with
+    | None ->
+      ( Log.log ~section:"on receive DidSaveTextDocument" @@ fun () ->
+        Log.msg "saved document is not in the store" [] );
+      Fiber.return state
+    | Some doc ->
+      (* we need new pipeline to get updated errors *)
+      let* doc = Document.doc_with_new_merlin_pipeline doc in
```
rgrinberg, line 814: This code is too similar to how we handle `TextDocumentDidChange`. In fact it's exactly the same if you take the list of changes to be empty. I suggest to combine the two and remove the `doc_with_new_merlin_pipeline` function altogether.

ulugbekna:
@rgrinberg Thanks for the review. I addressed your comments.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -134,20 +134,21 @@ let send_diagnostics ?diagnostics rpc doc =
...
+                    let loc = Loc.loc_of_report error in
+                    let range = Range.of_loc loc in
+                    let severity =
+                      match error.source with
+                      | Warning -> DiagnosticSeverity.Warning
+                      | _ -> DiagnosticSeverity.Error
+                    in
+                    let message =
+                      Loc.print_main Format.str_formatter error;
+                      String.trim (Format.flush_str_formatter ())
```
rgrinberg: You can also try replacnig Format.str_formatter with Format.asprintf. I don't quite remember what the exact problem was.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -802,7 +802,21 @@ let on_notification server (notification : Client_notification.t) :
...
+    match Document_store.get_opt state.store uri with
+    | None ->
+      ( Log.log ~section:"on receive DidSaveTextDocument" @@ fun () ->
+        Log.msg "saved document is not in the store" [] );
+      Fiber.return state
+    | Some doc ->
+      (* we need [update_text] with no changes to get a new merlin pipeline;
+         otherwise the diagnostics don't get updated *)
+      let* doc = Document.update_text doc [] in
+      Document_store.put store doc;
```
rgrinberg, line 817: What I meant was to move the code in `DidChangeDocument` into a function and just call it in both clauses. This is good as well however.

rgrinberg (approved):

ocaml-lsp-server/src/ocaml_lsp_server.ml
ulugbekna: I just reverted changes to send_diagnostics because it's not really the goal of this PR. I don't quite understand what we want from experimenting with Format fns.

PR #439, opened 2021-05-06
Vendor dune libraries using submodules

rgrinberg:
use the nasty trick of creating the submodules in some random directory
and then copying the things that we need into the vendor directory.

This has the advantage that it makes it easy for us to keep up with
upstream. This will be important to use the rpc client library.


rgrinberg:
Alright, I've vendored dune_rpc as well.

rgrinberg:
ping @tmattio

tmattio:
I find the setup a bit cumbersome. Could we simply add update scripts (the same way dune does) to update the dependencies? This would offer the same convenience to keep track with upstream, while not introducing so much complexity.

rgrinberg:

> I find the setup a bit cumbersome. Could we simply add update scripts
> (the same way dune does) to update the dependencies? This would offer
> the same convenience to keep track with upstream, while not
> introducing so much complexity.

I find it amusing to hear that given that we the dune is so cumbersome
that we're looking to move away from it to submodules (or subtrees) :)

The way I see it, dune's approach has the following fundamental
advantage:

* git clone $foo && cd $foo && make works like a charm

On the other hand, it has the following fundamental disadvantage:

* the git history is polluted with unrelated crap

Given that we have plenty of other dependencies, the easy clone and
build workflow is unrealistic. So it seems like we gain nothing for
committing a bunch of external code into our repo.

Perhaps you find the additional dune files cumbersome? Dune's setup
suffers from the same problem and we must write all dune files for the
vendored dependencies by hand.

tmattio:
> I find it amusing to hear that given that we the dune is so cumbersome that we're looking to move away from it to submodules

I'm surprised, to me, submodules are much heavier than bash scripts. The cost of having to rewrite dune for the vendored libs is a one-time cost, and has to be written for both the bach script and the submodule method.

>  the git history is polluted with unrelated crap

I didn't think of it as a problem, especially when you can ignore files from the diffs with Github by adding them in `.gitattribute`.

> Perhaps you find the additional dune files cumbersome?

It's that the setup is spread on three files for each library: a submodule, a dune file for the library and the rules in the root file to copy the files. But most of all, I find the CLI for `git submodules` really unintuitive to use.

All in all, the PR is still better than the current setup. I think bash scripts would be simpler, but if you're going to change the method you use in dune, I'm in favor of using the same one so that we follow some kind of global convention.

rgrinberg:
The pain points with the bash script based setup is that it adds friction to developing the vendored dependencies. It's a very specific to our repo and dune, but it makes a common workflow quite painful. If I'm working on a change across multiple repos, I have two options:

* Work with the vendored copy and then extract patches to apply upstream.
* Work upstream and then run the bash script.

The former makes it impossible to work with the git history of upstream. So you can forget about pulling pr's locally, rebasing to create a nice history for upstream, etc.

The latter makes it difficult to have a proper edit/run cycle. There's now an additional pull step that's cumbersome. At least I don't of a better way to work in this setup.

Now, this issue is usually isn't a problem if the deps are fairly static. But for us, merlin constantly changes, and from now on, so will dune. That's why I tried to avoid submodules for fiber/stdune initially, but finally gave up as the needs grew with dune-rpc.

I also find the submodules UI to be utter crap, but have yet find a good alternative. I've experimented with subtrees, but they seem to be only useful in special cases. I made a note to myself to try out [subrepo](https://github.com/ingydotnet/git-subrepo), but haven't gotten around to it yet. Perhaps if you are interested in this, you can experiment and see if it fits us our use case?

One thing that we can consider is to switch to bash scripts for static deps like uutf for example. It does not seem like submodules offer much benefits here. But perhaps it's worth the inconvenience to keep everything consistent?

tmattio:
> I made a note to myself to try out subrepo, but haven't gotten around to it yet. Perhaps if you are interested in this, you can experiment and see if it fits us our use case?

It looks really nice! I'm not sure adding a system dependency (other than git) is worth the benefits, but I'll remember to check it out when I have the occasion

tmattio:
I meant that the submodule method requirement seems temporary. Once the dust settles, we might find that bash scripts will become convenient once again.

> I don’t want to inflict the kind of pain ocamlformat does on users.

Hopefully, tools installation outside of the opam switch will come in Opam sooner than later 🤞

PR #440, opened 2021-05-07
[scheduler] switch to polling timer

rgrinberg:
This is going to be inline with the implementation inside dune's own
timer.

@tmattio we should have went with this from the start after all.


tmattio:
Is there a discussion I can follow on why polling is better? Intuitively, I would have thought that the barrier implementation was more optimized.

fiber-unix/src/scheduler.ml
```diff
@@ -189,60 +185,19 @@ let time_loop t =
...
     add_events t to_run;
-    Option.iter !earliest_next ~f:(fun s ->
-        with_mutex t.earliest_next_mutex ~f:(fun () ->
-            t.earliest_next <- Some s);
-        match Barrier.signal t.earliest_next_barrier with
-        | Ok () -> ()
-        | Error `Closed -> assert false);
-    with_mutex t.timers_available_mutex ~f:(fun () ->
-        Condition.wait t.timers_available t.timers_available_mutex);
+    Unix.sleepf t.timer_resolution;
```
tmattio, line 196: Do I understand correctly that we delegate the scheduling to the OS with the sleep here?

rgrinberg:

> Is there a discussion I can follow on why polling is better?
> Intuitively, I would have thought that the barrier implementation was
> more optimized.

No public discussions that I'm aware, but the reasoning is as follows:

* The complex implementation is mostly useful when there's long periods
  where nothing is scheduled. We don't actually have something like that
  in lsp and schedule things regularly.

* After a lot of work, our implementation is still wrong. There's a race
  between select and signal and it's a pain to fix.

* I'd like to upstream this code to dune, but their scheduler is already
  complex and I don't want to make things even more complex without
  really good justification.

rgrinberg:

> Do I understand correctly that we delegate the scheduling to the OS
> with the sleep here?

We poll with a regular interval to check which timers are expired.

PR #443, opened 2021-05-13
Dune RPC Integration

rgrinberg:
Still WIP, but gets us pretty far.

The main complication is handling the mismatch between how LSP and dune handle
notifications. LSP allows us updating notifications for a document in batch,
which means that we need to know the last notifications unrelated to dune, so
that we only update the dune section.

### DIFF
--- a/lsp/src/uri0.ml
+++ b/lsp/src/uri0.ml
@@ -24,6 +24,8 @@ let yojson_of_t t = `String (to_string t)
 
 let equal = ( = )
 
+let compare (x : t) (y : t) = Stdlib.compare x y
+
 let hash = Hashtbl.hash
 
 let to_dyn { scheme; authority; path } =
--- a/ocaml-lsp-server/src/diagnostics.ml
+++ b/ocaml-lsp-server/src/diagnostics.ml
@@ -1,0 +1,99 @@
+open Import
+
+module Uri = struct
+  include Uri
+
+  let compare x y = Ordering.of_int (Uri.compare x y)
+end
+
+module Uri_c = Comparable.Make (Uri)
+module Uri_set = Uri_c.Set
+
+type dune_status =
+  | Inactive
+  | Connected
+  | Disconnected
+
+type t =
+  { dune : (Drpc.Diagnostic.Id.t, Uri.t * Diagnostic.t) Table.t
+  ; mutable dune_status : dune_status
+  ; workspace_root : Uri.t Lazy.t
+  ; merlin : (Uri.t, Diagnostic.t list) Table.t
+  ; send : PublishDiagnosticsParams.t list -> unit Fiber.t
+  ; mutable dirty_uris : Uri_set.t
+  }
+
+let workspace_root t = Lazy.force t.workspace_root
+
+module Id = struct
+  include Drpc.Diagnostic.Id
+
+  let equal x y = compare x y = 0
+
+  let to_dyn = Dyn.Encoder.opaque
+end
+
+let create send ~workspace_root =
+  { dune = Table.create (module Id) 32
+  ; merlin = Table.create (module Uri) 32
+  ; dirty_uris = Uri_set.empty
+  ; dune_status = Inactive
+  ; send
+  ; workspace_root
+  }
+
+let update_dune_status t status =
+  Table.filteri_inplace t.dune ~f:(fun ~key:_ ~data:(uri, _) ->
+      t.dirty_uris <- Uri_set.add t.dirty_uris uri;
+      false);
+  t.dirty_uris <- Uri_set.add t.dirty_uris (Lazy.force t.workspace_root);
+  t.dune_status <- status
+
+let dune_status_diagnostic t =
+  match t.dune_status with
+  | Disconnected ->
+    Some
+      (Diagnostic.create ~severity:Information ~source:"dune"
+         ~range:Range.first_line
+         ~message:
+           "Dune diagnostic status may be stale. Please run dune in watch mode\n\
+            to see up to date diagnostics"
+         ())
+  | _ -> None
+
+let send t =
+  let pending = Table.create (module Uri) 32 in
+  Uri_set.iter t.dirty_uris ~f:(fun uri ->
+      let diagnostics = Table.Multi.find t.merlin uri in
+      Table.set pending uri diagnostics);
+  dune_status_diagnostic t
+  |> Option.iter ~f:(fun d ->
+         Table.Multi.cons pending (Lazy.force t.workspace_root) d);
+  Table.iter t.dune ~f:(fun (uri, diagnostic) ->
+      if Uri_set.mem t.dirty_uris uri then
+        Table.Multi.cons pending uri diagnostic);
+  t.dirty_uris <- Uri_set.empty;
+  Table.foldi pending ~init:[] ~f:(fun uri diagnostics acc ->
+      PublishDiagnosticsParams.create ~uri ~diagnostics () :: acc)
+  |> t.send
+
+let set t what =
+  let uri =
+    match what with
+    | `Dune (_, uri, _) -> uri
+    | `Merlin (uri, _) -> uri
+  in
+  t.dirty_uris <- Uri_set.add t.dirty_uris uri;
+  match what with
+  | `Dune (id, uri, diagnostics) -> Table.set t.dune id (uri, diagnostics)
+  | `Merlin (uri, diagnostics) -> Table.set t.merlin uri diagnostics
+
+let remove t = function
+  | `Dune id ->
+    Table.find t.dune id
+    |> Option.iter ~f:(fun (uri, _) ->
+           Table.remove t.dune id;
+           t.dirty_uris <- Uri_set.add t.dirty_uris uri)
+  | `Merlin uri ->
+    t.dirty_uris <- Uri_set.add t.dirty_uris uri;
+    Table.remove t.merlin uri
--- a/ocaml-lsp-server/src/diagnostics.mli
+++ b/ocaml-lsp-server/src/diagnostics.mli
@@ -1,0 +1,28 @@
+open Import
+
+type t
+
+val create :
+     (PublishDiagnosticsParams.t list -> unit Fiber.t)
+  -> workspace_root:Uri.t Lazy.t
+  -> t
+
+val send : t -> unit Fiber.t
+
+type dune_status =
+  | Inactive
+  | Connected
+  | Disconnected
+
+val update_dune_status : t -> dune_status -> unit
+
+val workspace_root : t -> Uri.t
+
+val set :
+     t
+  -> [ `Dune of Drpc.Diagnostic.Id.t * Uri.t * Diagnostic.t
+     | `Merlin of Uri.t * Diagnostic.t list
+     ]
+  -> unit
+
+val remove : t -> [ `Dune of Drpc.Diagnostic.Id.t | `Merlin of Uri.t ] -> unit
--- a/ocaml-lsp-server/src/dune.ml
+++ b/ocaml-lsp-server/src/dune.ml
@@ -1,43 +1,259 @@
 open! Import
 
-type state =
+let dev_null () =
+  (* TODO stdune should provide an api to simplify this *)
+  Unix.openfile
+    (if Sys.win32 then
+      "nul"
+    else
+      "/dev/null")
+    [ Unix.O_RDWR ] 0o666
+
+module Chan : sig
+  type t
+
+  val create : string -> t Fiber.t
+
+  val write : t -> Csexp.t list option -> unit Fiber.t
+
+  val read : t -> Csexp.t option Fiber.t
+
+  val pid : t -> Pid.t
+
+  val run : t -> unit Fiber.t
+end = struct
+  open Fiber.O
+
+  type t =
+    { in_thread : Scheduler.thread
+    ; out_thread : Scheduler.thread
+    ; out_chan : out_channel
+    ; in_chan : in_channel
+    ; pid : Pid.t
+    }
+
+  let pid t = t.pid
+
+  let write t sexp =
+    let+ res =
+      Scheduler.async_exn t.out_thread (fun () ->
+          match sexp with
+          | None ->
+            close_in_noerr t.in_chan;
+            close_out_noerr t.out_chan
+          | Some sexps ->
+            List.iter sexps ~f:(Csexp.to_channel t.out_chan);
+            flush t.out_chan)
+      |> Scheduler.await_no_cancel
+    in
+    match res with
+    | Error e -> Exn_with_backtrace.reraise e
+    | Ok s -> s
+
+  let read t =
+    let+ res =
+      Scheduler.async_exn t.in_thread (fun () ->
+          match Csexp.input_opt t.in_chan with
+          | Error _
+          | Ok None ->
+            close_in_noerr t.in_chan;
+            None
+          | Ok (Some s) -> Some s)
+      |> Scheduler.await_no_cancel
+    in
+    match res with
+    | Error e -> Exn_with_backtrace.reraise e
+    | Ok s -> s
+
+  let create dune =
+    let args = Array.of_list [ dune; "rpc"; "init"; "--wait" ] in
+    let pid, stdout, stdin =
+      let stdin_i, stdin_o = Unix.pipe () in
+      let stdout_i, stdout_o = Unix.pipe () in
+      let pid =
+        Unix.create_process dune args stdin_i stdout_o Unix.stderr |> Pid.of_int
+      in
+      (pid, stdout_i, stdin_o)
+    in
+    let* in_thread = Scheduler.create_thread () in
+    let+ out_thread = Scheduler.create_thread () in
+    let in_chan = Unix.in_channel_of_descr stdout in
+    let out_chan = Unix.out_channel_of_descr stdin in
+    { pid; in_chan; out_chan; in_thread; out_thread }
+
+  let run t =
+    let+ (_ : Unix.process_status) = Scheduler.wait_for_process t.pid in
+    ()
+end
+
+module Client =
+  Drpc.Client
+    (struct
+      include Fiber
+
+      let parallel_iter t ~f =
+        let stream = Fiber.Stream.In.create t in
+        Fiber.Stream.In.parallel_iter stream ~f
+    end)
+    (Chan)
+
+type run =
   | Binary_not_found
   | Out_of_date
-  | Running
 
-type t = { state : state Fiber.Ivar.t }
+type state =
+  | Waiting_for_init of { diagnostics : Diagnostics.t }
+  | Active of
+      { diagnostics : Diagnostics.t
+      ; finish : unit Fiber.Ivar.t
+      ; chan : Chan.t
+      }
+  | Closed
 
-let create () = { state = Fiber.Ivar.create () }
+type t = state ref
 
-let state t : state Fiber.t =
-  let open Fiber.O in
-  let* state = Fiber.Ivar.peek t.state in
-  match state with
-  | Some s -> Fiber.return s
-  | None -> (
+let stop (t : t) =
+  match !t with
+  | Closed -> Fiber.return ()
+  | Waiting_for_init _ ->
+    t := Closed;
+    Fiber.return ()
+  | Active { finish; chan; diagnostics = _ } ->
+    t := Closed;
+    let pid = Chan.pid chan in
+    Unix.kill (Pid.to_int pid) Sys.sigstop;
+    Fiber.Ivar.fill finish ()
+
+let create diagnostics = ref (Waiting_for_init { diagnostics })
+
+let lsp_of_dune uri dune =
+  let module D = Drpc.Diagnostic in
+  let range_of_loc loc =
+    let loc =
+      let loc_start = Drpc.Loc.start loc in
+      let loc_end = Drpc.Loc.stop loc in
+      { Loc.loc_start; loc_end; loc_ghost = false }
+    in
+    Range.of_loc loc
+  in
+  let range =
+    match D.loc dune with
+    | None -> Range.first_line
+    | Some loc -> range_of_loc loc
+  in
+  let severity =
+    D.severity dune
+    |> Option.map ~f:(function
+         | D.Error -> DiagnosticSeverity.Error
+         | Warning -> DiagnosticSeverity.Warning)
+  in
+  let make_message message = Format.asprintf "%a@." Pp.to_fmt message in
+  let relatedInformation =
+    match D.related dune with
+    | [] -> None
+    | related ->
+      Some
+        (List.map related ~f:(fun related ->
+             let message = make_message (D.Related.message related) in
+             let location =
+               let range = range_of_loc (D.Related.loc related) in
+               Location.create ~uri ~range
+             in
+             DiagnosticRelatedInformation.create ~location ~message))
+  in
+  let message = make_message (D.message dune) in
+  Diagnostic.create ?relatedInformation ~range ?severity ~source:"dune" ~message
+    ()
+
+let run_rpc (t : t) bin =
+  match !t with
+  | Closed -> Code_error.raise "dune already closed" []
+  | Active _ -> Code_error.raise "dune alrady running" []
+  | Waiting_for_init { diagnostics } ->
+    Diagnostics.update_dune_status diagnostics Disconnected;
+    let open Fiber.O in
+    let finish = Fiber.Ivar.create () in
+    let* chan = Chan.create bin in
+    t := Active { diagnostics; finish; chan };
+    let+ () =
+      Fiber.parallel_iter
+        ~f:(fun f -> f ())
+        [ (fun () -> Diagnostics.send diagnostics)
+        ; (fun () ->
+            let* () = Chan.run chan in
+            (* TODO ideally, we should notify the users that the diagnostics are
+               stale until they run dune again *)
+            Fiber.Ivar.fill finish ())
+        ; (fun () ->
+            let init =
+              Drpc.Initialize.create ~id:(Drpc.Id.make (Atom "ocamllsp"))
+            in
+            let handler =
+              let diagnostic evs =
+                List.iter evs ~f:(fun (ev : Drpc.Diagnostic.Event.t) ->
+                    let id =
+                      Drpc.Diagnostic.id
+                        (match ev with
+                        | Add x -> x
+                        | Remove x -> x)
+                    in
+                    match ev with
+                    | Remove _ -> Diagnostics.remove diagnostics (`Dune id)
+                    | Add d ->
+                      let uri : Uri.t =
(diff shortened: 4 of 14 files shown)


fiber-test/fiber_test.ml
```diff
@@ -45,7 +45,7 @@ let test ?(expect_never = false) to_dyn f =
   let f =
     let on_error exn =
       Format.eprintf "%a@." Exn_with_backtrace.pp_uncaught exn;
-      Fiber.return ()
+      Exn_with_backtrace.reraise exn
```
tmattio, line 48: I don't see any this exception handled anywhere, am I missing something?

lsp/src/uri0.ml
```diff
@@ -24,6 +24,8 @@ let yojson_of_t t = `String (to_string t)
 
 let equal = ( = )
 
+let compare (x : t) (y : t) = compare x y
```
tmattio, line 27: Why do you need to define this?

ocaml-lsp-server/src/diagnostics.ml
```diff
@@ -0,0 +1,64 @@
...
+
+  let compare x y = Ordering.of_int (Uri.compare x y)
+end
+
+module Uri_c = Comparable.Make (Uri)
+module Uri_set = Uri_c.Set
+
+type t =
+  { dune : (int, Uri.t * Diagnostic.t) Table.t
+  ; merlin : (Uri.t, Diagnostic.t list) Table.t
```
tmattio, line 14: Shouldn't dune provide the merlin diagnostics as well?

ocaml-lsp-server/src/state.ml
```diff
@@ -13,5 +13,6 @@ type t =
   ; configuration : Configuration.t
   ; trace : TraceValue.t
   ; ocamlformat : Fmt.t
+  ; diagnostics : Diagnostics.t
```
tmattio, line 16: Why is `Diagnostic.t` in both `State.t` and `Dune.t`?

ocaml-lsp-server/src/diagnostics.mli
```diff
@@ -0,0 +1,16 @@
...
+type t
+
+val create : (PublishDiagnosticsParams.t -> unit Fiber.t) -> t
+
+val send : t -> unit Fiber.t
+
+val set :
+     t
+  -> [ `Dune of int * Uri.t * Diagnostic.t
+     | `Merlin of Uri.t * Diagnostic.t list
```
tmattio, line 12: Should you know where the diagnostic come from? Why not `val set : t -> (Ur.t * Diagnostic.t) list -> unit`

Also, maybe an immutable API would be safer: 

```
val set : t -> (Ur.t * Diagnostic.t) list -> t
```

I'm not sure if this would become a bottleneck performance-wise.

ocaml-lsp-server/src/diagnostics.mli
```diff
@@ -0,0 +1,16 @@
...
+     ]
+  -> unit
+
+val remove : t -> [ `All_dune | `Dune of int | `Merlin of Uri.t ] -> unit
```
tmattio, line 16: Why is the variant `Dune` taking an `int`? You're passing a `Uri.t` in `set` as well, so this is a bit confusing reading the interface. Maybe some docstring if this is justified?

ocaml-lsp-server/src/dune.ml
```diff
@@ -1,46 +1,158 @@
...
+        (* TODO inform client instead *)
+        failwith "Unexpected termination from dune")
+    (fun () ->
+      Client.connect_persistent chan
+        ~on_connected:(fun () t ->
+          let sub = Drpc.Subscribe.Diagnostics in
+          Client.notification t Drpc.Notification.subscribe sub)
+        ~on_connect:(fun () ->
+          let init =
+            Drpc.Initialize.create ~id:(Drpc.Id.make (Atom "ocamllsp"))
```
tmattio, line 117: Should the ID of the session be dynamic? In case the connection is lost and we have to re-create one, I would expect to have a new session ID.

ocaml-lsp-server/src/dune.ml
```diff
@@ -1,46 +1,158 @@
...
+    D.severity dune
+    |> Option.map ~f:(function
+         | D.Error -> DiagnosticSeverity.Error
+         | Warning -> DiagnosticSeverity.Warning)
+  in
+  let message = Format.asprintf "%a@." Pp.to_fmt (D.message dune) in
+  Diagnostic.create ~range ?severity ~source:"dune" ~message ()
 
+let run_rpc t bin =
+  let stdin_i, stdin_o = Unix.pipe () in
```
tmattio, line 92: I don't see these being closed. Should `Chan` be responsible for closing the fd properly? If so, it could also be responsible to create these and provide a function `val fd: t -> (Unix.fd * Unix.fd)`

tmattio (commented):
This is great, congrats for tackling this @rgrinberg, that's the culmination of **A LOT** of work and I'm really excited to be able to add new features that depend on Dune RPC soon! 🎉 

I did an initial review pass, without getting too much into the implementation. I'll have to make another pass for the implementation itself.

fiber-test/fiber_test.ml
rgrinberg: it's just to make the test a little more resilient to fiber deadlocks. We can do one of two things:

* Wait for all fibers to complete and then print errors
* Print errors as they come

We're doing the latter, but without reraising the exceptions. That's a little fragile, but worked because we didn't rely on these exceptions in the tests.

lsp/src/uri0.ml
rgrinberg: Many functor signature require a comparison function. It's usually a nice to have for that reason.

ocaml-lsp-server/src/diagnostics.ml
rgrinberg: How can dune provide the merlin diagnostics? Dune doesn't know anything about merlin.

We need to separate the fields because the diagnostics are updated differently for merlin and dune:

* For merlin, the diagnostics are always calculated for every document. Therefore, we always update diagnostics per document in bulk.
* For dune, the diagnostics can come in any order with add/remove events. Therefore, we keep unique id's for diagnostics to support efficient remove.

I will probably implement diagnostic id's on the client side to make this whole thing more obvious.

ocaml-lsp-server/src/diagnostics.mli
rgrinberg: > Why not val set : t -> (Ur.t * Diagnostic.t) list -> unit

Because dune does not send us diagnostics per document. It sends add/remove events in arbitrary order. This is why I need the hash table keyed by ID to maintain the current set of diagnostics efficiently.

> I'm not sure if this would become a bottleneck performance-wise.

I'm not worried about the performance but the about boilerplate overhead. This can be experimented with however.

ocaml-lsp-server/src/dune.ml
rgrinberg: The session id doesn't have to be unique, it's just to have something "nice" to show in `$ dune rpc status`.

ocaml-lsp-server/src/state.ml
rgrinberg: Good question. We do make sure to only mutate` Diagnostics.t` in the main thread, so that gives a lot of safety right away.
The only issue that could come up is with `Diagnostics.send` and other functions mutating the state. However, I still can't see anything that might go wrong since `send` doesn't wait on a fiber before mutating anything.

ocaml-lsp-server/src/dune.ml
rgrinberg: We don't close the read end because we never stop reading I believe. The write end is closed when we close the channel (see `Chan.write`).

> If so, it could also be responsible to create these and provide a function val fd: t -> (Unix.fd * Unix.fd)

The thing is, it's only responsible 1 read and 1 write end of the pipe. So it's error prone that it needs to return fd's that will have to be closed by the spawned process. That's why I opted to just create them upfront.

lsp/src/uri0.ml
tmattio: I was asking because compare seems to already be in scope, but maybe it's coming from an `open` statement? Hard to know with the diff

ocaml-lsp-server/src/diagnostics.ml
tmattio: > How can dune provide the merlin diagnostics? Dune doesn't know anything about merlin.

Dune at least generates the merlin files. For some reason, I assumed that it would also provide merlin diagnostics through an RPC request.

> For merlin, the diagnostics are always calculated for every document.

Are Dune events always global? If so why do we need the `Uri.t` for Dune as well? (trying to understand a bit better, I clearly have some gaps in my understanding of how diagnostics are generated)

ocaml-lsp-server/src/diagnostics.mli
tmattio: You answered that above.

ocaml-lsp-server/src/dune.ml
tmattio: > We don't close the read end because we never stop reading I believe.

I remember a similar statement in another PR. That's fine if we assume `dune init` will never crash, but it's a bold assumption. Granted dune does not crash often, but it happens, and if it does at the moment, I think the only workaround will be to restart the language server, right? Just playing devil's advocate, I don't know how much effort we want to put in being fault-tolerant against dune.

fiber-test/fiber_test.ml
tmattio: I'm confused: you seem to say that `Exn_with_backtrace.reraise` prints the error and does not reraise, but the name indicates that the exception *is* reraised

rgrinberg: The previous code printed the error but did not reraise. That was a mistake. The new code prints and reraises as expected.

lsp/src/uri0.ml
rgrinberg: I'll use `Poly.compare` to be more clear.

ocaml-lsp-server/src/diagnostics.ml
rgrinberg: > Are Dune events always global? If so why do we need the Uri.t for Dune as well? (trying to understand a bit better, I clearly have some gaps in my understanding of how diagnostics are generated)

LSP requires every diagnostic to be associated to a file. That's why we need an associated Uri. Dune events are global but we can always find a suitable location that somewhat corresponds to the "source" of the error. This is what we're going to use to compute the uri.

ocaml-lsp-server/src/dune.ml
rgrinberg: Yes, that's a  good point. I've changed the code to work as follows:

* dune rpc init crashes
* we are unable to read from the read end of the pipe
* we now close the read of the pipe as a result

ocaml-lsp-server/src/dune.ml
```diff
@@ -28,6 +28,7 @@ module Chan = struct
           match Csexp.input_opt t.in_chan with
           | Error _
           | Ok None ->
+            Unix.close (Unix.descr_of_in_channel t.in_chan);
```
tmattio, line 31: Ideally, we could log this and send a notification to the client

ocaml-lsp-server/src/diagnostics.ml
```diff
@@ -0,0 +1,64 @@
+open Import
+
+module Uri = struct
+  include Uri
+
+  let compare x y = Ordering.of_int (Uri.compare x y)
```
tmattio, line 6: Where does `uri` come from? Can we upstream this implementation of `compare`?

ocaml-lsp-server/src/diagnostics.ml
```diff
@@ -0,0 +1,64 @@
...
+end
+
+module Uri_c = Comparable.Make (Uri)
+module Uri_set = Uri_c.Set
+
+type t =
+  { dune : (int, Uri.t * Diagnostic.t) Table.t
+  ; merlin : (Uri.t, Diagnostic.t list) Table.t
+  ; send : PublishDiagnosticsParams.t -> unit Fiber.t
```
tmattio, line 15: Why is this dynamic? I would expect the method of sending diagnostics to be general to all the diagnostics

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -847,21 +833,32 @@ let start () =
...
-      ; merlin
-      ; ocamlformat
-      ; scheduler
-      ; configuration
-      ; detached
-      ; dune
-      ; trace = `Off
-      }
+    Fdecl.set server
+      (Server.make handler stream
```
tmattio, line 849: Why use a forward declaration when you're setting the value in the same function? `server` could also be passed to `Diagnostics.create` to avoid having to define a `send` outside of `Diagnostics` as commented above.

ocaml-lsp-server/src/diagnostics.ml
```diff
@@ -0,0 +1,64 @@
...
+  ; mutable dirty_uris : Uri_set.t
```
tmattio, line 16: Not sure if that will be accessed concurrently, but we should probably lock the access to `dirty_uris` in `Diagnostics`, to be safe.

ocaml-lsp-server/src/dune.ml
```diff
@@ -1,46 +1,159 @@
...
+    let in_chan = Unix.in_channel_of_descr in_fd in
+    let out_chan = Unix.out_channel_of_descr out_fd in
+    { in_thread; out_thread; in_chan; out_chan }
+end
+
+module Drpc = Dune_rpc.V1
+
+module Client =
+  Drpc.Client
+    (struct
```
tmattio, line 52: Oh, `Drpc.Client` is functorized? I thought it depended on `stdune` already, why not use Dune's `Fiber`?

ocaml-lsp-server/src/dune.ml
```diff
@@ -1,46 +1,159 @@
...
+      Range.of_loc loc
+  in
+  let severity =
+    D.severity dune
+    |> Option.map ~f:(function
+         | D.Error -> DiagnosticSeverity.Error
+         | Warning -> DiagnosticSeverity.Warning)
+  in
+  let message = Format.asprintf "%a@." Pp.to_fmt (D.message dune) in
+  Diagnostic.create ~range ?severity ~source:"dune" ~message ()
```
tmattio, line 90: This function is surprisingly small considering that it's the bridge between Dune RPC and LSP! This is great 🎉

ocaml-lsp-server/src/dune.ml
```diff
@@ -1,46 +1,159 @@
...
-        match status with
-        (* TODO actually turn on *)
-        | Unix.WEXITED 0 -> Running
-        | _ -> Out_of_date
-      in
-      let+ () = Fiber.Ivar.fill t.state state in
-      state)
+      let* res = Scheduler.wait_for_process t.scheduler pid in
+      match res with
+      | Unix.WEXITED 0 -> Fiber.return ()
```
tmattio, line 107: This should be considered an error as well, right? We should inform the client here also

ocaml-lsp-server/src/dune.ml
```diff
@@ -1,46 +1,159 @@
...
+    let pid =
+      let args = Array.of_list [ bin; "rpc"; "--help=plain" ] in
+      Unix.create_process bin args stdin stdout stderr |> Stdune.Pid.of_int
+    in
+    let* status = Scheduler.wait_for_process t.scheduler pid in
+    match status with
+    | Unix.WEXITED 0 ->
+      let+ () = run_rpc t bin in
+      Ok ()
+    | _ -> Fiber.return (Error Out_of_date))
```
tmattio, line 159: If dune is not found in Path, the error will be a bit cryptic for users here

ocaml-lsp-server/src/state.ml
tmattio: I still find it a bit confusing that we have multiple instances of `diagnostics`. It make it hard to make a mental model of the diagnostics workflow:

- Where are the diagnostics sent from?
- Where are Dune diagnostics and merlin diagnostics stored, in which instance?

Could we pass a function to add/remove diagnostics to the modules that need access to it instead? It makes sense to have a global instance of Diagnostics in the state IMO, so that the diagnostics IO can be centralized.

ocaml-lsp-server/src/dune.ml
rgrinberg: Yeah, that makes sense. Note that this case always corresponds to a bug in `dune rpc init`. It should never crash or terminate on its own.

ocaml-lsp-server/src/diagnostics.ml
rgrinberg: It's needed to avoid circular dependencies. I'll see if I can do it in another way though.

ocaml-lsp-server/src/diagnostics.ml
rgrinberg: We cannot because `Ordering` is from stdune and we are trying to avoid using stdune in lsp.

ocaml-lsp-server/src/dune.ml
rgrinberg: Two reasons:

* `Fiber` isn't public yet.
* We want people to write clients with Lwt, Async, and other concurrency libraries they might think of.

ocaml-lsp-server/src/diagnostics.ml
rgrinberg: I don't think using mutexes defensively is a good idea. They can introduce their own problems like additional deadlocks. After all, we are using fiber to (mostly) avoid the issues of multi threaded programming. We can use `Fiber.Mutex.t` when we want to do something atomically across more than 1 fiber, but it's clear we don't need it here.

ocaml-lsp-server/src/dune.ml
rgrinberg: Thanks! Hopefully this paves the way for adding dune rpc features to vscode as well.

ocaml-lsp-server/src/dune.ml
rgrinberg: Yes, that's why this function checks if the binary exists first :)

ocaml-lsp-server/src/dune.ml
rgrinberg: Yes indeed. Again this can only occur under exceptional circumstances.

ocaml-lsp-server/src/state.ml
rgrinberg: > I still find it a bit confusing that we have multiple instances of diagnostics. It make it hard to make a mental model of the diagnostics workflow:

We do not have multiple instances of `Diagnostics.t` - `Diagnostics.create` is only called once. We have two references to it in `State.t` and `Dune.t`. That's mostly because `Dune.t` cannot refer to `State.t` due to cycles.

> Could we pass a function to add/remove diagnostics to the modules that need access to it instead? It makes sense to have a global instance of Diagnostics in the state IMO, so that the diagnostics IO can be centralized.

Maybe you could post a diff to show what you mean? I don't quite see where this is going. In my mind, the current code already achieves  the goal of centralizing the sending of diagnostics: there's a `Diagnostics.t` value and it is responsible for sending diagnostics

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: It's done to break the following cycle:

* `dune` depends on `diagonstics`
* `server` depends on `dune`
* `diagnostics` depends on `server`

Perhaps it's possible to avoid this cycle with `lazy` and `let .. and`, but it's more work because some of the bindings require fibers.

ocaml-lsp-server/src/diagnostics.ml
rgrinberg: I tried to, but I really don't see a good alternative. It's possible to define State.t and Dune.t in a mutually recursive manner, but that means leaking the internal fields of Dune.t. That seems much uglier than passing an ugly send callback.

tmattio: Alright! I'm not familiar enough with the whole architecture to have suggestions here, so I trust your judgement 🙂

ocaml-lsp-server/src/dune.ml
tmattio: I see you left a TODO, I'll resolve the conversation

ocaml-lsp-server/src/ocaml_lsp_server.ml
tmattio: Understand, thanks!

ocaml-lsp-server/src/diagnostics.ml
rgrinberg: I don't necessarily trust my own judgement, but it's just that I'm not sure what to do to improve this code without introducing other issues. So an improvement is definitely is possible, but it will require additional thinking and unlikely to come in this PR unless some concrete suggestions are made.

rgrinberg:
@tmattio I think this is a mergeable state. Please give this another review when you can.

There's still some issues, but they require fixes on the dune side rather than in this PR.

tmattio:
>  Please give this another review when you can.

I'll have a last review pass by Friday at the latest 🙂

rgrinberg:
I just realized that this integration is a bit too simple for workspaces with more than one folder. In such setups, we need an instance of dune per folder. For now this is an acceptable limitation but it will need to be lifted before the release.

ocaml-lsp-server/src/diagnostics.ml
```diff
@@ -0,0 +1,100 @@
...
+  |> Option.iter ~f:(fun d ->
+         Table.add_exn pending (Lazy.force t.workspace_root) [ d ]);
+  Uri_set.iter t.dirty_uris ~f:(fun uri ->
+      let diagnostics = Table.Multi.find t.merlin uri in
+      Table.set pending uri diagnostics);
+  Table.iter t.dune ~f:(fun (uri, diagnostic) ->
+      if Uri_set.mem t.dirty_uris uri then
+        Table.Multi.cons pending uri diagnostic);
+  t.dirty_uris <- Uri_set.empty;
+  Table.foldi pending ~init:[] ~f:(fun uri diagnostics acc ->
```
tmattio, line 73: It's strange that `foldi` does not pass the index to `~f`

ocaml-lsp-server/src/dune.ml
```diff
@@ -1,43 +1,251 @@
...
+    (Chan)
+
+type run =
   | Binary_not_found
   | Out_of_date
-  | Running
 
-type t = { state : state Fiber.Ivar.t }
+type state =
+  | Waiting_for_init of { diagnostics : Diagnostics.t }
```
tmattio, line 97: How can you have diagnostics if you're still waiting for init? If this is from a previous connection, wouldn't Closed need that as well?

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -847,20 +833,40 @@ let start () =
...
+    Diagnostics.create ~workspace_root (fun diagnostics ->
+        let server = Fdecl.get server in
+        let state = Server.state server in
+        task_if_running state ~f:(fun () ->
+            let batch = Server.Batch.create server in
+            List.iter diagnostics ~f:(fun d ->
+                Server.Batch.notification batch (PublishDiagnostics d));
+            Server.Batch.submit batch))
+  in
+  let dune = Dune.create diagnostics in
```
tmattio, line 853: If I understand, these are general diagnostics, not necessarily dune-generated diagnostics, right? Is there a reason for passing them to `Dune`?

ocaml-lsp-server/test/e2e/__tests__/textDocument-diagnostics.ts
```diff
@@ -53,7 +53,6 @@ describe("textDocument/diagnostics", () => {
               },
             ],
             "uri": "file:///test.ml",
-            "version": 0,
```
tmattio, line 56: Why removing the `version`? Aren't you reverting part of #445?

ocaml-lsp-server/src/dune.ml
```diff
@@ -1,43 +1,251 @@
...
-        match status with
-        (* TODO actually turn on *)
-        | Unix.WEXITED 0 -> Running
-        | _ -> Out_of_date
+  match Bin.which "dune" with
+  | None -> Fiber.return (Error Binary_not_found)
+  | Some bin -> (
+    let bin = Fpath.to_string bin in
+    (* TODO this is wrong on windows *)
+    let dev_null () = Unix.openfile "/dev/null" [ Unix.O_RDWR ] 0o666 in
```
tmattio, line 231: I think this should work:
```suggestion
    let dev_null () = 
      let f = if Sys.win32 then "nul" else "/dev/null" in
      Unix.openfile f [ Unix.O_RDWR ] 0o666
    in
```

tmattio (approved):
I left a few questions, but the PR looks good! Thanks again for the amazing work 🎉

ocaml-lsp-server/src/diagnostics.ml
rgrinberg: The index in `Table.foldi` is the key in the hash table. In this case that's passed as `uri`.

ocaml-lsp-server/src/dune.ml
rgrinberg: It's perhaps a bit confusing but `Diagnostics.t` is the thing that updates diagnostics on the client in a way that doesn't that does not remove any previous diagnostics from other sources (like merlin). When waiting for initialization, we need a reference to this value because it will be used when switching to the active state. In the closed state, we will no longer be able to transition to Acitive so the reference to diagnostics isn't needed anymore.

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: `Diagnostics.t` is the api for sending ocamllsp's diagnostics. Seems like dune needs to use it to send its diagnostics.

ocaml-lsp-server/test/e2e/__tests__/textDocument-diagnostics.ts
rgrinberg: Yes, I am. Unfortunately, I don't see how it can work with diagnostics sourced from dune. The issue is that lsp requires you to send all diagnostics for a document in a single notification, but only the notifications from merlin depend on the source. The notifications from dune are "as is" and are computed based on the source on the file system and hence applying a version to them doesn't make sense. Lsp doesn't permit versioning only a subset of diagnostics, so the entire thing has to go.

ocaml-lsp-server/src/dune.ml
rgrinberg: Indeed. Thanks.

rgrinberg:
Okay, things are working solid now. Feel free to test with dune from master. Merging now, but more improvements will follow shortly.

PR #445, opened 2021-05-15
Add versions to diagnostics

rgrinberg:


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -83,13 +83,13 @@ let send_diagnostics ?diagnostics rpc doc =
...
-  let create_publishDiagnostics uri diagnostics =
+  let create_publishDiagnostics ~version uri diagnostics =
     Server_notification.PublishDiagnostics
-      (PublishDiagnosticsParams.create ~uri ~diagnostics ())
+      (PublishDiagnosticsParams.create ?version ~uri ~diagnostics ())
   in
   match diagnostics with
   | Some diagnostics ->
-    let notif = create_publishDiagnostics uri diagnostics in
+    let notif = create_publishDiagnostics ~version:None uri diagnostics in
```
tmattio, line 92: `send_diagnostics` receives `doc`. Why not set the version of the diagnostic here?

rgrinberg: Some of the diagnostics are independent of the version of the document. Since they depend on whether the binary is present.

PR #450, opened 2021-05-19
Remove explicit schedulers

rgrinberg:
Anything that needs the scheduler must now return Fiber.t

This brings us our code closer to dune's scheduler which we'll eventually adopt.

### DIFF
--- a/fiber-unix/src/scheduler.ml
+++ b/fiber-unix/src/scheduler.ml
@@ -257,7 +257,8 @@ let async_exn t f =
 
 let stop (t : thread) = Worker.stop t.worker
 
-let cancel_timers t =
+let cancel_timers () =
+  let* t = Fiber.Var.get_exn me in
   let timers = ref [] in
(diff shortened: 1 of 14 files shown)


fiber-unix/src/scheduler.ml
```diff
@@ -167,8 +168,6 @@ let is_empty table = Table.length table = 0
 
 let me = Fiber.Var.create ()
 
-let scheduler () = Fiber.Var.get_exn me
```
tmattio, line 170: You use `Fiber.Var.get_exn me` at different places. Maybe it's worth keeping a function for it?

tmattio (approved):

fiber-unix/src/scheduler.ml
rgrinberg: I'm trying to refrain from helpers that contain no logic and are just chains of projections.

PR #460, opened 2021-06-08
Report progress for dune

rgrinberg:
This makes it possible for the user to track their build progress inside the the editor.

Currently, this can't be tested because of some bugs in dune (will be fixed shortly).


ocaml-lsp-server/src/state.mli
```diff
@@ -13,7 +13,6 @@ type t =
   ; trace : TraceValue.t
   ; ocamlformat : Fmt.t
   ; diagnostics : Diagnostics.t
-  ; dune : Dune.t
```
tmattio, line 16: Interesting.. That's because the dune instance is not used anywhere except in the `start` function?
FWIU, ocamlformat and dune have a similar lifecycle, can't we get rid of `ocamlformat` in the state as well?

ocaml-lsp-server/src/progress.mli
```diff
@@ -0,0 +1,17 @@
+open Import
+
+type t
+
+val create :
+     ClientCapabilities.t
+  -> (Server_notification.Progress.t ProgressParams.t -> unit Fiber.t)
+  -> (WorkDoneProgressCreateParams.t -> unit Fiber.t)
```
tmattio, line 8: I would maybe add labeled arguments. It's not immediately clear from the signature what these functions do (for me at least)

ocaml-lsp-server/src/progress.ml
```diff
@@ -0,0 +1,78 @@
...
+      let open Fiber.O in
+      let* () = end_build t ~message:"Starting new build" in
+      let token = `String ("dune-build-" ^ Int.to_string t.build_counter) in
+      t.token <- Some token;
+      t.build_counter <- t.build_counter + 1;
+      t.create_task (WorkDoneProgressCreateParams.create ~token))
+
+let build_progress t (progress : Drpc.Progress.t) =
+  match t with
+  | Disabled -> Code_error.raise "progress reporting is not supported" []
```
tmattio, line 55: Raising seems odd to me: the `Enabled/Disabled` state is hidden from the user, so I would expect `build_progress` to do nothing if we are `Disabled`, instead of raising. This would also allow us to get rid of the check for `should_report_build_progress`

ocaml-lsp-server/src/progress.ml
```diff
@@ -0,0 +1,78 @@
...
+           (Server_notification.Progress.End
+              (WorkDoneProgressEnd.create ~message ())))
+
+let end_build_if_running = function
+  | Disabled -> Fiber.return ()
+  | Enabled e -> end_build e ~message:"Build interrupted"
+
+let build_event t (event : Drpc.Build.Event.t) =
+  match t with
+  | Disabled -> Code_error.raise "progress reporting is not supported" []
```
tmattio, line 39: Same comment as below

ocaml-lsp-server/src/progress.ml
```diff
@@ -0,0 +1,78 @@
...
+    match event with
+    | Finish -> end_build t ~message:"Build finished"
+    | Fail -> end_build t ~message:"Build failed"
+    | Interrupt -> end_build t ~message:"Build interrupted"
+    | Start ->
+      let open Fiber.O in
+      let* () = end_build t ~message:"Starting new build" in
+      let token = `String ("dune-build-" ^ Int.to_string t.build_counter) in
+      t.token <- Some token;
+      t.build_counter <- t.build_counter + 1;
```
tmattio, line 50: `build_counter` is the number of targets built by dune?
If so, this is going to be **a lot** of notifications. can we send only some of them? I wouldn't expect users to notice a difference if we only send e.g. 1/10.

ocaml-lsp-server/src/progress.ml
```diff
@@ -0,0 +1,78 @@
...
+  ; progress_report :
+      Server_notification.Progress.t ProgressParams.t -> unit Fiber.t
+  ; create_task : WorkDoneProgressCreateParams.t -> unit Fiber.t
+  }
+
+type t =
+  | Disabled
+  | Enabled of enabled
+
+let create (client_capabilities : ClientCapabilities.t) progress_report
```
tmattio, line 15: I would only send the relevant capability here, to reduce the dependencies of the module

tmattio (approved):
Looks good! I left a few comments, but nothing blocking, except maybe the performance concern for sending a notification per target, but feel free to merge as-is if you think this won't be an issue.

ocaml-lsp-server/src/progress.ml
```diff
@@ -0,0 +1,78 @@
...
+let build_progress t (progress : Drpc.Progress.t) =
+  match t with
+  | Disabled -> Code_error.raise "progress reporting is not supported" []
+  | Enabled { token; progress_report; _ } ->
+    let token =
+      match token with
+      | Some token -> token
+      | None -> Code_error.raise "no running build to report progress" []
+    in
+    Format.eprintf "reporting progress@.%!";
```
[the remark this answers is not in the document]
rgrinberg: Whoops, we don't need this line at all. Deleted.

ocaml-lsp-server/src/progress.ml
rgrinberg: build_encounter is the internal counter for the number of builds started by dune. We just use this number to create a progress reporting token for every build. We do send a lot of notifications, but not on build events, but rather on build progress. I'll see how it works, if it's slowing  the client, i will debounce accordingly.

tmattio: Great, thanks for the clarification :)

ocaml-lsp-server/src/state.mli
rgrinberg: > ocamlformat and dune have a similar lifecycle, can't we get rid of ocamlformat in the state as well?

It's possible but more difficult. We need an instance of ocamlformat to serve the formatting of requests.

I would like to get rid of this state thing eventually, but a proper design to replace it needs to come up first.

ocaml-lsp-server/src/progress.mli
rgrinberg: done

ocaml-lsp-server/src/progress.ml
rgrinberg: It was like that before, but the checks were added to safe guard us from a very annoying bug: we are wasting a lot of cycles getting progress information from dune without making sure that we are going to do something productive with this information. I changed it to a programming error to make sure this doesn't happen.

ocaml-lsp-server/src/progress.ml
rgrinberg: If the option was always present, I would do that, but I'd rather not split the logic for the default value and the flag itself across functions.

PR #465, opened 2021-06-14
Support merlin `holes` command

ulugbekna:
I implemented a custom request since the functionality didn't fit with any existing request. The user can jump from any location to the first hole that comes after that position. If there are no position _after_ the current one, the cursor jumps to the first hole in the file.

In vscode: 

https://user-images.githubusercontent.com/16353531/121877703-55fb9580-cd24-11eb-8473-b30267552339.mov

Soon will make a PR to the VOP (vscode ocaml platform) repo with the code handling this request

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,26 @@
 
 - Add support for navigating to a symbol inside a workspace (#398)
 
+- Add support for jumping to the next typed hole
+  - Note: implemented as a custom request, this will require a language server client that
+    knows about this request. At the moment, only VS Code OCaml Platform client support is
+    planned.
+  - Merlin has a concept of "typed holes" that are syntactically represented as `_`. Files
+    that incorporate typed holes are not considered valid OCaml, but Merlin and OCaml-LSP
+    support them. One example when such typed holes can occur is when on "destructs" a
+    value, e.g., destructing `(Some 1)` will generate code
+
+    ```ocaml
+    match Some 1 with Some _ -> _ | None -> _
+    ```
+
+    While the first underscore is a valid "match-all"/wildcard pattern, the rest of
+    underscores are typed holes. It is reasonable that user wants to jump to such typed
+    holes to be able to edit them. One should note that at the time of this writing, there
+    is work on Merlin, which allows to generate real code for such typed holes
+    automatically based on the type of the hole and available values in the current
+    context.
+
 # 1.6.1 (05/17/2020)
 
 ## Fixes
--- a/ocaml-lsp-server/docs/ocamllsp/nextHole-spec.md
+++ b/ocaml-lsp-server/docs/ocamllsp/nextHole-spec.md
@@ -1,0 +1,45 @@
+# Next Hole Request
+
+Merlin has a concept of "typed holes" that are syntactically represented as `_`. Files
+that incorporate typed holes are not considered valid OCaml, but Merlin and OCaml-LSP
+support them. One example when such typed holes can occur is when on "destructs" a value,
+e.g., destructing `(Some 1)` will generate code `match Some 1 with Some _ -> _ | None ->
+_`. While the first underscore is a valid "match-all"/wildcard pattern, the rest of
+underscores are typed holes.
+
+It is reasonable that user wants to jump to such typed holes to be able to edit them. One
+should note that at the time of this writing, there is work on Merlin, which allows to
+generate real code for such typed holes automatically based on the type of the hole and
+available values in the current context.
+
+## Client capability
+
+nothing that should be noted
+
+## Server capability
+
+property name: `handleNextHole`
+
+property type: `boolean`
+
+## Request
+
+- method: `ocamllsp/nextHole`
+- params:
+
+```json
+{
+  "uri": DocumentUri,
+  "position": Position
+}
+```
+
+## Response
+
+- result: `Range | Null`
+  - `Range` if the next hole was found
+  - `Null` if there are no holes in the file
+- error: code and message set in case an exception happens during the
+  `ocamllsp/nextHole` request.
+  - in case of any errors in finding a next hole in a file, the handler throws an
+    exception, which is returned from the language server.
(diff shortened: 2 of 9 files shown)


rgrinberg:
Do holes show up in the diagnostics currently?

ulugbekna:
> Do holes show up in the diagnostics currently?
As errors? No. I think merlin handles them as valid syntax. 

image

rgrinberg:
The holes prevent the code from building. Thus they should be displayed as errors in the diagnostics list.

In any case, however the holes are displayed in the diagnostics, you should be able to get at them from the client side by fetching the diagnostics and filtering. From there, you can navigate to the next, previous, or any other hole present in the document.

Rudi.

ulugbekna:
Hi! I’ve been experimenting with a way to implement “jump to the first hole in a destruct” and have a quick design discussion.

Currently destruct is implemented as a code action that sends a range and a new code that should replace the current code in that range. We want to jump to the first typed hole (`_` but not a wildcard pattern (?)) in that new code. I think there are two ways to achieve this: 

1. Since code actions allow to embed a command in them (the command is run after the edit is made) but those commands are usually editor-specific, we can 
  1.1. Add support for knowing what editor client has (if we go this way I plan to implement "jump to first hole" only on vscode but can try extend to other editors if that's possible)
  1.2. When we know that the editor is vscode, we can send an empty edit that cleans the current range and send a vscode-specific `editor.action.insertSnippet` command to insert the destruct code with holes as snippets. 

2. Since we have `ocaml.next-hole` implemented, we could reuse that already implemented command, but it depends on holes as error diagnostics. The problem is that when the code action is made, the command is run before the user gets diagnostics for the newly edited code. Thus, we need a custom request to get holes and another command similar to `ocaml.next-hole` that uses the custom request. I think having "next hole" commands to use different ways of finding holes will lead to puzzling behavior for users. 

I thought about pull-based diagnostics implemented as custom requests, where `ocaml.next-hole` command would pull in the diagnostics and then jump to the first hole, but that seems like jumping through many loops to avoid a custom request to get holes and definitely less efficient, as getting all diagnostics should be more expensive than just holes. 

cc @rgrinberg

rgrinberg:
Both solutions are fine to me.

The first solution will receive first class support eventually as this ticket acknowledges the limitation: https://github.com/microsoft/language-server-protocol/issues/724

Rudi.

rgrinberg:
> I thought about pull-based diagnostics implemented as custom requests, where ocaml.next-hole command would pull in the diagnostics and then jump to the first hole, but that seems like jumping through many loops to avoid a custom request to get holes and definitely less efficient, as getting all diagnostics should be more expensive than just holes.

I should have replied to this point as well. The "efficiency" gained here isn't relevant. The diagnostics already have to be recomputed after a destruct is inserted, sending a few extra entries in this list across the wire will not affect the performance in any noticeable way compared to all the other work that has to be done to serve this request.

As for jumping through many loops to avoid custom requests, I should clarify why I dislike custom requests so much. In my humble opinion, the implementation of complex should features should isolated to as little components as possible. Yes, in this situation it's impossible to avoid the fact that we need to orchestrate this feature across 3 different projects, but can we at least keep one component as dumb, stateless, and simple as possible? Since this an editor specific feature, I think it should be up to the editor to handle all the "glue" if it is less convenient to do so. Doing it this way, you will be able to continuously improve your feature without worrying about moving an additional project in lockstep.

rgrinberg:
> I must say, however, that in the particular case of holes and diagnostics I feel like that'd be tight coupling of not very related things

That's where I disagree. In my eyes, the two are intimately related.

> For example, if we at some point allow to configure diagnostics in various ways (turn off syntax errors when editing, clear diagnostics on request, etc.), we would keep facing different issues with our approach to piggyback holes functionality onto diagnostics, which may make touching diagnostics error-prone/ugly/undesirable;

How is it possible to turn off syntax errors when editing without turning off holes? As far as I know, merlin needs a syntactically valid source before it can detect holes. If you mean just not showing syntax errors in the diagnostics window, then the client already has all the necessary information to display errors according to the user's preference.

> Another thing is that even when the pull-based diagnostics land, since it is the client who decides to pull diagnostics, ocaml-lsp will still be dependent on custom commands on the client side to ask client to pull diagnostics (unless lsp standardizes client commands and all clients will implement same interface for their commands), which means that ocaml-lsp again would most likely only support vscode but also deal with diagnostics-holes problems

This is what pulling diagnostics will look like on the client side: https://github.com/microsoft/vscode-languageserver-node/blob/main/protocol/src/common/proposed.diagnostics.md#textDocument_diagnostic

So it should be quite portable to all editors.

ulugbekna:
> How is it possible to turn off syntax errors when editing without turning off holes? As far as I know, merlin needs a syntactically valid source before it can detect holes.

1. I think that it does not necessarily need valid OCaml code to detect holes.

image

2. So if we use a custom request (and not diagnostics) to get information about holes in the file, users will be able to jump around holes _without latest diagnostics_. 

2.1. One of the examples how we can benefit from this: allow users to get diagnostics not on file edit but on file save, which is a reasonable thing to ask [1]. Of course, we could send holes as diagnostics on file edits even when user configures diagnostics (errors and warnings) to be sent only on file save, but that seems to create a somewhat perplexing case in which holes are and are not diagnostics at the same time, so we'll need to explain why we still send diagnostics even when the user configures not to send them. 

2.2. There are vscode-ocaml-platform users who disable diagnostics and get their errors & warnings from the compiler/dune. Those people would not be able to use jump around holes functionality, which I think is quite unrelated to getting their diagnostics.

All in all, holes seem to be a feature introduced by merlin to 
- (?) have holes in the code that's being written without getting type errors (also conciser than `failwith "implement me"`)
- allow using `merlin construct` with more ease
so tying that feature with errors & warnings seems limiting.

[1] I think it is reasonable because diagnostics on edit usually not very useful since the user isn't done completing their code which results in already known syntax and type errors. Not getting these diagnostics also decreases typing latency, AFAICT.

PR #466, opened 2021-06-21
Show typed holes as errors

ulugbekna:
image

https://github.com/ocamllabs/vscode-ocaml-platform/pull/640 uses this PR to add support for commands to jump to previous/next typed hole. 

This is the first PR in a series to add support for merlin `Construct` command in ocaml-lsp.

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -157,15 +157,12 @@ let set_diagnostics rpc doc =
               in
               let holes_as_err_diags =
                 Query_commands.dispatch pipeline Holes
-                |> List.map ~f:(fun (loc, _type) ->
+                |> List.map ~f:(fun (loc, typ) ->
                        let range = Range.of_loc loc in
                        let severity = DiagnosticSeverity.Error in
                        let message =
-                         "This is a typed hole.\n\
-                          If you meant to use a wildcard pattern, here an \
-                          expression is expected not a pattern. Your code will \
-                          not compile if you do not replace this with a valid \
-                          expression."
+                         "This typed hole should be replaced with an \
+                          expression of type " ^ typ
                        in
                        (* we set specific diagnostic code = "hole" to be able to
                           filter through diagnostics easily *)
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -83,8 +83,8 @@ let task_if_running (state : State.t) ~f =
 let set_diagnostics rpc doc =
   let state : State.t = Server.state rpc in
   let uri = Document.uri doc in
-  let create_diagnostic ?relatedInformation ?severity range message =
-    Diagnostic.create ?relatedInformation ?severity ~range ~message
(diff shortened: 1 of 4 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -125,33 +125,54 @@ let set_diagnostics rpc doc =
...
+                Query_commands.dispatch pipeline Holes
+                |> List.map ~f:(fun (loc, _type) ->
+                       let range = Range.of_loc loc in
+                       let severity = DiagnosticSeverity.Error in
+                       let message =
+                         "This is a typed hole.\n\
+                          If you meant to use a wildcard pattern, here an \
+                          expression is expected not a pattern. Your code will \
+                          not compile if you do not replace this with a valid \
+                          expression."
```
rgrinberg, line 168: Is it possible to include the type of the expression needed by the hole? Also, the error is a bit long in my eyes. I'd prefer something to the tune of "Typed hole awaiting an expression of type ..."`

ulugbekna: image

Does this look better?

rgrinberg: Looks much better now. Perhaps we should replace `typed hole` with just `hole`? There are no untyped holes, so the qualifier seems redundant.

rgrinberg:
I'd like to see a test case for this as well.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -125,33 +125,51 @@ let set_diagnostics rpc doc =
...
+                       let message =
+                         "This typed hole should be replaced with an \
+                          expression of type " ^ typ
+                       in
+                       (* we set specific diagnostic code = "hole" to be able to
+                          filter through diagnostics easily *)
+                       create_diagnostic ~code:(`String "hole") range message
+                         ~severity)
+              in
+              List.append holes_as_err_diags merlin_diagnostics)
```
rgrinberg, line 172: We should make sure that all diagnostics are ordered by location.

A test case for this would be useful too.

ulugbekna: I'll add test cases for new diagnostics and diagnostics ordering. Out of curiosity: why do we want the diagnostics to be ordered?

ocaml-lsp-server/src/ocaml_lsp_server.ml
ulugbekna: Great. I am just concerned that new (and perhaps even seasoned) ocamlers may not know what a "hole" is, so I made it "typed hole" to be more googlable & slightly more formal (if that makes sense). So if there are no strong objections, I

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: When one is walking through the document with "next diagnostic", we don't want the cursor to be jumping back and forth.

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: That's fine with me.

PR #467, opened 2021-06-24
Add custom request `ocamllsp/typedHoles`

ulugbekna:
Adds a custom request `ocamllsp/typedHoles`, which is used by https://github.com/ocamllabs/vscode-ocaml-platform/pull/643

### DIFF
--- a/ocaml-lsp-server/docs/ocamllsp/typedHoles-spec.md
+++ b/ocaml-lsp-server/docs/ocamllsp/typedHoles-spec.md
@@ -1,0 +1,48 @@
+# Typed Holes Request
+
+## Description
+
+### What typed holes are
+
+Merlin has a concept of "typed holes" that are syntactically represented as `_`. Files
+that incorporate typed holes are not considered valid OCaml, but Merlin and OCaml-LSP
+support them. One example when such typed holes can occur is when on "destructs" a value,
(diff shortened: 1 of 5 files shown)


ocaml-lsp-server/docs/ocamllsp/typedHoles-spec.md
```diff
@@ -0,0 +1,48 @@
+# Typed Holes Request
```
rgrinberg, line 1: The description should include the level of stability we guarantee for this request. In this case, the level of stability is 0 and we will drop support for this request as soon as pulling diagnostics for a document is possible (as far as I understand, this will come in 3.17).

ocaml-lsp-server/src/custom_requests/req_typed_holes.ml
```diff
@@ -0,0 +1,48 @@
+open Import
+
+let capability = ("handleTypedHoles", `Bool true)
+
+let meth = "ocamllsp/typedHoles"
+
+let failwith_err err = failwith @@ Printf.sprintf "%s: %s" meth err
```
rgrinberg, line 7: We have two kinds of errors, differing by what is their ultimate cause:

* When the cause is a programming error on the server side, we should raise a code error.
* When the cause is a misbehaving client, we should raise `Response.Error.E`.

Therefore, we should never use failwith.

PR #468, opened 2021-06-24
Jump to the first hole on destruct

ulugbekna:
- Is based on #467 and https://github.com/ocamllabs/vscode-ocaml-platform/pull/643
- I am not sure whether embedding in a code action a command, which is not available in an editor results in user-visible warning or error; if that's the case, I can add the command only for vscode client

The gif below shows how jumps to first holes work along with jumping to prev/next holes.

https://user-images.githubusercontent.com/16353531/123289585-05591900-d52a-11eb-97db-f5d94b1441b9.mov

### DIFF
--- a/ocaml-lsp-server/src/code_actions/action_destruct.ml
+++ b/ocaml-lsp-server/src/code_actions/action_destruct.ml
@@ -3,8 +3,9 @@ open Import
 let action_kind = "destruct"
 
 let code_action_of_case_analysis doc uri (loc, newText) =
+  let range : Range.t = Range.of_loc loc in
   let edit : WorkspaceEdit.t =
-    let textedit : TextEdit.t = { range = Range.of_loc loc; newText } in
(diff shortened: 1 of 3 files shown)


ocaml-lsp-server/src/code_actions/action_destruct.ml
```diff
@@ -15,8 +16,13 @@ let code_action_of_case_analysis doc uri (loc, newText) =
     WorkspaceEdit.create ~documentChanges:[ `TextDocumentEdit edit ] ()
   in
   let title = String.capitalize_ascii action_kind in
+  let command =
+    Command.create ~title:"Jump to First Hole" ~command:"ocaml.next-hole"
```
rgrinberg, line 20: "First Hole" implies to me that we are jumping to the first hole in the document. The title should match the next hole behavior.

rgrinberg (approved):

PR #472, opened 2021-07-07
Construct as completion suggestions

ulugbekna:
Implements support for merlin `Construct`. 

The main reservation that I have so far is that we're not flexible enough about the `depth` parameter for `Construct`. How can we fix this?  

From our discussion with voodoos: `Local` param for `Construct` isn't ready yet, and we should specify `depth` param later, when it's more efficient.

To discuss: Maybe it makes more sense to replace `_`s with snippets, eg calling construct in `_ arg1 arg2` will result in `(fun _ _ -> _) arg1 arg2` and the cursor will jump to the third underscore (typed hole), while the user likely wants to change the function arguments first.

Otherwise, I used the implementation myself. It seems quite good to me. Demo (the demo used Construct with param set to `Local to use local values, which we turned off now) 

https://user-images.githubusercontent.com/16353531/124743582-4f34fc80-df37-11eb-811a-1c6661ceff81.mov


ocaml-lsp-server/test/e2e/__tests__/textDocument-completion.test.ts
```diff
@@ -425,18 +425,201 @@ describe_opt("textDocument/completion", () => {
...
 
     let items: Array<any> = await queryCompletion(Types.Position.create(5, 8));
     expect(
       items.filter((compl) => compl.label === "x" || compl.label === "y"),
     ).toHaveLength(0);
   });
+
+  it("completion for holes", async () => {
+    await openDocument(outdent`
+let u : Lexing.position = _
```
rgrinberg, line 445: Could you modify this test to be:

* Less noisy. We don't need so many candidates in the results.
* Rely on the standard library as little as possible. That would mean defining all the types you are using. This would make the test more robust when upgrading to newer versions of OCaml

ocaml-lsp-server/src/code_actions/action_construct.ml
```diff
@@ -0,0 +1,54 @@
...
+    else
+      let code_action =
+        let command : Command.t =
+          Command.create ~title:"Trigger Suggest"
+            ~command:"editor.action.triggerSuggest" ()
+        in
+        CodeAction.create ~title:"Construct an expression"
+          ~kind:(CodeActionKind.Other action_kind) ~command ()
+      in
+      Document.with_pipeline_exn doc @@ fun pipeline ->
```
rgrinberg, line 30: I'd prefer not to use the `@@` form. That makes it easier to slip in accidental work into the merlin thread as you've noticed in your first commit.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -240,7 +243,7 @@ module Formatter = struct
     | Result.Error e ->
       let message = Fmt.message e in
       let error = jsonrpc_error e in
-      let msg = ShowMessageParams.create ~message ~type_:Error in
+      let msg = ShowMessageParams.create ~message ~type_:Warning in
```
rgrinberg, line 246: Is this change done on purpose? If so, I don't see the reasoning. A failure to format is an error not a warning.

ocaml-lsp-server/src/code_actions/action_construct.ml
```diff
@@ -0,0 +1,54 @@
...
+      let prefix = Compl.prefix_of_position ~short_path:false src pos in
+      let hole_syntax = "_" in
+      String.equal prefix hole_syntax
+    in
+    if not can_be_hole then
+      Fiber.return None
+    else
+      let code_action =
+        let command : Command.t =
+          Command.create ~title:"Trigger Suggest"
```
rgrinberg, line 24: This "trigger suggest" action deserves a comment to explain why it's needed.

ocaml-lsp-server/src/compl.ml
```diff
@@ -119,62 +115,124 @@ let range_prefix (lsp_position : Position.t) prefix : Range.t =
...
+  let+ r = Document.dispatch doc command in
+  match r with
+  | Error { exn = Merlin_analysis.Construct.Not_a_hole; _ } -> Error `Not_a_hole
+  | Error e -> Error (`Exn e)
+  | Ok (loc, exprs) ->
+    let range = Range.of_loc loc in
+    List.mapi exprs ~f:(fun ix expr ->
+        let textEdit = `TextEdit { TextEdit.range; newText = expr } in
+        let preselect =
+          if ix = 0 then
```
rgrinberg, line 182: Should just be `let preselect = ix = 0 in`.

Better yet is to handle this tweak after generating the list:

```ocaml
let items = List.mapi exprs ~f:(...) in
match items with
| x :: xs -> { x with preselect = true } :: xs
| [] -> []
```

This is not for optimization, I just find adding the a check that will only fire on the first element to be distasteful.

ocaml-lsp-server/src/compl.ml
```diff
@@ -119,62 +115,124 @@ let range_prefix (lsp_position : Position.t) prefix : Range.t =
...
+    let textDocument = TextDocumentIdentifier.create ~uri:(Document.uri doc) in
+    CompletionParams.create ~textDocument ~position:pos ()
     |> CompletionParams.yojson_of_t
   in
-  let items = List.mapi ~f:(item ~compl_info) items in
-  `CompletionList { CompletionList.isIncomplete = false; items }
+  List.mapi completion_entries
+    ~f:(completionItem_of_completion_entry ~range ~compl_params)
+
+let construct doc position =
```
rgrinberg, line 170: The return type for this function is needlessly informative. Let's go over the three cases:

* There's an exception we don't know how to handle. We can't do anything except bubble it up to the client. No use in informing the caller.
* We can provide completion candidates - okay, let's return them.
* There are no completion candidates (for whatever reason, no hole, nothing to suggest) - an empty list is all the caller needs to know.

Thus I expect function should return `CompletionItem.t list Fiber.t`.

rgrinberg (changes requested):
Thanks for your efforts.  I did an initial review and the PR should be good to go after some minor fixes.

rgrinberg:
> Maybe it makes more sense to replace _s with snippets, eg calling construct in _ arg1 arg2 will result in (fun _ _ -> _) arg1 arg2 and the cursor will jump to the third underscore (typed hole), while the user likely wants to change the function arguments first.

I would wait until it is possible to insert snippets with code actions before exploring this. It would be very confusing UI if the editor only offers snippets only via completions.

ocaml-lsp-server/test/e2e/__tests__/textDocument-completion.test.ts
ulugbekna: I updated the test to be minimally noisy, but the problem is that along with calling `construct`, we also look for completion for `_` prefix, which brings a lot of results such as `__FILE__`, `__POS__`, which we can't rid of (?), unless we do not offer completion when prefix is `_`. I thought about not offering completion on `_`, since usually variables starting with `_` are "not used", but that seems limiting from a user's perspective, if that makes sense.

ocaml-lsp-server/src/ocaml_lsp_server.ml
ulugbekna: Not on purpose, but would be great to have this change (maybe in a separate PR). Reasoning: 

When a user saves the file, they get this scary _error_ pop-up

image

And even though I KNOW (at this point) the file is saved, every time it feels like the file wasn't saved because there is an **error**. I'm sure there are others who feel same way. 

A failure to format is an error, but the error being shown on a file save seems bothering (at least to me)

ulugbekna:
I hope that trying to make two merlin dispatches at once (fcd625b) is worth the complexity. But I tried to make the code modular, readable, and maintainable.

ocaml-lsp-server/src/client.ml
```diff
@@ -0,0 +1,54 @@
...
+        Currently, the module includes custom commands introduced by "Ocaml
+        Platform" extension *)
+    module Custom : sig
+      (** Request client cursor to jump to the next hole. Looks for a hole
+          starting at position [start_position], if provided; otherwise, uses
+          the cursor position. Will not show a pop-up notification if
+          [notify-if-no-hole] is set to [false] (the default value is [true]) *)
+      val next_hole :
+           ?start_position:Position.t
+        -> ?notify_if_no_hole:bool
```
rgrinberg, line 26: As far as I can tell, this argument is only passed with `false`.
I didn't look closely at `start_position` but its "optionality" is also suspect. Please make it required unless there's strong justification for it.

ocaml-lsp-server/src/client.ml
```diff
@@ -0,0 +1,54 @@
+open Import
+
+module Vscode = struct
```
rgrinberg, line 3: This module is missing `client.mli`. All the useful comments would be better placed there.

ocaml-lsp-server/src/compl.ml
```diff
@@ -134,62 +130,152 @@ let range_prefix (lsp_position : Position.t) prefix : Range.t =
...
+          (construct_cmd_resp, compl_by_prefix_resp))
+    in
+    let construct_completionItems =
+      Complete_with_construct.process_dispatch_resp construct_cmd_resp
+    in
+    let compl_by_prefix_completionItems =
+      match compl_by_prefix_resp with
+      | Ok resp -> Complete_by_prefix.process_dispatch_resp doc pos resp
+      | Error _ ->
+        (* TODO: log the error in future *)
```
rgrinberg, line 271: All of this `dispatch_cmd` vs `dispatch_cmd_exn` stuff was introduced to make it possible to recover hole suggestions if completion fails, right? That doesn't like a good trade-off to me. There's no need to try and recover from bugs in the server, let's just fail the request and move on. The important thing here is that the user can give us a useful bug report, and not if some partial functionality is preserved.

ocaml-lsp-server/test/e2e/__tests__/textDocument-completion.test.ts
```diff
@@ -533,4 +533,177 @@ let x : t = \`I
...
+              "line": 0,
+            },
+            "start": Object {
+              "character": 14,
+              "line": 0,
+            },
+          },
+        },
+      },
+      Object {
```
rgrinberg, line 562: To make this test less noisy, how about we filter all completion candidates that start with `__`?

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: Alright, I don't see much harm in making it a warning. We do need make this error message less noisy though.

rgrinberg (changes requested):
> I hope that trying to make two merlin dispatches at once (fcd625b) is worth the complexity. But I tried to make the code modular, readable, and maintainable.

It's definitely worth it.

The new code looks much better. Some more minor tweaks and it should be good to go.

ocaml-lsp-server/src/client.ml
ulugbekna: This function aims to replicate 1-1 the interface provided by that command, so those configs are optional as in the command's interface. It is true that we now usually use both of these optional commands, but that may not be true in the future, and for example, a new comer will find this function's new interface puzzling or limiting. So my goal here was to keep 1-1 interface with the command. Does this make a strong justification? 

---

A note about the command: `ocaml.next-hole` is a command introduced by vscode-ocaml-platform, that has optional (both) arguments "position" and "notify-if-no-hole". They both have default values, where "position" defaults to the current position of the cursor, and "notify-if-no-hole" defaults to `true`.

ocaml-lsp-server/src/compl.ml
ulugbekna: If your main argument is to avoid failing silently: 

I agree that failing silently is undesirable, but I think showing a warning message there may be slightly more appropriate there:

1. If the completion by prefix fails at that particular place, it's very likely because of a merlin problem, not server. 
2. If a user doesn't get completion suggestions once in a while (since completion bugs are quite rare from my experience), they go on with their lives, but if they get a warning message prompting to submit a bug report with an appropriate log with the backtrace from the server, they are much more likely to report things. 

So I think it may be better here to show a warning, log the exception, and offer our best completion suggestions we could deliver. Wdyt? 

If you more want to avoid the redundancy of `dispatch_cmd` and `dispatch_cmd_exn`, I also agree. I could leave only non-exn one and adapt `process_dispatch_resp` to work with a result, but it would just be reraising.

Side notes: 
- I'm really sorry for long argumentation. I really have the best in mind for the user.
- We miss a good logger in ocaml-lsp. In my understanding, if we put the logger in `State.t`, we will have to propagate it as an argument across all functions, say in this PR: `complete -> dispatch_cmd -> ...`. Should we maybe have a global logger (in `lsp-fiber` library?) that keeps an option ref, initialized by ocaml-lsp-server?

rgrinberg: The issue that I'm pointing out isn't that we're failing silently, but that we're trying too hard to recover from buggy code. The instant we know we're in the buggy path, we should just abort everything and do our best to guide the user to submit a bug report. Of course we can't take this literally and  terminate the server in such instances - it would make the user experience very bad, but this particular edge case doesn't deserve any special treatment.

The code inside the dispatch should just get the completion and typed hole candidates. Should this fail for any reason, then it's OK to just abort the entire completion request. The user will get an appropriate error message in their client and they can submit a bug report. 

> they go on with their lives, but if they get a warning message prompting to submit a bug report with an appropriate log with the backtrace from the server, they are much more likely to report things.

If a completion request fails, then they will see an error message. Is this not enough of an incentive to report the bug? If so, bugs in all other requests are being left unreported.

> So I think it may be better here to show a warning, log the exception, and offer our best completion suggestions we could deliver.

In this particular case, I do not think that "best effort" is a good policy. The benefits are very slim and the additional code is extra weight.

> Thank you for patience for my long arguments

Not at all, I always try to make time for those who contribute to this project. I sense that there's a bit of a difference in maintenance philosophy between us. This project both moves fast and has small dev resources behind it, therefore I'd like it if every line of code would carry its weight. That also means that we are going to leave some edge cases unpolished if the failure mode isn't catastrophic.

ocaml-lsp-server/src/client.ml
rgrinberg: I see two different concerns here:

* Future proofing the need to use the full interface vscode provides us. I'm not a fan, but I can let it be if there's some vague plans to actually use this stuff in the future.

* Using optional arguments in the API. Let's consider the newcomer you mentioned, should we make it easy for him to avoid thinking about which arguments to provide when using functionality? It seems like an easy way to introduce bugs.

> They both have default values, where "position" defaults to the current position of the cursor,

That is a concerning API then. I would much prefer if the current position was an explicit value.

> and "notify-if-no-hole" defaults to true.

I do not see what is gained by not providing the intended value every time.

ocaml-lsp-server/src/compl.ml
ulugbekna: > If a completion request fails, then they will see an error message.

Just to be on the same page, in vscode, there will be no error message shown, it's just there will be no completion suggestions, that's it. (same situation with all exceptions thrown by ocaml-lsp) The user simply doesn't get completion suggestions and needs to infer that they might need to go to the `Output` pane to see if there is an exception message buried somewhere, assuming (1) they know about the output pane (I'm working on doc "how to debug") and (2) they have verbose mode on. Both assumptions are rare in my opinion. And in the worst case, they get their language server crash 5 times (very unlikely given how stable completion is), at which point they'll need to reload vscode. (this info is more FYI for future than to change your mind on this; I'm okay with an exception, as completion has been quite stable) 

> Not at all, I always try to make time for those who contribute to this project. 

I appreciate it a lot! :-)

rgrinberg: That's a very good point. I've been meaning to explore a more systematic way of reporting these issues using your metrics PR + [telemetry](https://github.com/microsoft/vscode-extension-telemetry). We should revisit this at some point. In any case, this seems like an issue for all requests (what happens when offering code actions fails?), so we should treat it uniformly.

For now, we can add a workaround that sends a warning hinting that we may need a bug report if a request fails.

ocaml-lsp-server/src/compl.ml
```diff
@@ -208,15 +208,26 @@ module Complete_with_construct = struct
     | Error e -> Exn_with_backtrace.reraise e
     | Ok (loc, constructed_exprs) ->
       let range = Range.of_loc loc in
+      let deparen_constr_expr expr =
+        if
+          (not (String.equal expr "()"))
+          && Char.equal expr.[0] '('
+          && Char.equal expr.[String.length expr - 1] ')'
```
rgrinberg, line 215: How about:

```ocaml
if String.is_prefix expr ~prefix:"(" && String.is_suffix expr ~suffix:")" then
  String.sub expr ~pos:1 ~len:(String.length expr - 2)
else
  expr
```
This way you have 1 less conditional and you don't need to think about the empty string case.

rgrinberg (approved):
Pushed a few tweaks.

Looks good to go.

PR #473, opened 2021-07-08
Fix completion for polymorphic variants

ulugbekna:
When reconstructing prefix, we were ignoring backtick, which resulted in wrong completion entries, see the first commit.

I modified prefix building function to 
- not ignore a backtick 
- make sure that if the prefix that we were building already included a dot, e.g., the prefix so far is `Fiber.return`, we ignore the backtick that, say, comes before that prefix ie `Fiber.return. (I understand that the user shouldn't put the backtick there in the first place, but let's provide the best completions when we can)

I used some mutations along with `String.rfindi` instead of hand-writing the loop because I though it's still easier to reason about that a hand-written loop.

### DIFF
--- a/ocaml-lsp-server/src/compl.ml
+++ b/ocaml-lsp-server/src/compl.ml
@@ -46,35 +46,50 @@ let prefix_of_position ~short_path source position =
       min (String.length text - 1) (index - 1)
     in
     let pos =
-      let ident_or_infix_char = function
-        | 'a' .. 'z'
-        | 'A' .. 'Z'
-        | '0' .. '9'
-        | '\''
-        | '_'
-        (* Infix function characters *)
-        | '$'
-        | '&'
-        | '*'
-        | '+'
-        | '-'
-        | '/'
-        | '='
-        | '>'
-        | '@'
-        | '^'
-        | '!'
-        | '?'
-        | '%'
-        | '<'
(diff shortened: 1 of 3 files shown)


ocaml-lsp-server/src/compl.ml
```diff
@@ -46,35 +46,50 @@ let prefix_of_position ~short_path source position =
...
+          | ':'
+          | '~'
+          | '#' ->
+            true
+          | '`' ->
+            if !has_seen_dot then
+              false
+            else (
+              should_terminate := true;
+              true
```
rgrinberg, line 85: This is a bit tidier:
```ocaml
if not !has_seen_dot then should_terminate := true;
not !has_seen_dot
```

Alternatively, we could track the negative with `waiting_for_dot` and remove the extra `not`.

rgrinberg (approved):
Looks good. Please add a CHANGES entry since this is bug of our own making.

ocaml-lsp-server/src/compl.ml
ulugbekna: If you do not have strong preference here, I'd keep the current version: it seems simpler to understand (to me).

ulugbekna:
Thanks for the review! I added the changes entry.

PR #475, opened 2021-07-11
Improve display of merlin errors

rgrinberg:
Dune already has a library to parse compiler errors into a more structured form for better error reporting in watch mode. So we might as well use it to improve the errors we get from merlin.

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -83,6 +83,70 @@ let task_if_running (state : State.t) ~f =
   | false -> Fiber.return ()
   | true -> Fiber.Pool.task state.detached ~f
 
+let extract_related_errors uri raw_message =
+  match Ocamlc_loc.parse_raw raw_message with
+  | `Message message :: related ->
+    let string_of_message message =
+      String.trim
+        (match (message : Ocamlc_loc.message) with
+        | Raw s -> s
+        | Structured { message; severity; file_excerpt = _ } ->
+          let severity =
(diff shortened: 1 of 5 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -83,6 +83,67 @@ let task_if_running (state : State.t) ~f =
...
+              | None, None -> assert false)
+        in
+        String.trim (sprintf "%s: %s" severity message)
+    in
+    let related =
+      let rec loop acc = function
+        | `Loc (_, loc) :: `Message m :: xs -> loop ((loc, m) :: acc) xs
+        | [] -> List.rev acc
+        | _ ->
+          (* give up when we see something unexpected *)
```
tmattio, line 111: Could we log this somewhere, so this doesn't go unnoticed?

ocaml-lsp-server/test/e2e/__tests__/textDocument-diagnostics.ts
```diff
@@ -165,6 +165,88 @@ describe("textDocument/diagnostics", () => {
...
+                          "line": 2,
+                        },
+                        "start": Object {
+                          "character": 2,
+                          "line": 2,
+                        },
+                      },
+                      "uri": "file:///test.ml",
+                    },
+                    "message": "Expected declaration
```
tmattio, line 210: This should be trimmed I imagine?

tmattio (approved):
LGTM! I left a couple of comments, but nothing critical.

PR #478, opened 2021-07-24
Fix: correctly rename var when it's punned

ulugbekna:
is based on #399 ; it's best if we merge that PR first

https://user-images.githubusercontent.com/16353531/126861794-3922c017-2585-4d0f-9672-def12ef35486.mov

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -590,10 +590,33 @@ let rename (state : State.t)
   let open Fiber.O in
   let+ locs = Document.dispatch_exn doc command in
   let version = Document.version doc in
+  let source = Document.source doc in
   let edits =
-    List.map locs ~f:(fun loc ->
+    List.map locs ~f:(fun (loc : Warnings.loc) ->
         let range = Range.of_loc loc in
-        { TextEdit.newText = newName; range })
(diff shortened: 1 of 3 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -590,10 +590,30 @@ let rename (state : State.t)
...
+        let occur_start_pos =
+          Position.of_lexical_position loc.loc_start |> Option.value_exn
+        in
+        match occur_start_pos with
+        | { character = 0; _ } -> make_edit ()
+        | pos -> (
+          let mpos = Position.logical pos in
+          let (`Offset index) = Msource.get_offset source mpos in
+          let source_txt = Msource.text source in
+          match source_txt.[index - 1] with
```
rgrinberg, line 607: Can `index` ever be 0?

ulugbekna: As far as I understand, in order for `index` to be equal to 0, `line` and `character` both need to be equal to 0, but I handle case of `line = _; character = 0` separately, so we should never get `index = 0`. 

I'll add `assert (index > 0)`.

rgrinberg (approved):
Looks good. But you were better off creating a separate PR. I don't think this change conflicted in anyway with the qualify/unqualify stuff.

Issue #421, opened 2021-04-19
Fix log forwarding to client


rgrinberg:
#414 discovered that we don't forward logs correctly to the client. To fix this, we need a way to queue up logs and then proceed to send them in a fiber. So we need a blocking queue where the read end uses a fiber. E.g.

```ocaml
val read : 'a stream_in -> 'a Fiber.t
val write : 'a stream_out -> 'a -> unit
```

PR #480, opened 2021-08-02
Easier & cleaner logging

ulugbekna:
Sets up easier logging facility without passing around the `server` or a `logger` callback. Very much a proof of concept.

@rgrinberg What do you think of such a logger? It seems to make logging easier and cleaner. Kind of does what #421 says (?)

I am trying to improve logging generally, which would also include adding support for merlin logs.

### DIFF
--- a/ocaml-lsp-server/src/logger.mli
+++ b/ocaml-lsp-server/src/logger.mli
@@ -1,0 +1,32 @@
+open Import
+
+(** A logger module that allows to
+
+    - avoid passing a logging callback or server instance to functions that
+      don't actually need access to server for anything but logging
+    - set an application-wide logging level
+
+    Each log level has an small integer value given below. Setting [log_level]
+    on initialization tells the logger to ignore logs that have higher logging
+    value. For users, it makes sense to put [Info] or something even lower.
+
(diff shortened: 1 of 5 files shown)


ocaml-lsp-server/src/logger.mli
```diff
@@ -0,0 +1,32 @@
...
+  -> send_notification:(Server_notification.t -> unit Fiber.t)
+  -> unit
+
+val log_e : (unit -> string) -> unit Fiber.t
+
+val log_w : (unit -> string) -> unit Fiber.t
+
+val log_i : (unit -> string) -> unit Fiber.t
+
+val log_l : (unit -> string) -> unit Fiber.t
```
rgrinberg, line 32: In general, I don't like api's. Keep the API small whenever possible. In this case, a single function can do the work of all 4 with 1 extra argument.

rgrinberg:
> What do you think of such a logger? It seems to make logging easier and cleaner. Kind of does what #421 says

It's not quite the same. #421 suggests a solution to send logs to the client when outside the fiber. As far as I can tell, this PR does not address this limitation.

The rest of the changes seem OK to me. However, instead of using a global variable you should use a fiber variable to store the logger instance.

PR #481, opened 2021-08-02
Support Local mode in Construct -- user configuration

ulugbekna:
- adds support for `initializationOptions`, a field in the `initialize` request params
- adds support for Local mode for `Construct` 
- does some refactoring & clean up, especially in `import.ml`, which I can move to another PR if we want


rgrinberg:
The PR description and the code are not enough for me to understand the purpose of this configuration option. If it is not clear to me, it's probably not going to be clear to users how they should set this configuration option.

ulugbekna:
> The PR description and the code are not enough for me to understand the purpose of this configuration option. If it is not clear to me, it's probably not going to be clear to users how they should set this configuration option.

I included a more descriptive changelog entry here and in vscode-ocaml-platform PR https://github.com/ocamllabs/vscode-ocaml-platform/pull/675/commits/3918bef70952a9fdc7008e4db400de445271246e.

CHANGES.md
```diff
@@ -46,12 +46,30 @@
...
+  ```ocaml
+  type t = { name: string }
+
+  let john : t =
+    let doe = { name = "John" } in
+    _ (* <- construct is triggered on this typed hole *)
+  ```
+
+  With `Local` mode _on_, the suggested values will include the local variable `doe`
+  along with other expressions such as `{ name = _ }`.
```
rgrinberg, line 64: What would be the advantage of turning off this "Local" mode? To me it seems like a straight up improvement.

ulugbekna:
Local mode is unstable in merlin and can show many useless results. I
personally would still use the local mode, but other users may not like
such results. I should've added a note about instability of the feature and
that the setting may be removed in the future.

rgrinberg:
So it sounds like feature is still unstable and under active development, and the motivation for this configuration option is for devs to experiment and improve it further. I'm not against adding config options to facilitate this, but a lot of this config stuff is kind of overkill. You could have just as easily added an env var to toggle it.  Also, the documentation isn't even needed. This is not something that that user's will need to tinker with. Initially, I erroneously assumed that this was something for end users but it's clear that the they don't want this feature enabled until it's stable.

ulugbekna:
Will reopen in future with an implementation using an environment variable.

PR #482, opened 2021-08-05
Add support for a custom request to get AST node enclosing the cursor (for REPL support)

ulugbekna:
This PR adds support for "Evaluate expression under cursor".

This PR will soon have a counter-part in vscode-ocaml-platform repo. From that PR's changelog:

Let's see some examples (`<n>` is cursor position for example number `n`)

```ocaml
let k = <1> 1
<2>
module M = struct<3>
  let a =
    let <4> b = 1 in
    b + 1
  <5>
  let u = ()
end
```

| Your cursor | Evaluated expression        | Rationale                                                                                        |
| ----------- | --------------------------- | ------------------------------------------------------------------------------------------------ |
| <1>         | `let k = 1`                 | Toplevel expression under cursor                                                                 |
| <2>         | whole code block            | Cursor is in-between, so whole "file" is evaluated                                               |
| <3>         | `module M = ... end`        | Cursor is on the module definition                                                               |
| <4>         | `let a = let b = 1 in b + 1 | "Toplevel" expression for the cursor; it's used because it's "closer" than the module definition |
| <5>         | `let a = ... let u = ()`    | It's often useful to have module's definitions in the REPL environment                           

cc @arozovyk in case you're interested in the implementation of a custom request at the intersection of vscode-ocaml-platform, LSP, and merlin

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -76,6 +76,10 @@
   let f x = x.Unix.tms_stime, x.Unix.tms_utime
   ```
 
+- Add support for a custom request `ocamllsp/wrappingAstNode`, which brings the smallest
+  AST (Abstract Syntax Tree) node enclosing given position. This request is used by VS
+  Code OCaml Platform and is not directly used by OCaml LSP users (#482)
+
 - Bug fix: do not show "random" documentation on hover
   - fixed by [merlin#1364](https://github.com/ocaml/merlin/pull/1364)
   - fixes duplicate:
--- a/ocaml-lsp-server/src/custom_requests/custom_request.ml
+++ b/ocaml-lsp-server/src/custom_requests/custom_request.ml
@@ -1,0 +1,27 @@
+open Import
+
+type 't req_params_spec =
+  { params_schema : Jsonrpc.Message.Structured.t
+  ; of_jsonrpc_params : Jsonrpc.Message.Structured.t -> 't option
+  }
+
+let of_jsonrpc_params_exn spec params =
+  let raise_invalid_params ?data ~message () =
+    Jsonrpc.Response.Error.raise
+    @@ Jsonrpc.Response.Error.make ?data
+         ~code:Jsonrpc.Response.Error.Code.InvalidParams ~message ()
+  in
+  match params with
+  | None -> raise_invalid_params ~message:"Expected params but received none" ()
+  | Some params -> (
+    match spec.of_jsonrpc_params params with
+    | Some t -> t
+    | None ->
+      let error_json =
+        `Assoc
+          [ ("params_expected", (spec.params_schema :> Json.t))
+          ; ("params_received", (params :> Json.t))
+          ]
+      in
+      raise_invalid_params ~message:"Unexpected parameter format"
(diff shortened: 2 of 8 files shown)


ulugbekna:
I made the PR draft because I thought I'm missing support for mli files, but it seems like "evaluate expression" command doesn't need to work in mli files, e.g., 

```ocaml
─( 22:07:16 )─< command 6 >──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # 
val a : int
;;
Error: Value declarations are only allowed in signatures
```

which makes sense. Should be ready for a review. There are a couple of hiccups with vscode-ocaml-platform counterpart of the PR, which will soon be fixed. One should be able to test it already however.

CHANGES.md
```diff
@@ -97,8 +105,7 @@
   Merlin has a concept of "typed holes" that are syntactically represented as `_`. Files
   that incorporate typed holes are not considered valid OCaml, but Merlin and OCaml-LSP
   support them. One example when such typed holes can occur is when on "destructs" a value,
-  e.g., destructing `(Some 1)` will generate code `match Some 1 with Some _ -> _ | None ->
-  _`. While the first underscore is a valid "match-all"/wildcard pattern, the rest of
+  e.g., destructing `(Some 1)` will generate code `match Some 1 with Some _ -> _ | None -> _`. While the first underscore is a valid "match-all"/wildcard pattern, the rest of
```
rgrinberg, line 108: What happened to the wrapping here?

ocaml-lsp-server/src/custom_requests/custom_request.ml
```diff
@@ -0,0 +1,31 @@
+open Import
+
+type 't req_params_spec =
```
rgrinberg, line 3: mli?

ulugbekna: Added an mli file. I'm not 100% sure about how `req_params_spec` thing is going to work out, eg whether other contributors will understand how and why to use it, but it seems somewhat useful.

CHANGES.md
ulugbekna: Markdown autoformatting at play here. Sorry, fixed.

ocaml-lsp-server/src/custom_requests/req_wrapping_ast_node.ml
```diff
@@ -0,0 +1,63 @@
...
+  type t =
+    { text_document_uri : Uri.t
+    ; cursor_position : Position.t
+    }
+
+  let params_schema =
+    `Assoc
+      [ ("uri", `String "<DocumentUri>"); ("position", `String "<Position>") ]
+
+  let parse params : t option =
```
rgrinberg, line 17: I'd prefer name of `of_jsonrpc_params` or `of_params`.

rgrinberg (approved):
Looks correct. I'm actually skeptical this is necessary to develop a good repl exprience, but I don't see any issues with adding this.

PR #483, opened 2021-08-10
Fix a bug in removable queue & reimplement queue

ulugbekna:
I was implementing channels (as in concurrency primitive channels) to simplify some code in the scheduler and came across a bug in `Removable_queue` [1]. Fixed the bug. Noticed that the implementation seems to be somewhat complicated and thought to reimplement the queue via a sentinel-based doubly linked list, getting inspiration from https://github.com/janestreet/jenga/blob/114.04/tenacious/lib/ring.ml along the way.

The implementation is easy to understand module use of `Obj.magic` once, but even that simplifies code a lot. 

[1] I thought the problem was with channels, so that was a good opportunity to try the new ocamlearlybird integration in vscode-ocaml-platform -- I couldn't get it to work though :p 
[2] I also wrote some property-based tests but didn't include them because they are too incomplete. This was also the reason to take the list implementation into a toplevel library rather than putting its tests along scheduler tests.


doubly-linked/src/doubly_linked.mli
```diff
@@ -0,0 +1,32 @@
...
+(** Removes the first element in the list *)
+val detach_head : 'a t -> 'a option
+
+(** [append lst v] adds a new node containing [v] to the end of the list *)
+val append : 'a t -> 'a -> 'a node
+
+(** Removes the last element in the list *)
+val detach_tail : 'a t -> 'a option
+
+val detach : 'a node -> (unit, [ `Already_detached ]) result
```
rgrinberg, line 32: Do we need this (and other) additional function? I don't see it being used anywhere. If it's absolutely necessary for the tests, please mark it as such in the public api.

rgrinberg:
>The implementation is easy to understand module use of Obj.magic once, but even that simplifies code a lot.

I'd still get rid of it. Obj.magic is known to subtly break various optimizations and I'd prefer lsp more likely to work when using experimental features of the compiler such as flambda.

doubly-linked/src/doubly_linked.mli
ulugbekna: I was aiming for a more complete implementation of a linked list. I removed the unused code.

ulugbekna:
> I'd still get rid of it. Obj.magic is known to subtly break various optimizations and I'd prefer lsp more likely to work when using experimental features of the compiler such as flambda.

Remove its use.

fiber-unix/src/removable_queue.mli
```diff
@@ -3,9 +3,11 @@ type 'a t
 
 type 'a node
 
-val remove : _ node -> unit
+val remove : 'a node -> unit
```
rgrinberg, line 6: Tbh, I prefer the old style of not naming type variables that are unused.

rgrinberg (approved):

PR #488, opened 2021-08-13
Action add-rec

### LINKED ISSUE #319: "add `rec` keyword" code action
Code action "add `rec` keyword" if current let binding's name is used inside the binding AND it's not available in the environment already, i.e., 

```ocaml
let reverse acc = function | [] -> acc | x::xs -> reverse<code action proposed here>
```

but not here

```ocaml
let reverse = List.rev 

let reverse = reverse [1; 2; 3]
```

This should be useful _at least_ for beginners.

jfeser:
Add an "add missing rec keyword" action. Closes #319.

### DIFF
--- a/ocaml-lsp-server/src/code_actions/action_add_rec.ml
+++ b/ocaml-lsp-server/src/code_actions/action_add_rec.ml
@@ -1,0 +1,86 @@
+open Import
+
+let action_title = "Add missing `rec` keyword"
+
+let let_bound_vars bindings =
+  List.filter_map
+    ~f:(fun vb ->
+      match vb.Typedtree.vb_pat.pat_desc with
+      | Typedtree.Tpat_var (id, loc) -> Some (id, loc)
+      | _ -> None)
+    bindings
+
+(** If the cursor position is inside a let binding which should have a ret tag
+    and does not, return the Location.t of the binding. *)
+let has_missing_rec pipeline pos_start =
+  let open Option.O in
+  (* Find identifier under cursor *)
+  let* ident =
+    Compl.reconstruct_ident (Mpipeline.raw_source pipeline) pos_start
+  in
+
+  (* Find most enclosing nonrecursive let binding that binds ident *)
+  let typer = Mpipeline.typer_result pipeline in
+  let browse = Mbrowse.of_typedtree (Mtyper.get_typedtree typer) in
+  Mbrowse.enclosing (Mpipeline.get_lexing_pos pipeline pos_start) [ browse ]
+  |> List.find_map ~f:(function
+       | ( _
+         , Browse_raw.Structure_item
+             ({ str_desc = Tstr_value (Nonrecursive, bound); _ }, _) )
+       | ( _
+         , Browse_raw.Expression
+             { exp_desc = Texp_let (Nonrecursive, bound, _); _ } ) ->
+         let bound_vars = let_bound_vars bound in
+         if
+           List.exists bound_vars ~f:(fun (id, _) ->
+               String.equal ident (Ident.name id))
+         then
+           (* Return the location of the first pattern in the let binding (the
+              rec goes right before it) *)
+           let+ first_pat = List.hd_opt bound in
+           let first_pat_loc = first_pat.vb_pat.pat_loc in
+           { first_pat_loc with loc_end = first_pat_loc.loc_start }
+         else
+           None
+       | _ -> None)
+
+let code_action_add_rec uri diagnostics doc loc =
+  let edit =
+    let textedit : TextEdit.t =
+      { range = Range.of_loc loc; newText = "rec " }
+    in
+    let version = Document.version doc in
+    let textDocument =
+      OptionalVersionedTextDocumentIdentifier.create ~uri ~version ()
+    in
+    let edit =
+      TextDocumentEdit.create ~textDocument ~edits:[ `TextEdit textedit ]
+    in
+    WorkspaceEdit.create ~documentChanges:[ `TextDocumentEdit edit ] ()
+  in
+  CodeAction.create ~diagnostics ~title:action_title
+    ~kind:CodeActionKind.QuickFix ~edit ~isPreferred:false ()
+
+let code_action doc (params : CodeActionParams.t) =
+  let open Fiber.O in
+  let pos_start = Position.logical params.range.start in
+
+  let m_diagnostic =
+    List.find params.context.diagnostics ~f:(fun d ->
+        let is_unbound () =
+          String.is_prefix d.Diagnostic.message ~prefix:"Unbound value"
+        and in_range () =
+          match Position.compare_inclusion params.range.start d.range with
+          | `Outside _ -> false
+          | `Inside -> true
+        in
+        in_range () && is_unbound ())
+  in
+  match m_diagnostic with
+  | Some d ->
+    let+ loc =
+      Document.with_pipeline_exn doc (fun pipeline ->
+          has_missing_rec pipeline pos_start)
+    in
+    Option.map loc ~f:(code_action_add_rec params.textDocument.uri [ d ] doc)
+  | None -> Fiber.return None
--- a/ocaml-lsp-server/src/compl.ml
+++ b/ocaml-lsp-server/src/compl.ml
@@ -93,28 +93,35 @@ let suffix_of_position source position =
   match Msource.text source with
   | "" -> ""
   | text ->
+    let (`Offset index) = Msource.get_offset source position in
     let len = String.length text in
-    let from =
-      let (`Offset index) = Msource.get_offset source position in
-      min index (len - 1)
-    in
-    let len =
-      let ident_char = function
-        | 'a' .. 'z'
-        | 'A' .. 'Z'
-        | '0' .. '9'
-        | '\''
-        | '_' ->
-          true
-        | _ -> false
+    if index >= len then
+      ""
+    else
+      let from = index in
+      let len =
+        let ident_char = function
+          | 'a' .. 'z'
+          | 'A' .. 'Z'
+          | '0' .. '9'
+          | '\''
+          | '_' ->
+            true
+          | _ -> false
+        in
+        let until =
+          String.findi ~from text ~f:(fun c -> not (ident_char c))
+          |> Option.value ~default:len
+        in
+        until - from
       in
-      let until =
-        String.findi ~from text ~f:(fun c -> not (ident_char c))
-        |> Option.value ~default:len
-      in
-      until - from
-    in
-    String.sub text ~pos:from ~len
+      String.sub text ~pos:from ~len
+
+let reconstruct_ident source position =
+  let prefix = prefix_of_position ~short_path:false source position in
+  let suffix = suffix_of_position source position in
+  let ident = prefix ^ suffix in
+  Option.some_if (ident <> "") ident
 
 let range_prefix (lsp_position : Position.t) prefix : Range.t =
   let start =
--- a/ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
+++ b/ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
@@ -9,8 +9,24 @@ function findAnnotateAction(actions) {
   return actions.find((action) => action.kind == "type-annotate");
 }
 
+function findAddRecAnnotation(actions) {
+  return actions.find(
+    (action) =>
+      action.kind == "quickfix" && action.title == "Add missing `rec` keyword",
+  );
+}
+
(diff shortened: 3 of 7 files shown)


ulugbekna:
Fun! I had a similar PR lying around :-)

ulugbekna:
@jfeser I quickly went through the code. This looks promising! I see three things that we need to fix before going further with this: 

1. We can't use `Str` module because it's not thread-safe (the module contains a global mutable state). One can use `re`. But for this functionality, we shouldn't need a regexp engine: one can find the range occupied by the identifier that needs to get `rec ` prepended and prepend rec. For example, `let f x = f 1`: we find `f` with range `((0, 5), (0,6))`, we should be able to send a text edit with new text `rec ` and range `((0,5), (0,5))`. 
2. Here comes the second thing: you use `Typedtree.let_bound_idents`, but that function returns bound identifiers from patterns that do not support `rec` keyword, e.g., one can't do `let rec { a } = ...`; we should only get the identifier from `Tpat_var`. So the code action now returns "false positives"
3. We usually fail fast with exceptions, so we usually don't do `Error exn -> Exn_with_backtrace.reraise exn`
4. Should we have a slightly friendlier name for the code action than `Add-rec`? `Add rec keyword` or something

I include my (incomplete and possibly erroneous at places) implementation (it's from the merlin codebase, so there is no LSP-related functionality) if you want to compare for yourself:

```ocaml
  | Add_rec pos -> begin
    let typer = Mpipeline.typer_result pipeline in
    match Mtyper.get_typedtree typer with
    | `Interface _ -> None
    | `Implementation impl ->
      let structures = Mbrowse.of_typedtree (Mtyper.get_typedtree typer) in
      let pos = Mpipeline.get_lexing_pos pipeline pos in
      let enclosing = Mbrowse.enclosing pos [structures] in (

      Logger.notify ~section:"add-rec" "%s" @@ (
          String.concat ~sep:"\n" @@
            List.map enclosing ~f:(fun (_, node) -> Browse_raw.string_of_node node)
      );

      match enclosing with
      | [] -> None
      | (_, leaf)::ancestors -> begin

        let rec find_ident_in_ancestors ident ancestors: Location_aux.t option  =
          match ancestors with
          | [] -> None
          | (_,  node) :: rest_of_ancs -> begin
            let search_in_parent_node () = find_ident_in_ancestors ident rest_of_ancs in 

            match node with
            | Browse_raw.Expression {
                exp_desc = Texp_let (Asttypes.Nonrecursive, bdgs, _) ; _ 
              } -> begin
              let rec find_pat_with_same_ident = function 
                | Typedtree.Tpat_var (_, { txt; loc }) -> 
                  Option.some_if (String.equal txt ident) loc
                | _ -> None
              in
              match
                List.find_map bdgs ~f:(fun { Typedtree.vb_pat; vb_loc; _ } ->
                    find_pat_with_same_ident vb_pat.pat_desc)
              with
              | l -> Some l
              | exception Not_found -> search_in_parent_node ()
            end

            | Value_binding { vb_pat; vb_expr; vb_attributes; vb_loc } -> begin
              match vb_pat.pat_desc with
              | Tpat_var ( _ident , { txt; loc }) when String.equal txt ident ->
                Some loc
              | _ -> search_in_parent_node ()
            end

            | _ -> search_in_parent_node ()
          end

        in

        match reconstruct_identifier pipeline pos None with
        | [] -> None
        | { txt = ident; _ } :: _ ->
          find_ident_in_ancestors ident ancestors
      end
    )
  end
```

jfeser:
Thanks for taking a look!

> 1. We can't use `Str` module because it's not thread-safe...

Done. I'm using the location of the first pattern in the binding to place the rec.

> 2. Here comes the second thing: you use `Typedtree.let_bound_idents`...

Added a new function `let_bound_vars` that only looks at `Tpat_var`.

> 3. We usually fail fast with exceptions, so we usually don't do `Error exn -> Exn_with_backtrace.reraise exn`

Done.

> 4. Should we have a slightly friendlier name for the code action than `Add-rec`? `Add rec keyword` or something

Done.

One concern from me: I'm searching the parsetree for the identifier under the cursor. I think merlin has better ways to do this, but I'm not sure how they work.

ulugbekna:
1. 
> I'm searching the parsetree for the identifier under the cursor. I think merlin has better ways to do this, but I'm not sure how they work.

I think we can reuse `Compl.prefix_of_position` and `Compl.suffix_of_position` to reconstruct the identifier under the cursor.

```ocaml
let reconstruct_ident (source : Msource.t) (position : [< Msource.position ]) :
    string =
  let prefix = prefix_of_position ~short_path:false source position in
  let suffix = suffix_of_position source position in
  prefix ^ suffix
```

and then we should have a simpler `find_ident`:

```ocaml
let find_ident (pipeline : Mpipeline.t) (pos_start : [> Msource.position ]) :
    string option =
  let reconstructed_ident_s =
    Compl.reconstruct_ident (Mpipeline.raw_source pipeline) pos_start
  in
  match reconstructed_ident_s with
  | "" -> None
  | nonempty_ident_s -> (
    match Longident.parse nonempty_ident_s with
    | Longident.Lident ident -> Some ident
    | _ -> None)
```

(I gave this implementation a try. When I test manually, the code action works as intended, but `missing-rec-2` and `missing-rec-3` tests fail, even though the action works with their test code.)

2. Another thing I noticed: the code action can be called on any let-bound identifier. Should we show the action only when we have a `Unbound value ...` diagnostic linked to the code-action-params range? LSP should allow that.
3. LSP also allows to indicate which diagnostics are resolved by a code action. Doing the point (2) above, we'd get a nice workflow where user gets a diagnostic fixed when they call this code action. 
4. Together with (2) and (3), we could then change the kind of the code action to `QuickFix` and prioritize it in the code actions list, since it's a quick fix that's more likely to be used than other code actions (since they won't likely work with code with `Unbound value` diagnostic)

I hope this isn't overwhelming. Let me know if you need help.

jfeser:
> I think we can reuse `Compl.prefix_of_position` and `Compl.suffix_of_position` to reconstruct the identifier under the cursor.

Makes sense. I'll integrate that.

> 1. Another thing I noticed: the code action can be called on any let-bound identifier. Should we show the action only when we have a `Unbound value ...` diagnostic linked to the code-action-params range? LSP should allow that.
> 2. LSP also allows to indicate which diagnostics are resolved by a code action. Doing the point (2) above, we'd get a nice workflow where user gets a diagnostic fixed when they call this code action.
> 3. Together with (2) and (3), we could then change the kind of the code action to `QuickFix` and prioritize it in the code actions list, since it's a quick fix that's more likely to be used than other code actions (since they won't likely work with code with `Unbound value` diagnostic)

This would be nice! Ideally, this would be one of a family of quickfix actions that take care of diagnostics like unused value, unused open, etc. Are there any quickfix actions in the codebase that I could look at?

Also, how does lsp handle warnings from the compiler right now? If we want to respond to warnings, we'd like to have a `Warnings.t` and a location. That should be enough to handle the "unused X" cases.

ulugbekna:
> Ideally, this would be one of a family of quickfix actions that take care of diagnostics like unused value, unused open, etc.

Great idea! 

> Are there any quickfix actions in the codebase that I could look at?

Quick fixes are just code actions with the action kind = `CodeActionKind.QuickFix`, AFAIU. 

> how does lsp handle warnings from the compiler right now? 

We get diagnostics (errors and warnings) from merlin; see [code](https://github.com/ocaml/ocaml-lsp/blob/227a37559621c87c591a894a2dbb311d8aa2d298/ocaml-lsp-server/src/ocaml_lsp_server.ml#L192-L206)

> If we want to respond to warnings, we'd like to have a Warnings.t and a location. That should be enough to handle the "unused X" cases.

Hm, I think it is easier if we match on the warning as a string (eg, by matching the prefix `Unused `). If you look into code I linked above, currently, we take errors and warnings as strings, so there needs to be quite some work done if we want to pattern match on `Warnings.t`.

jfeser:
> Hm, I think it is easier if we match on the warning as a string (eg, by matching the prefix `Unused `). If you look into code I linked above, currently, we take errors and warnings as strings, so there needs to be quite some work done if we want to pattern match on `Warnings.t`.

Oh well. The strings probably don't change much.

Btw, `reconstruct_ident` doesn't work when the cursor is at the end of the file. I think there's an off by one error in `suffix_of_position`. Is there an easy way to add some unit tests? ppx_inline_test is pretty nice.

ulugbekna:
> Is there an easy way to add some unit tests? ppx_inline_test is pretty nice.

I don't have a good answer for this yet. If I have to debug something like that, I usually do printf debugging by raising an exception and seeing the output in the vscode `output` pane for `OCaml Language Server`. 

Maybe we could add `ppx_inline_test` as a dependency (should that happen in another dune profile to not depend on ppx_inline_test from end user's perspective?)

@rgrinberg do you have suggestions on how to make adding unit tests/debugging easier in `compl.ml` code?

ocaml-lsp-server/src/code_actions/action_add_rec.ml
```diff
@@ -0,0 +1,87 @@
...
+
+let let_bound_vars bindings =
+  let vars = ref [] in
+  let add id = vars := id :: !vars in
+  List.iter
+    ~f:(fun vb ->
+      match vb.Typedtree.vb_pat.pat_desc with
+      | Typedtree.Tpat_var (id, loc) -> add (id, loc)
+      | _ -> ())
+    bindings;
```
rgrinberg, line 17: This is better written with `List.filter_map`.

rgrinberg:
Adding ppx_inline_test to user end libraries doesn't work very well. The end to end tests for this feature seem sufficient to me, but if you really need some unit tests, you can add a test library and add some tests.

ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
```diff
@@ -601,4 +619,262 @@ let x = _
       ]
     `);
   });
+
+  it("", async () => {
```
ulugbekna, line 623: it would be nice to have names for tests; here it could be "add missing rec in toplevel let-binding"

ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
```diff
@@ -601,4 +619,262 @@ let x = _
...
+    let fn = "file:///missing-rec-1.ml";
```
ulugbekna, line 624: Doesn't `fn` strongly associate with the word `function` for you?

ulugbekna (commented):
I left some finishing-touch minor comments -- some of them are preferential and won't prevent a merge, I think

ocaml-lsp-server/src/code_actions/action_add_rec.ml
```diff
@@ -0,0 +1,102 @@
+open Import
+
+let debug = false
```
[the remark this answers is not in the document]
jfeser: Sure.

ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
jfeser: Not really, but changed to `uri`

ocaml-lsp-server/src/code_actions/action_add_rec.ml
```diff
@@ -0,0 +1,102 @@
...
+  in
+  CodeAction.create ~diagnostics ~title:action_title
+    ~kind:CodeActionKind.QuickFix ~edit ~isPreferred:false ()
+
+let code_action doc (params : CodeActionParams.t) =
+  let open Fiber.O in
+  let pos_start = Position.logical params.range.start in
+
+  let m_diagnostic =
+    List.find params.context.diagnostics ~f:(fun d ->
```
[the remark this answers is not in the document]
jfeser: I don't think the client has to return diagnostics in sorted order, so probably better not.

ocaml-lsp-server/src/compl.ml
```diff
@@ -93,28 +93,39 @@ let suffix_of_position source position =
...
-          true
-        | _ -> false
-      in
-      let until =
-        String.findi ~from text ~f:(fun c -> not (ident_char c))
-        |> Option.value ~default:len
+    if index >= len then
+      ""
+    else
+      let from = min index (len - 1) in
```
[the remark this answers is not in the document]
jfeser: In the current version, for `suffix_of_position "text" 4` returns `"t"`. I changed it to return `""`. This is what you want if the cursor is at the end of an identifier and we want to reconstruct the identifier with `prefix ^ suffix`.

ulugbekna:
Thanks @jfeser ! This's a cool feature that users will dearly use, I'm sure.

PR #502, opened 2021-09-03
Code actions for marking & removing unused code

jfeser:
Adds code actions for marking unused code (i.e. by prepending '_' to variable names) and for removing unused code. Currently it can remove unused types, modules, opens, constructors, and variables.

### DIFF
--- a/ocaml-lsp-server/src/code_actions/action_mark_remove_unused.ml
+++ b/ocaml-lsp-server/src/code_actions/action_mark_remove_unused.ml
@@ -1,0 +1,148 @@
+open Import
+
+let slice doc (range : Range.t) =
+  let src = Document.source doc in
+  let (`Offset start) = Msource.get_offset src @@ Position.logical range.start
+  and (`Offset end_) = Msource.get_offset src @@ Position.logical range.end_ in
+  String.sub (Msource.text src) ~pos:start ~len:(end_ - start)
+
+(* Return contexts enclosing `pos` in order from most specific to most
+   general. *)
+let enclosing_pos pipeline pos =
+  let browse =
+    Mpipeline.typer_result pipeline
+    |> Mtyper.get_typedtree |> Mbrowse.of_typedtree
+  in
+  Mbrowse.enclosing
+    (Mpipeline.get_lexing_pos pipeline @@ Position.logical pos)
+    [ browse ]
+
+(* `name` is an unused binding. `contexts` is a list of Mbrowse.t enclosing an
+   unused definition of `name`, in order from most general to most specific.
+
+   Returns an edit that silences the 'unused value' warning. *)
+let rec mark_value_unused_edit name contexts =
+  let open Option.O in
+  match contexts with
+  | Browse_raw.Pattern { pat_desc = Tpat_record (pats, _); _ } :: cs -> (
+    let m_field_edit =
+      List.find_map pats
+        ~f:
+          (function
+           | ( { loc = field_loc; _ }
+             , _
+             , { pat_desc = Tpat_var (ident, _); pat_loc; _ } )
+             when Ident.name ident = name ->
+             (* Special case for record shorthand *)
+             if
+               field_loc.loc_start = pat_loc.loc_start
+               && field_loc.loc_end = pat_loc.loc_end
+             then
+               let+ end_pos = Position.of_lexical_position pat_loc.loc_end in
+               TextEdit.
+                 { range = Range.create ~start:end_pos ~end_:end_pos
+                 ; newText = " = _"
+                 }
+             else
+               let+ start_pos =
+                 Position.of_lexical_position pat_loc.loc_start
+               in
+               TextEdit.
+                 { range = Range.create ~start:start_pos ~end_:start_pos
+                 ; newText = "_"
+                 }
+           | _ -> None
+            :    Longident.t Loc.loc
+                 * Types.label_description
+                 * Typedtree.value Typedtree.general_pattern
+              -> TextEdit.t option)
+    in
+    match m_field_edit with
+    | Some e -> Some e
+    | None -> mark_value_unused_edit name cs)
+  | Pattern { pat_desc = Tpat_var (ident, _); pat_loc = loc; _ } :: _ ->
+    if Ident.name ident = name then
+      let+ start = Position.of_lexical_position loc.loc_start in
+      TextEdit.{ range = Range.create ~start ~end_:start; newText = "_" }
+    else
+      None
+  | _ :: cs -> mark_value_unused_edit name cs
+  | _ -> None
+
+let code_action_mark_value_unused doc (diagnostic : Diagnostic.t) =
+  let open Option.O in
+  Document.with_pipeline_exn doc (fun pipeline ->
+      let var_name = slice doc diagnostic.range in
+      let pos = diagnostic.range.start in
+      let+ text_edit =
+        enclosing_pos pipeline pos
+        |> List.rev_map ~f:(fun (_, x) -> x)
+        |> mark_value_unused_edit var_name
+      in
+      let edit = Document.edit doc text_edit in
+      CodeAction.create ~diagnostics:[ diagnostic ] ~title:"Mark as unused"
+        ~kind:CodeActionKind.QuickFix ~edit ~isPreferred:false ())
+
+(* Takes a list of contexts enclosing a binding of `name`. Returns the range of
+   the most specific binding. *)
+let enclosing_value_binding_range name =
+  let open Option.O in
+  List.find_map ~f:(function
+    | Browse_raw.Expression
+        { exp_desc =
+            Texp_let
+              ( _
+              , [ { vb_pat = { pat_desc = Tpat_var (_, { txt = name'; _ }); _ }
+                  ; _
+                  }
+                ]
+              , body )
+        ; exp_loc
+        ; _
+        }
+      when name = name' ->
+      let* start = Position.of_lexical_position exp_loc.loc_start in
+      let+ end_ = Position.of_lexical_position body.exp_loc.loc_start in
+      Range.create ~start ~end_
+    | _ -> None)
+
+(* Create a code action that removes [range] and refers to [diagnostic]. *)
+let code_action_remove_range doc (diagnostic : Diagnostic.t) range =
+  let edit = Document.edit doc { range; newText = "" } in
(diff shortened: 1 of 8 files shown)


ocaml-lsp-server/src/code_actions/action_remove_unused.ml
```diff
@@ -0,0 +1,142 @@
+open Import
+
+let action_title = "Remove unused"
+
+let slice doc (range : Range.t) =
+  let src = Document.source doc in
+  let (`Offset start) = Msource.get_offset src @@ Position.logical range.start
+  and (`Offset end_) = Msource.get_offset src @@ Position.logical range.end_ in
+  String.sub (Msource.text src) ~pos:start ~len:(end_ - start)
```
rgrinberg, line 9: String.drop?

ocaml-lsp-server/src/code_actions/action_remove_unused.ml
```diff
@@ -0,0 +1,142 @@
...
+
+(* Create a code action that removes [range] and refers to [diagnostic]. *)
+let code_action_remove_range uri doc (diagnostic : Diagnostic.t) range =
+  let edit = edit_single_doc uri doc range "" in
+  CodeAction.create ~diagnostics:[ diagnostic ] ~title:action_title
+    ~kind:CodeActionKind.QuickFix ~edit ~isPreferred:false ()
+
+let code_action_remove_diagnostic_range uri doc (diagnostic : Diagnostic.t) =
+  code_action_remove_range uri doc diagnostic
+    (Range.create ~start:diagnostic.range.start ~end_:diagnostic.range.end_)
```
rgrinberg, line 89: Why not just pass the diagnostic range directly?

rgrinberg (approved):
Looks good. A few small tweaks and it should be good to go.

ulugbekna (changes requested):
The code actions don't work for me since the error messages have the form (after rebasing to the latest master):

image

I guess we could search for a substring but pessimistic about such solution's performance. Wdyt? 

There are a couple more changes that I have in mind before merging the PR, which I'll add tomorrow hopefully

jfeser:
This still needs some changes to pick up the right error messages.

How does ocamllsp decide which diagnostics to show? I have a test project with all warnings turned on, but I only see a subset of them in my editor. When I build with dune I see all of the warnings.

ulugbekna:
I know that ocaml-lsp uses merlin to get the warnings, but I'm not sure how
merlin decides which warnings to show -- it may either get configs from
dune or have only support for some of the warnings

jfeser:
My understanding is that merlin gets configs from dune (through dune ocaml-merlin?). So I'm not sure why it's not picking up my dune file config.

jfeser:
It seems that merlin is in fact seeing my config. However, it doesn't support several of the "unused X" warnings.

Supported: 26, 27
Unsupported: 32-37 

Not sure how hard it would be to get support for 32-37 into merlin.

rgrinberg:
Can you upload the project where you're testing this stuff? AFAIK, merlin doesn't special case any warnings and just reads the warning config from dune.

jfeser:
https://github.com/jfeser/warnings-test

Merlin doesn't special case warnings, but I think they only hook into certain parts of the compiler (https://github.com/ocaml/merlin/issues/343). They also do some filtering (https://github.com/ocaml/merlin/issues/419).

rgrinberg:
In that case what you have is fine. Soon, some diagnostics will come directly from the dune build so your code actions will be picked up from there.

ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
```diff
@@ -879,4 +900,606 @@ let (f, x) = 1 + (f x)
...
+      uri,
+    );
+    let start = Types.Position.create(1, 9);
+    let end = Types.Position.create(1, 10);
+    let context = {
+      diagnostics: [
+        mkUnusedDiagnostic(
+          "value",
+          Types.Position.create(1, 6),
+          Types.Position.create(1, 12),
```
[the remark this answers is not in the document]
jfeser: No, `start` and `end` are the cursor position.

ocaml-lsp-server/src/code_actions/action_remove_unused.ml
jfeser: Doesn't that just drop the start of the string?

jfeser:
That's good to hear.

@ulugbekna Have another look. The code actions should work now.

ocaml-lsp-server/src/code_actions/action_mark_unused.ml
```diff
@@ -0,0 +1,26 @@
+open Import
+
+let action_title = "Mark as unused"
+
+let logfile = open_out "/tmp/ocaml-lsp.log"
+
+let log msg =
+  output_string logfile msg;
+  flush logfile
+
```
ulugbekna, line 10: the logging code should be removed I think

ocaml-lsp-server/src/code_actions/action_mark_unused.ml
```diff
@@ -0,0 +1,26 @@
...
+    ~kind:CodeActionKind.QuickFix ~edit ~isPreferred:false ()
+
+let code_action doc (params : CodeActionParams.t) =
+  List.filter params.context.diagnostics ~f:(fun (d : Diagnostic.t) ->
+      log @@ sprintf "Diag: %s\n" d.message;
+
+      match Position.compare_inclusion params.range.start d.range with
+      | `Outside _ -> false
+      | `Inside -> true)
+  |> List.find ~f:(fun d -> Diagnostic_util.(is_unused_variable d))
```
ulugbekna, line 24: 1. When we `List.filter` and only take one of the filtered values, we do unnecessary work. We could only `List.find`

```ocaml
List.find params.context.diagnostics ~f:(fun (d : Diagnostic.t) ->
     match Position.compare_inclusion params.range.start d.range with
      | `Outside _ -> false
      | `Inside -> Diagnostic_util.is_unused_variable d)
```

The following `Option.map` could also be inlined without much loss of readability imho. 

2. We don't usually do `open`s such as `D_util.(is_unused ...)` -- we `open` only for infix & let operators usually

ocaml-lsp-server/src/code_actions/action_remove_unused.ml
```diff
@@ -0,0 +1,81 @@
...
+
+let enclosing_value_binding_range pipeline pos name =
+  let open Option.O in
+  let typer = Mpipeline.typer_result pipeline in
+  let browse = Mbrowse.of_typedtree (Mtyper.get_typedtree typer) in
+  Mbrowse.enclosing
+    (Mpipeline.get_lexing_pos pipeline @@ Position.logical pos)
+    [ browse ]
+  |> List.find_map ~f:(function
+       | ( _
```
ulugbekna, line 19: Here we could take the second value of the tuple to make pattern matching more readable, ie
`List.find_map ~f:(fun n -> match snd n with ...`

ocaml-lsp-server/src/code_actions/diagnostic_util.ml
```diff
@@ -0,0 +1,5 @@
+open Import
+
+let is_unused_variable (d : Diagnostic.t) =
+  String.is_prefix d.message ~prefix:"Error (warning 26): unused variable"
+  || String.is_prefix d.message ~prefix:"Error (warning 27): unused variable"
```
ulugbekna, line 5: 1. Naming: `is_unused_var_warning` seems more understandable at once to me
2. Matching on prefix with the error code should be enough, ie `Error (warning 26)`

ocaml-lsp-server/src/document.ml
```diff
@@ -217,3 +217,14 @@ let get_impl_intf_counterparts uri =
     | to_switch_to -> to_switch_to
   in
   List.map ~f:Uri.of_path files_to_switch_to
+
+let edit doc range new_text =
```
ulugbekna, line 221: (nonessential)

Wdyt of a more explicit arg type 
```ocaml
new_text: [`Remove | `Replace_with of string] 

(* eg *) 
let _ = Document.edit doc range `Remove
```
?

ulugbekna (changes requested):
Thanks, the code looks much better and works like a charm! :-) 

I think that these two code actions are activated in a very similar setting & should be combined to avoid duplicate work of filtering diagnostics.

Regarding testing, ideally, we would pass code to ocaml-lsp, get diagnostics for the code, and use those real diagnostics to test the code. Thus, we would catch if the diag messages for warnings 26 and 27 are changed. 

I also left some suggestions inline.

ocaml-lsp-server/src/document.ml
jfeser: I probably wouldn't, since `Remove and `Replace_with do the same thing on empty strings anyway.

ulugbekna:
We should better handle patterns. 

1. We should make sure that `mark as unused` for fields in a record have a different edit, ie `{ a; b = _ }`. Currently, we propose an edit that results in an error.

image

We need to inspect the context in which we propose the code action. Ideally, we should do that in the same pattern match as `enclosing_value_binding_range`.

2. Optionally, removing unused record fields also seems plausible: we should check whether the pattern for the record is closed or not, based on that we can simply remove the field or remove the field and put `; _ ` after the last field. This doesn't have to happen in this PR though.

3. Optionally, we could also offer removing unused vars from patterns, eg `(a, b, c)` -> `(a, b)`. More challenging than the rest of course.

---

Importantly, I think we don't need to handle `structure_item`s in `enclosing_value_binding_range`, because the following will not trigger a diagnostic in merlin as of now: 

```
(* a.ml *) 
let a = 1
```

Only on build we get a warning 32 about unused val declaration.

ocaml-lsp-server/src/code_actions/action_mark_remove_unused.ml
```diff
@@ -0,0 +1,144 @@
...
+let code_action_mark_value_unused doc (diagnostic : Diagnostic.t) =
+  let open Option.O in
+  Document.with_pipeline_exn doc (fun pipeline ->
+      let var_name = slice doc diagnostic.range in
+      let pos = diagnostic.range.start in
+      let+ text_edit =
+        let contexts =
+          enclosing_pos pipeline pos
+          |> List.map ~f:(fun (_, x) -> x)
+          |> List.rev
```
rgrinberg, line 80: `List.rev_map`.

PR #505, opened 2021-09-09
Add dune promotion information to diagnostics

rgrinberg:
This will allow vscode to display diffs. I'm not sure if the client should provide a capability that tells us if it can do anything with this information. Of course it can always just ignore this data.

### DIFF
--- a/dune
+++ b/dune
@@ -18,7 +18,7 @@
   (subdir
    dune_filesystem_stubs
    (copy_files#
-     %{workspace_root}/submodules/dune/otherlibs/stdune-unstable/dune_filesystem_stubs/*.*)
+    %{workspace_root}/submodules/dune/otherlibs/stdune-unstable/dune_filesystem_stubs/*.*)
    (library
     (name dune_filesystem_stubs)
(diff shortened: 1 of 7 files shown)


tmattio (approved):
Very nice!

- Should we add a tag to differentiate the promotion diagnostics from the others? This would be useful if we do something specific with them in VSCode (e.g. promotion diff UI). Not sure if this is supported by LSP though.
- This is missing a changelog entry

rgrinberg:
> Should we add a tag to differentiate the promotion diagnostics from the others? This would be useful if we do something specific with them in VSCode (e.g. promotion diff UI). Not sure if this is supported by LSP though.

There's no appropriate tag for this yet.

> This is missing a changelog entry

It's not really visible to end users. When we add the corresponding client side functionality, we can add the CHANGES entry there.

By the way, I am planning to guard this behind a capability. So it will not be visible unless explicitly asked for.

rgrinberg:
Okay, I modified the PR two ways:

* The client must now provide a capability to signal that it is interested in this promotion info
* The server signals that this promotion information is now available.

PR #506, opened 2021-09-13
Update to ocaml 4.13

ulugbekna:


dune-project
```diff
@@ -53,9 +53,9 @@ possible and does not make any assumptions about IO.
   (pp (>= 1.1.2))
   (csexp (>= 1.4))
   (result (>= 1.5))
-  (ocamlformat-rpc-lib (and (>= 0.18.0) (< 0.19.0)))
+  (ocamlformat-rpc-lib (and (>= 0.19.0) (< 0.20.0)))
```
rgrinberg, line 56: Are we not compatible with 0.18 anymore?

ulugbekna: I think so:

image

ulugbekna:
Btw, how do we decide when to vendor a dependency? For example, why don't we vendor `re` or `ocamlformat-rpc-lib`?

dune-project
rgrinberg: It's a little strange that an rpc library is going to be incompatible with a new version of OCaml. I wonder if they are just being conservative.

But even still, we still don't need the constraint. The solver will never select 0.18 for 4.13 anyway because of the constraint we inherit from ocamlformat-rpc-lib. Our constraints should only reflect interactions between our package and the dependencie, and not interactions between dependencies.

rgrinberg:
There's no hard and fast rule. In the beginning, I wanted to vendor to avoid imposing constraints for end users and making sure that ocamllsp would be installable no matter what versions of dependencies they would use. At this point, we can reconsider some of those early decisions. For example, I would think that depending on octavius and omd should be fine. Cmdliner we can get rid of completely and go back to arg - we are never going to have a complex CLI. Uutf should stay vendored as it doesn't use dune and introduces depopts.

voodoos:
@Ulugbek I just found an issue with short paths in Merlin for 4.13 preview. The server hangs in some situations. 
This is probably not trivial so I am just going to deactivate short-paths until we find a solution.

ulugbekna:
The upgrade already happened in another PR, so closing.

PR #507, opened 2021-09-17
Make short-paths configurable with dune

ulugbekna:
- remove forced (*) "short-paths" from merlin config
- update a test -- The main goal of that test is to see whether we stackoverflow on mutually recursive type definitions, which isn't happening. The response on hover changed because we don't enforce short-paths, AFAIU. If we built the testing code with dune, user would still get the desired short-paths behavior (see the note at the end) 

* "forced" in the sense that user should be able to configure
whether they want "short-paths" from dune, but we override that config.

Note: dune sets "short-paths" by default, if not overridden by user.


rgrinberg:
I believe short paths were already configurable by dune even before this PR. The current code does not "force" short paths but simply makes it the default to match dune's default. That is desirable because that is makes new files not yet visible to dune type check the same way. Why should we change the default back to long paths?

ulugbekna:
What you say makes sense. Then the behavior that Ulysse experienced with configuring short-paths using dune is puzzling -- from Slack:

> @Ulugbek @rgrinberg does ocaml-lsp enforce short-paths ?
> The problem arise when both -principal and -short-paths are enabled.
> Disabling -short-paths in my dune configuration will fix the issue when calling Merlin directly but not when using ocaml-lsp as is short-paths was still enabled.

rgrinberg:
I just thought of another way to address this. We can make dune give a proper "default" configuration for sources that don't yet exist. This should also fix the issue with warnings and other flags.

PR #508, opened 2021-09-17
Split ocaml_lsp_server into a lean executable and heavy-lifting library

ulugbekna:


ocaml-lsp-server/bin/dune
```diff
@@ -0,0 +1,5 @@
+(executable
+ (name main)
+ (package ocaml-lsp-server)
+ (public_name ocamllsp)
+ (libraries cmdliner dune-build-info ocaml_lsp_server result))
```
rgrinberg, line 5: I don't see `result` used directly here.

rgrinberg:
Looks ok. The `lib/` directory should be called `src/`.

Also, if at all possible it might make sense to move all the version stuff to the binary and initialize the server with version. This can be in a separate PR though.

rgrinberg (approved):

rgrinberg:
I merged myself to avoid conflicts with upcoming PR's

Issue #504, opened 2021-09-08
Diagnostics are played back slowly if didChange events are coming in faster than diagnostics are produced


copy:
Here's a demo:

https://user-images.githubusercontent.com/1816456/132448167-768090e1-86ba-484e-9403-81f2a1e435f5.mp4

lsp clients (e.g. neovim) send `didChange` events on every character (in order to support completion, and since diagnostics are implicit in lsp). ocaml-lsp seems to queue up these events and sends a diagnostic for each of them. If the user types faster than diagnostics are being produced, these delays add up, causing the effect shown in the video.

ulugbekna:
Could you report 
1. ocaml-lsp version
2. client info - which extension and version
3. your OS

ocaml-lsp implements debouncing which should prevent the thing we see on the video; maybe you have file auto-save ON

copy:
1. I've reproduced this in 1.7.0 and 1.8.0
2. neovim with nvim-lspconfig
3. Linux

The plugin does have an option to debounce text changes sent to the lsp server, however that increases the delay I get a diagnostics (for example, after setting the debounce to one second I need to wait up to one second to get the latest diagnostic) and I believe this could be fixed on the ocaml-lsp side.

> ocaml-lsp implements debouncing which should prevent the thing we see on the video; maybe you have file auto-save ON?

What do you mean by file auto-save? I believe lsp clients send `didChange` events on every keystroke by design (to support completion; and because the lsp client doesn't know if the server is working on diagnostics).

It seems that the lsp server calculates exactly one diagnostic per incoming `didChange` event, so perhaps the debouncing is not working correctly?

rgrinberg:
> so perhaps the debouncing is not working correctly?

Could be.

Try increasing the setting to see if it does anything https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/src/configuration.ml#L5

copy:
> Try increasing the setting to see if it does anything https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/src/configuration.ml#L5

Yes, setting it to a higher value fixes the issue, and setting it to 0 makes the issue worse. That said, I prefer the setting of 0 since it reduces the latency before getting diagnostics (if it weren't for the queueing up of diagnostics). Or, in other words, it would be nice if ocaml-lsp could send diagnostics as fast as possible (one at a time), but collapse multiple change events instead of queueing them up.

rgrinberg:
[...]

PR #509, opened 2021-09-17
Doc update performance

rgrinberg:
@copy ~~I did not touch the debouncing yet, but I did hopefully speed up the updates~~ I did address one issue with the debouncing. Let me know if this is any better for you.

### DIFF
--- a/lsp/src/text_document.ml
+++ b/lsp/src/text_document.ml
@@ -1,57 +1,61 @@
 open Types
+module String = StringLabels
 
-module Encoding = struct
-  let recode ?nln ?encoding out_encoding
-      (src : [ `Channel of in_channel | `String of string ])
-      (dst : [ `Channel of out_channel | `Buffer of Buffer.t ]) =
-    let rec loop d e =
-      match Uutf.decode d with
+exception Invalid_utf8
+
+let find_offset ~utf8 ~utf16_range:range =
+  let dec =
+    Uutf.decoder
+      ~nln:(`ASCII (Uchar.of_char '\n'))
+      ~encoding:`UTF_8 (`String utf8)
+  in
+  let utf16_codepoint_size = 4 in
+  let utf16_codepoints_buf = Bytes.create utf16_codepoint_size in
+  let enc = Uutf.encoder `UTF_16LE `Manual in
+  let rec find_char line char =
+    if char = 0 || Uutf.decoder_line dec = line then
+      Uutf.decoder_byte_count dec
+    else
+      match Uutf.decode dec with
+      | `Await -> raise Invalid_utf8
+      | `End -> Uutf.decoder_byte_count dec
+      | `Malformed _ ->
+        invalid_arg "Text_document.find_offset: utf8 string is malformed"
       | `Uchar _ as u ->
-        ignore (Uutf.encode e u);
-        loop d e
-      | `End -> ignore (Uutf.encode e `End)
-      | `Malformed _ ->
-        ignore (Uutf.encode e (`Uchar Uutf.u_rep));
-        loop d e
-      | `Await -> assert false
-    in
-    let d = Uutf.decoder ?nln ?encoding src in
-    let e = Uutf.encoder out_encoding dst in
-    loop d e
-
-  let nln = `ASCII (Uchar.of_char '\n')
-
-  let reencode_string in_enc out_enc str =
-    let buf = Buffer.create (String.length str) in
-    let () = recode ~nln ~encoding:in_enc out_enc (`String str) (`Buffer buf) in
-    Buffer.contents buf
-
-  let utf8_to_utf16 = reencode_string `UTF_8 `UTF_16LE
-
-  let utf16_to_utf8 = reencode_string `UTF_16LE `UTF_8
-
-  let utf16_line_offsets (text : string) =
-    let rec loop d acc =
-      let old_line = Uutf.decoder_line d in
-      match Uutf.decode d with
-      | `Uchar _
-      | `Malformed _ ->
-        let new_line = Uutf.decoder_line d in
-        if new_line > old_line then
-          let line_ofs = Uutf.decoder_byte_count d / 2 in
-          (* UTF16 encodes on 2 bytes *)
-          loop d (line_ofs :: acc)
-        else
-          loop d acc
-      | `End ->
-        let end_ofs = Uutf.decoder_byte_count d / 2 in
-        Array.of_list (List.rev (end_ofs :: acc))
-      | `Await -> assert false
-    in
-    let encoding = `UTF_16LE in
-    let decoder = Uutf.decoder ~nln ~encoding (`String text) in
-    loop decoder [ 0 ]
-end
+        Uutf.Manual.dst enc utf16_codepoints_buf 0 utf16_codepoint_size;
+        (match Uutf.encode enc u with
+        | `Partial ->
+          (* we always have space for one character *)
+          assert false
+        | `Ok -> ());
+        let char =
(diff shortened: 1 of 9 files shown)


rgrinberg:
My bad, this PR isn't quite correct. It's wrong to make Text_document mutable in the presence of threads. Will fix.

rgrinberg:
Should be good now.

rgrinberg:
I bit the bullet and just removed all the unecessary unicode conversions.

copy:
Can confirm this is significantly faster and fixes #504, thanks.

rgrinberg:
Good. Thanks for helping me diagnose. I'll make a point release once this PR is merged.

ocaml-lsp-server/src/document.ml
```diff
@@ -142,37 +145,43 @@ let make_config uri =
...
-    List.fold_left changes ~init:doc.tdoc ~f:(fun acc change ->
+let update_text ?version ({ merlin; tdoc; _ } as t) changes =
+  match
+    List.fold_left changes ~init:tdoc ~f:(fun acc change ->
         Text_document.apply_content_change ?version acc change)
-  in
-  let open Fiber.O in
-  let+ pipeline = make_pipeline doc.merlin tdoc in
-  { doc with tdoc; pipeline }
+  with
```
ulugbekna, line 173: This is interesting: I wonder whether we could save up on some computations sharing information across change applications. The k-th change is based on the state of document with (k-1)-th change applied, so it's not very obvious to me now.

ocaml-lsp-server/src/document.ml
```diff
@@ -142,37 +145,43 @@ let make_config uri =
...
-          make_config uri
+  ( lazy (Msource.make (Text_document.text tdoc))
+  , Lazy_fiber.create (fun () ->
+        let async_make_pipeline =
+          Scheduler.async_exn thread (fun () ->
+              let config =
+                let uri = Text_document.documentUri tdoc in
+                make_config uri
+              in
+              Mpipeline.make config (Msource.make (Text_document.text tdoc)))
```
ulugbekna, line 156: `Msource.make (Text_document.text tdoc)` is computed twice - lines 148 and 156, is it intentional?

ocaml-lsp-server/src/document.ml
```diff
@@ -142,37 +145,43 @@ let make_config uri =
...
-  let open Fiber.O in
-  let+ res = await async_make_pipeline in
-  match res with
-  | Ok s -> s
-  | Error e -> Exn_with_backtrace.reraise e
+        let open Fiber.O in
+        let+ res = await async_make_pipeline in
+        match res with
+        | Ok s -> s
+        | Error e -> Exn_with_backtrace.reraise e) )
```
ulugbekna, line 162: seems like we could use `await_exn`

ulugbekna (commented):
Looks good to me. I left some inline comments, most of which are minor. 

It would be great if we could postpone the release for a day or two until I fix quite a disruptive bug in "jump to a hole on destruct". 

Quite an interesting PR to read, btw, thanks! :-)

ocaml-lsp-server/src/document.ml
rgrinberg: Probably the easiest thing to do that can speed this up is allocating the final string only once after `k` consecutive changes without a read. Right now we are creating a lot of GC pressure by throwing away a bunch of intermediate strings.

lsp/src/text_document.ml
```diff
@@ -1,114 +1,89 @@
...
-    loop decoder [ 0 ]
-end
+let find_offset ~utf8 ~utf16_range:range =
+  let dec =
+    Uutf.decoder
+      ~nln:(`ASCII (Uchar.of_char '\n'))
+      ~encoding:`UTF_8 (`String utf8)
+  in
+  let bytes = Bytes.create 4 in
+  let enc = Uutf.encoder `UTF_16LE `Manual in
```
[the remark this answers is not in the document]
rgrinberg: It is the endianness used by MS Windows. LSP just inherits it.

ocaml-lsp-server/src/document.ml
rgrinberg: It was intentional, but it's actually needlessly complicated. The fact that the earlier `Msource.make` was lazy made it impossible to share across threads. I removed the pointless lazy and now it's possible to share it.

lsp/src/text_document.ml
```diff
@@ -1,114 +1,89 @@
...
+    let start_offset, end_offset = find_offset ~utf8:t.text ~utf16_range in
+    let text =
+      String.concat ~sep:""
+        [ String.sub t.text ~pos:0 ~len:start_offset
+        ; change.text
+        ; String.sub t.text ~pos:end_offset
+            ~len:(String.length t.text - end_offset)
+        ]
+    in
+    { t with text }
```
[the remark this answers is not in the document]
rgrinberg: Fixed. Thanks.

This reminds me that the way we update multiple changes is wrong. We should be incrementing on every single change and not just setting the final value once.

lsp/src/text_document.ml
```diff
@@ -1,114 +1,89 @@
...
-    let decoder = Uutf.decoder ~nln ~encoding (`String text) in
-    loop decoder [ 0 ]
-end
+let find_offset ~utf8 ~utf16_range:range =
+  let dec =
+    Uutf.decoder
+      ~nln:(`ASCII (Uchar.of_char '\n'))
+      ~encoding:`UTF_8 (`String utf8)
+  in
+  let bytes = Bytes.create 4 in
```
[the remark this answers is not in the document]
rgrinberg: done.

lsp/src/text_document.ml
```diff
@@ -1,114 +1,89 @@
...
+      match Uutf.decode dec with
+      | `Await -> raise Invalid_utf8
+      | `End -> Uutf.decoder_byte_count dec
+      | `Malformed _ -> assert false
+      | `Uchar _ as u ->
+        Uutf.Manual.dst enc bytes 0 4;
+        (match Uutf.encode enc u with
+        | `Partial -> assert false
+        | `Ok -> ());
+        let char = char - ((4 - Uutf.Manual.dst_rem enc) / 2) in
```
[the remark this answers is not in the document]
rgrinberg: I added a temp var that should clarify. Feel free to tweak it further.

lsp/src/text_document.ml
```diff
@@ -1,114 +1,89 @@
...
+  let bytes = Bytes.create 4 in
+  let enc = Uutf.encoder `UTF_16LE `Manual in
+  let rec find_char line char =
+    if char = 0 || Uutf.decoder_line dec = line then
+      Uutf.decoder_byte_count dec
+    else
+      match Uutf.decode dec with
+      | `Await -> raise Invalid_utf8
+      | `End -> Uutf.decoder_byte_count dec
+      | `Malformed _ -> assert false
```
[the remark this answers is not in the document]
rgrinberg: We assert false because we assume that the current state of the text document is a valid utf8 string. If it was invalid, we should have raised when it became invalid and not here.

ocaml-lsp-server/src/document.ml
rgrinberg: await_exn isn't defined in this module and I hate to introduce it. I don't like these little helpers that don't encapsulate any logic.

rgrinberg:
> It would be great if we could postpone the release for a day or two until I fix quite a disruptive bug in "jump to a hole on destruct".

Sure, a few days is fine. But note that minimizing the number of point releases is a non goal. We can make a release on every bug fix for all I care.

PR #510, opened 2021-09-19
Improve completion kinds

rgrinberg:
Introduce these mappings:

label -> field
variant -> enum member

For items without precise kinds, we no longer select a related but
inaccurate kind.

### DIFF
--- a/ocaml-lsp-server/src/compl.ml
+++ b/ocaml-lsp-server/src/compl.ml
@@ -17,15 +17,14 @@ end
 let completion_kind kind : CompletionItemKind.t option =
   match kind with
   | `Value -> Some Value
-  | `Constructor -> Some Constructor
-  | `Variant -> None
-  | `Label -> Some Property
-  | `Module
-  | `Modtype ->
-    Some Module
-  | `Type -> Some TypeParameter
+  | `Variant -> Some EnumMember
+  | `Label -> Some Field
+  | `Module -> Some Module
+  | `Modtype -> Some Interface
   | `MethodCall -> Some Method
   | `Keyword -> Some Keyword
+  | `Constructor -> Some Constructor
+  | `Type -> Some TypeParameter
 
 let prefix_of_position ~short_path source position =
   match Msource.text source with
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+- More accurate completion kinds.
+  New completion kinds for variants and fields. Removed inaccurate completion
+  kinds for constructors and types. (#510)
+
 # 1.8.2 (09/14/2021)
 
 - Disable experimental dune support. It was accidentally left enabled.


ocaml-lsp-server/src/compl.ml
```diff
@@ -17,15 +17,16 @@ end
...
+  | `Label -> Some Field
   | `Module
   | `Modtype ->
     Some Module
-  | `Type -> Some TypeParameter
   | `MethodCall -> Some Method
   | `Keyword -> Some Keyword
+  | `Constructor
+  | `Type ->
+    None
```
ulugbekna, line 29: I would be more in favor of the mapping below. On vscode, having TypeParameter kind for types helps to understand that the completion item is a type, so removing it is less desirable in my opinion, even if the mapping isn't ideal. Module type could fit into `Interface` 1) to diff from a `Module` 2) because it's semantically close to an interface as well. Constructors are enum members for sure but could be marked as `Enum` to be different from Variants. 

```ocaml
  match kind with
  | `Constructor -> Some EnumMember (* could also be Enum to diff from [`Variant] *)
  | `Keyword -> Some Keyword
  | `Label -> Some Variable
  | `MethodCall -> Some Method
  | `Modtype -> Some Interface
  | `Module -> Some Module
  | `Type -> Some TypeParameter
  | `Value -> Some Value
  | `Variant -> Some EnumMember
```

rgrinberg: Okay, I'll revert the approximate descriptions and use some of your suggestions after all. I don't care  too much but I do think that new users can easily be confused by this.

> having TypeParameter kind for types helps to understand that the completion item is a type, so removing it is less desirable in my opinion, even if the mapping isn't ideal

What are we going to do once we can complete actual type parameters?

> Module type could fit into Interface 1) to diff from a Module 2) because it's semantically close to an interface as well

What about object types and class types? Those are semantically close as well.

Why did you change Labels to Variable? Field certainly seems more correct to me.

ulugbekna: > What are we going to do once we can complete actual type parameters?

I'd make that decision when we can actually complete them, but I'd assume that the tick in the type param should be good of a signal to differentiate a type from a type param, so we can use the same for both, at least until we have a separate completion kind for types.

> What about object types and class types? Those are semantically close as well.

I'm not sure, but since classes and objects are rare, we should prioritize having nicer support for modules and module types IMHO. 

> Why did you change Labels to Variable? Field certainly seems more correct to me.

I'd say a "field" relates to either a field in a struct/record/module/object. I'm not sure why a label is a field. But that's minor, we can keep `field` for labels.
--

Also, I'm not sure why `Constructor is mapped to "Constructor" because the former is a value constructor, AFAIU, and should map to `EnumMember`, while the latter is a class constructor. But again, shouldn't be a problem.

rgrinberg: Labels are in compiler parlance are record fields.

PR #512, opened 2021-09-21
Fix completion for symbols passed as named arg

ulugbekna:
`prefix_of_position` is getting complicated; I hope we're not breaking anything with this fix

### DIFF
--- a/ocaml-lsp-server/src/compl.ml
+++ b/ocaml-lsp-server/src/compl.ml
@@ -31,7 +31,7 @@ let completion_kind kind : CompletionItemKind.t option =
 let prefix_of_position ~short_path source position =
   match Msource.text source with
   | "" -> ""
-  | text -> (
+  | text ->
     let from =
       let (`Offset index) = Msource.get_offset source position in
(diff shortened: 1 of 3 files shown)


ocaml-lsp-server/src/compl.ml
```diff
@@ -87,7 +88,14 @@ let prefix_of_position ~short_path source position =
       | Some pos -> pos + 1
     in
     let len = from - pos + 1 in
-    String.sub text ~pos ~len
+    let reconstructed_prefix = String.sub text ~pos ~len in
+    (* if we reconstructed [~f:ignore], we should take only [ignore], so: *)
+    if not (String.is_prefix reconstructed_prefix ~prefix:"~") then
```
rgrinberg, line 93: Without looking at this very deeply, why doesn't the problem exist for optional arguments?

ulugbekna: Excellent remark. Fixed handling for optional args as well.

rgrinberg (approved):

PR #515, opened 2021-09-24
Use %{project_root} instead of %{workspace_root}

nojb:
Makes it more composable.


rgrinberg:
Indeed it does. Thanks

PR #523, opened 2021-10-15
feature: move to our own merlin config reading

rgrinberg:
This fixes issues with merlin running its own processes. In the future,
we'll try and re-use dune rpc instead of running yet another binary.

cc @voodoos

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Unreleased
 
+## Breaking Change
+
+- ocamllsp drops support for `.merlin` files, and as a consequence no longer
+  depends on dot-merlin-reader. (#523)
+
 ## Features
 
 - New code action to automatically remove values, types, opens (#502)
--- a/dune-project
+++ b/dune-project
@@ -49,7 +49,6 @@ possible and does not make any assumptions about IO.
   (re (>= 1.5.0))
   (ppx_yojson_conv_lib (>= "v0.14"))
   dune-build-info
-  dot-merlin-reader
   (pp (>= 1.1.2))
   (csexp (>= 1.5))
   (result (>= 1.5))
--- a/ocaml-lsp-server.opam
+++ b/ocaml-lsp-server.opam
@@ -24,7 +24,6 @@ depends: [
   "re" {>= "1.5.0"}
   "ppx_yojson_conv_lib" {>= "v0.14"}
   "dune-build-info"
-  "dot-merlin-reader"
   "pp" {>= "1.1.2"}
   "csexp" {>= "1.5"}
   "result" {>= "1.5"}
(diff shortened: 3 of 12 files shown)


voodoos:
(I did not look at the PR in depth)
You bypass Merlin spinning up `dot-merlin-reader` by not calling `Mconfig.get_external_config`, that seems a reasonable thing to do but what was the initial issue that led you to do this change ?

I also agree that the next obvious step is to use Dune RPC to get the configuration. 

However would that mean that the user would have to have Dune's watch mode running in order to get configuration ?

gares:
Hello, I just found out about this while working with @clodyhug on a project which does not use dune for building.
Is there a way to convert by hand a working .merlin file into the (usually dune generated) metadata the lsp server reads?

rgrinberg:
Ocamllsp now reads the metadata by directly querying dune. So there's no way to provide it manually at all.

One workaround is to write dune files for development only. This is what the OCaml compiler does to provide merlin/lsp support for example. It's a bit more boilerplate, but at least your lsp configuration is guaranteed to be correct.

gares:
You also need to build with dune, otherwise the cmi are not in the right place.

Assuming everything is built with dune is kind of harsh. At least the error should be clear: no dune file, no intellisense.

rgrinberg:
Granted, it's at little harsh. The old system had its costs in usability though. I'd get numerous issues from people misconfiguring lsp with stray .merlin files and the additional dependency isn't great either. In the future, I'd like to rely on dune for even more functionality so I just thought we should bite the bullet now.

gares:
btw, the error I was getting wan even harder to understand, like a missing -rectypes... go figure.

voodoos:
@gares it is not a very satisfying solution, but you might be able to work with the [deprecated OCaml extension](https://marketplace.visualstudio.com/items?itemName=hackwaly.ocaml) from @hackwaly. It still works quite well with recent versions of Merlin (I am using it when performing upgrades to newer version of OCaml that are thus not supported by ocaml-lsp yet).

And because it uses vanilla Merlin and Merlin keeps reading `.merlin` files when they are present it should work out of the box. (But other issues might appear since the development of this extension stopped).

(For it to work with my 4.14 version of Merlin I have to write the absolute path to the precise binary `/home/tom/.opam/414-shapes-opt/bin/ocamlmerlin` in the settings and reload the window.)

maxkorp:
Small note since I'm looking at a partial reversion, the commit listed here seems to have gone poof from the repo history, the new sha for where it ended up is ebc0032c44e84eefd9ed809d064d6bfac0b51d7b

https://github.com/ocaml/ocaml-lsp/commit/ebc0032c44e84eefd9ed809d064d6bfac0b51d7b

PR #539, opened 2021-10-21
refactor: rename fmt to ocamlformat

rgrinberg:


rgrinberg:
Yes, we'll be deprecating refmt support eventually. Your suggestion sounds good, will give it a try.

PR #540, opened 2021-10-21
feature: format dune files

rgrinberg:
Run the dune formatter through lsp. @ulugbekna @mnxn we need to co-ordinate disabling formatting on vscode for these files.

### DIFF
--- a/ocaml-lsp-server/src/dune.ml
+++ b/ocaml-lsp-server/src/dune.ml
@@ -153,6 +153,8 @@ type config =
 module Instance : sig
   type t
 
+  val format_dune_file : t -> Document.t -> string Fiber.t
+
   val stop : t -> unit Fiber.t
 
   val run : t -> unit Fiber.t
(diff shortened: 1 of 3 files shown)


ulugbekna:
Is this ready for review?

I wonder how the interaction among the vscode extension, ocaml-lsp, and dune should be like. Should ocaml-lsp dynamically register a capability for formatting dune files, then the dune formatter in the extension registered on startup would be disposed? This interaction seems complicated because:

1. the extension needs to make sure ocaml-lsp can format dune files
2. ocaml-lsp needs to make sure the installed version of dune supports rpc

is that right?

rgrinberg:
Handling 1. and 2. would be nice, but keep in mind that we're only going to have a first class experience for dune 3.0 and above. We don't want to spend our precious time worrying about old versions of dune and lsp. This is after all the entire reason we put so much effort into making it easy to upgrade dune. 

Therefore,  vscode should just drop the ability to format dune files if we see the server is recent enough to handle it. Now the question remains on how to best signal this capability to the clients. We can do it dynamically as you've suggested, but we can also just a plain old custom server capability if that's easier.

PR #543, opened 2021-10-21
feature: dynamically register for cram/dune

rgrinberg:


ulugbekna:
I'll review this tomorrow, sorry for the wait.

ulugbekna:
`Ocaml_lsp_server.on_notification` needs to handle notifications based on the file syntax; otherwise, we get 

```
[Trace - 11:28:59 AM] Sending notification 'textDocument/didOpen'.
Params: {
    "textDocument": {
        "uri": "file:///Users/ulugbekna/code/ocaml-lsp/lsp-fiber/src/dune",
        "languageId": "dune",
        "version": 1,
        "text": "(library\n (name lsp_fiber)\n (libraries\n  dyn\n  fiber\n  fiber_unix\n  jsonrpc\n  jsonrpc_fiber\n  lsp\n  ppx_yojson_conv_lib\n  result\n  stdune\n  yojson))\n"
    }
}


Uncaught error when handling notification:
{
  "params": {
    "textDocument": {
      "uri": "file:///Users/ulugbekna/code/ocaml-lsp/lsp-fiber/src/dune",
      "languageId": "dune",
      "version": 1,
      "text":
        "(library\n (name lsp_fiber)\n (libraries\n  dyn\n  fiber\n  fiber_unix\n  jsonrpc\n  jsonrpc_fiber\n  lsp\n  ppx_yojson_conv_lib\n  result\n  stdune\n  yojson))\n"
    }
  },
  "method": "textDocument/didOpen",
  "jsonrpc": "2.0"
}
Error:
[ { exn =
      "jsonrpc response error {\n\
      \  \"data\": { \"language_id\": \"dune\" },\n\
      \  \"code\": -32600,\n\
      \  \"message\": \"invalid language ID\"\n\
       }"
  ; backtrace =
      "Raised at Jsonrpc.Response.Error.raise in file \"jsonrpc/src/jsonrpc.ml\", line 200, characters 18-29\n\
       Called from Ocaml_lsp_server.set_diagnostics in file \"ocaml-lsp-server/src/ocaml_lsp_server.ml\", line 180, characters 8-27\n\
       Called from Ocaml_lsp_server.on_notification in file \"ocaml-lsp-server/src/ocaml_lsp_server.ml\", line 1032, characters 14-40\n\
       Called from Fiber.Execution_context.run_jobs in file \"submodules/dune/src/fiber/fiber.ml\", line 204, characters 8-13\n\
       "
  }
]
```

ulugbekna:
Also, 

1. why do we register for `didOpen` notification only for those files? 
2. do I understand correctly that we don't register statically (I have https://github.com/ocamllabs/vscode-ocaml-platform/blob/7870a8ecaaac8a81bd59764956880f54e6e095e6/src/extension_instance.ml#L20-L27 in mind) because we don't want to get requests/notifications beyond specific ones? 
3. do you happen to know whether just specifying `language` in `DocumentFilter` for dune and cram files enough to support those files in clients besides vscode? I know that `dune`, etc. are file IDs specified on the vscode extension side, but not sure whether they'd be supported in other editors.

---

@pitagha is planning to add support for `textDocument/completion` in `.ocamlformat` files, so I wonder whether we should dynamically register `textDocument/completion` for ocamlformat? (in a separate PR of course)

Also, our vscode extension defines a file ID "ocaml.ocamlformat" [see](https://github.com/ocamllabs/vscode-ocaml-platform/blob/7870a8ecaaac8a81bd59764956880f54e6e095e6/package.json#L720-L730), which we could use to specify documents for which we want to register `textDocument/completion` request, but I guess we should provide a `DocumentFilter` for .ocamlformat files on the ocaml-lsp side as well, so that they would work correctly for all clients, which do not have `ocaml.ocamlformat` file IDs as in vscode extension.

rgrinberg:
1. I forgot to register the other handlers.
2. The reason is actually because I don't want to change clients like vscode. It's a pain to synchronize this information between the two projects. With dynamic registration, the logic lives in one place.
3. It's not enough. I'm beginning to think we shouldn't rely on languageId identifiers. They simply cannot be relied to be consistent across editors.

>  so I wonder whether we should dynamically register textDocument/completion for ocamlformat? (in a separate PR of course)

We should use dynamic registration whenever it's an option. For the reason I listed in 2.

> but I guess we should provide a DocumentFilter for .ocamlformat files on the ocaml-lsp side as well, so that they would work correctly for all clients, which do not have ocaml.ocamlformat file IDs as in vscode extension.

Precisely

rgrinberg:
For neovim, the table is maintained here:

https://github.com/mjlbach/nvim-lspconfig/blob/55dc63e884c564db9898a9f590ef52f6e659093f/lua/lspconfig/ocamllsp.lua#L4

But note that this isn't even the only lsp client just for this editor.

Issue #495, opened 2021-08-25
Better ocamlformat-rpc intergration


ulugbekna:
Currently, we format code in the hover, we should also be able to 

- [x] format source code (currently, we call `ocamlformat` binary to format source files, which can be avoided if ocamlformat-rpc is installed)
- [x] format generated code:
  - [x] destructed code
  - [x] generated intf code

Other suggestions?

rgrinberg:
What about replacing using ocamlformat-rpc for user format queries when it's available? Currently we still run the binary

panglesd:
I opened PR #549 to fix this issue :)

rgrinberg:
@panglesd I assigned this ticket to you.

panglesd:
Yes, formatting destructed code and generated interface is working when #599 is merged.

PR #549, opened 2021-11-15
Better ocamlformatrpc integration

panglesd:
This PR addresses issue #495.

- In case of generated code via destruct or "inferred interface", ocaml-lsp now tries to format them using ocamlformat-rpc. If this fails, the non-formatted text is used. (When there are "merlin holes" no formatting is possible untill the merged https://github.com/ocaml-ppx/ocamlformat/pull/1698 is released).
- In case of user querying formatting the whole file, ocaml-lsp first tries ocamlformat-rpc and fallback to regular ocamlformat if it fails.

### DIFF
--- a/ocaml-lsp-server/src/code_actions/action_destruct.ml
+++ b/ocaml-lsp-server/src/code_actions/action_destruct.ml
@@ -3,6 +3,8 @@ open Fiber.O
 
 let action_kind = "destruct"
 
+let kind = CodeActionKind.Other action_kind
+
 let code_action_of_case_analysis doc uri (loc, newText) =
   let range : Range.t = Range.of_loc loc in
   let edit : WorkspaceEdit.t =
@@ -24,7 +26,7 @@ let code_action_of_case_analysis doc uri (loc, newText) =
   CodeAction.create ~title ~kind:(CodeActionKind.Other action_kind) ~edit
     ~command ~isPreferred:false ()
 
-let code_action doc (params : CodeActionParams.t) =
+let code_action (state : State.t) doc (params : CodeActionParams.t) =
   let uri = params.textDocument.uri in
   match Document.kind doc with
   | Intf -> Fiber.return None
@@ -34,9 +36,15 @@ let code_action doc (params : CodeActionParams.t) =
       let finish = Position.logical params.range.end_ in
       Query_protocol.Case_analysis (start, finish)
     in
-    let+ res = Document.dispatch doc command in
+    let* res = Document.dispatch doc command in
     match res with
-    | Ok res -> Some (code_action_of_case_analysis doc uri res)
+    | Ok (loc, newText) -> (
+      let+ formattedText =
+        Ocamlformat_rpc.(format_type state.ocamlformat_rpc ~typ:newText)
+      in
+      match formattedText with
+      | Ok formattedText -> Some (code_action_of_case_analysis doc uri (loc, formattedText))
+      | Error _ -> Some (code_action_of_case_analysis doc uri (loc, newText)))
     | Error
         { exn =
             ( Merlin_analysis.Destruct.Wrong_parent _ | Query_commands.No_nodes
@@ -45,8 +53,7 @@ let code_action doc (params : CodeActionParams.t) =
             | Merlin_analysis.Destruct.Nothing_to_do )
         ; backtrace = _
         } ->
-      None
+      Fiber.return None
     | Error exn -> Exn_with_backtrace.reraise exn)
 
-let t =
-  { Code_action.kind = CodeActionKind.Other action_kind; run = code_action }
+let t state = { Code_action.kind; run = code_action state }
--- a/ocaml-lsp-server/src/code_actions/action_destruct.ml
+++ b/ocaml-lsp-server/src/code_actions/action_destruct.ml
@@ -39,11 +39,11 @@ let code_action (state : State.t) doc (params : CodeActionParams.t) =
     let* res = Document.dispatch doc command in
     match res with
     | Ok (loc, newText) -> (
-      let+ formattedText =
-        Ocamlformat_rpc.(format_type state.ocamlformat_rpc ~typ:newText)
+      let+ formatted_text =
+        Ocamlformat_rpc.format_type state.ocamlformat_rpc ~typ:newText
       in
-      match formattedText with
-      | Ok formattedText -> Some (code_action_of_case_analysis doc uri (loc, formattedText))
+      match formatted_text with
+      | Ok formatted_text -> Some (code_action_of_case_analysis doc uri (loc, formatted_text))
       | Error _ -> Some (code_action_of_case_analysis doc uri (loc, newText)))
     | Error
         { exn =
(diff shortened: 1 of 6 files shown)


rgrinberg:
The way this works is by sending formatting snippets, but the end result isn't necessarily formatted. I'm curious if you tested this out, does it seem to work well in practice?

ocaml-lsp-server/src/code_actions/action_destruct.ml
```diff
@@ -34,9 +36,15 @@ let code_action doc (params : CodeActionParams.t) =
...
     in
-    let+ res = Document.dispatch doc command in
+    let* res = Document.dispatch doc command in
     match res with
-    | Ok res -> Some (code_action_of_case_analysis doc uri res)
+    | Ok (loc, newText) -> (
+      let+ formattedText =
+        Ocamlformat_rpc.(format_type state.ocamlformat_rpc ~typ:newText)
+      in
+      match formattedText with
```
rgrinberg, line 45: `formatted_text`

ocaml-lsp-server/src/code_actions/action_destruct.ml
```diff
@@ -34,9 +36,15 @@ let code_action doc (params : CodeActionParams.t) =
       let finish = Position.logical params.range.end_ in
       Query_protocol.Case_analysis (start, finish)
     in
-    let+ res = Document.dispatch doc command in
+    let* res = Document.dispatch doc command in
     match res with
-    | Ok res -> Some (code_action_of_case_analysis doc uri res)
+    | Ok (loc, newText) -> (
+      let+ formattedText =
+        Ocamlformat_rpc.(format_type state.ocamlformat_rpc ~typ:newText)
```
rgrinberg, line 43: Why do you need the module open? I think `Ocamlformat_rpc.format_type` should be enough.

ocaml-lsp-server/src/code_actions/action_inferred_intf.ml
```diff
@@ -23,9 +23,15 @@ let code_action_of_intf doc intf range =
 let code_action (state : State.t) doc (params : CodeActionParams.t) =
   match Document.kind doc with
   | Impl -> Fiber.return None
-  | Intf ->
-    let+ intf = Inference.infer_intf ~force_open_impl:true state doc in
-    Some (code_action_of_intf doc intf params.range)
+  | Intf -> (
+    let* intf = Inference.infer_intf ~force_open_impl:true state doc in
+    let+ formattedIntf =
```
rgrinberg, line 28: `formatted_intf`

ocaml-lsp-server/src/code_actions/action_destruct.ml
panglesd: I'll update my variable names. Should `newText` also be renamed to `new_text`, or has this variable a reason to be formatted in camelCase?

ocaml-lsp-server/src/code_actions/action_destruct.ml
panglesd: You are right, it is enough. I don't remember why I opened it.

panglesd:
> The way this works is by sending formatting snippets, but the end result isn't necessarily formatted. I'm curious if you tested this out, does it seem to work well in practice?

It works, I would say, better than without formatting, even though a subsequent call to ocamlformat for the whole file is needed to have the right formatting. However, I don't think automatically formatting the whole file is good for user experience, as it modifies the file at position far from the cursor without telling the user.

Peek 2021-11-17 11-39

As a comparison, emacs does the same (formatting the generated code only), but is better just in the fact that it indent well the generated lines (but I think this is thanks to tuareg-mode and not merlin-related).

ulugbekna:
I have a feeling that the formatting using the OCamlFormat RPC doesn't take into account `.ocamlformat`. I pinned ocamlformat master and formatting works despite my `.ocamlformat` version requiring 0.19.0.

ulugbekna:
Just to be sure, it seems that when using ocamlformat rpc the version put in `.ocamlformat` is not respected, eg see the video below where I have ocamlformat master pinned, and formatting works with `.ocamlformat` `version=0.19.0` (but sometimes it doesn't -- I don't quite understand; it would be nice to make sure the rpc respects the .ocamlformat options, for example, as a test)

https://user-images.githubusercontent.com/16353531/142423565-5fcdf099-456d-47b6-8135-d0d77dabca6e.mov

ocaml-lsp-server/src/code_actions/action_destruct.ml
ulugbekna: camelCase naming is fine if used later for named argument punning, I believe

rgrinberg:
Merged after a few tweaks.

ulugbekna:
AFAIU, `Ocamlformat_rpc.configure` fn, which is called in `Ocamlformat_rpc.(create : unit -> t)`, specifies formatting config tailored to suit "vscode popups", which isn't what we want when we format source files (we want to get the config from the `.ocamlformat` file). Shall we revert and let @panglesd continue working on the PR or have a follow up PR?

rgrinberg:
panglsed can keep working on it in a separate PR. Even without the configuration, it's still an improvement over no formatting at all.

Rudi.

ulugbekna:
I think you're missing that this PR formats source files as well, and the current master is pretty much unusable since any source file reformatted incorrectly on save (if they have formatting on save on). @panglesd do you have time to fix that?

panglesd:
Sure, I'll try to fix this soon (I hope by tomorrow evening)

rgrinberg:
We'll give you the week :)

panglesd:
I am not anymore sure that we should format the entire file with `ocamlformat_rpc`. 

As `ocamlformat_rpc` acts completely independently from the `.ocamlformat` file, we need to find and parse this file, to send to the ocamlformat_rpc server all the options of `.ocamlformat`.
The problem is that `ocamlformat_rpc` does not have a counterpart to the "version" value in the .ocamlformat file: To fail on wrong versions, we would also need to parse the file to find the required version, call `ocamlformat-rpc --version` to know the installed version, compare the two and fail if they differ.

I start to think that there are more disadvantages in using ocamlformat_rpc instead of calling ocamlformat directly, as we need to reinvent the wheel for two reasons: sending `.ocamlformat` options, and compare `.ocamlformat` version with installed version. Thats' a lot! 

What do you think? Should I revert back to `ocamlformat` for formatting whole file? Or should continue the ocamlformat_rpc way? (I have something working except for the version comparison)

rgrinberg:
IMO this is a bug in ocamlformat_rpc's server. I don't see how it makes sense for their rpc server to just ignore the .ocamlformat file. It's clearly an implementation detail when using the binary, so why is using the protocol any different?

Rudi.

ulugbekna:
That indeed seems limitation on the ocamlformat-rpc side, which should be fixed. 

I don't know how much we would like to handle `.ocamlformat` files ourselves, but 
1) parsing ocamlformat files is easy
2) we anyway would like to add support for auto-completion in ocamlformat files
3) we could add various features based on handling ocamlformat files, e.g., on ocamlformat file change propose to reformat all files (preferably with debouncing however)

rgrinberg:
I don't mind adding these features but I want all the heavy lifting to be on the ocamlformat side through rpc. It's their format and should the interpretation of these files should be a black box to us as much as possible.

> on ocamlformat file change propose to reformat all files (preferably with debouncing however)

That's something that should be done through dune.

panglesd:
> parsing ocamlformat files is easy
Sure, but I agree with Rudi that its their format so parsing should be done on their side.

I will make a PR reverting file formatting to using ocamlformat, and file an issue to the ocamlformat repo about using `.ocamlformat` for the rpc server.

rgrinberg:
Instead of reverting you can also add a flag to disable it. Might save the effort of having to reintroduce the patch after the issue is fixed upstream.

PR #551, opened 2021-11-22
fix: use filenames as backups if languageId fails

rgrinberg:


rgrinberg:
As discussed, it's always wrong to error based on not recognizing a languageId. These things aren't reliable across editors.

PR #554, opened 2021-11-24
Separate ocamlformat_rpc servers for popup and for file formatting

panglesd:
This PR fixed a regression introduced in #549, where ocamlformat_rpc is used to format the whole file, but with the special configuration used for popups.

With this PR, there is two instances of the ocamlformat_rpc server: One with the options for popups, and one with the default options.

Until we have a way of having the `.ocamlformat` options instead of the default option, the formatting of a file fallback automatically to using the ocamlformat binary.

Including `.ocamlformat` options is not trivial for the following reasons:
- First, we have to aggregate the information for multiple files (`.ocamlformat` in the directory or parent directory, global `.ocamlformat` file, `.ocamlformat-ignore`...). All of these files have to be parsed.
- Moreover, not all options in `.ocamlformat` files are accepted by the rpc server: as of today, the "version=..." is not accepted by ocamlformat_rpc, but is valid in `.ocamlformat`.

I think that this should be done on the ocamlformat side, as the code is already there for the ocamlformat binary. One way of doing it would be to send a path to the server, and the server adopts all the options that ocamlformat would have when run on path. 
I can file an issue on ocamlformat repo, if you think that would be a good solution, and start to work on a PR.

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -380,7 +380,14 @@ module Formatter = struct
     let state : State.t = Server.state rpc in
     if Document.is_merlin doc then
       let* res =
-        let* res = Ocamlformat_rpc.format_doc state.ocamlformat_rpc doc in
+        let* res =
+          if false then
+            (* Until ocamlformat_rpc use the .ocamlformat file, we fallback to
+               the ocamlformat binary *)
+            Ocamlformat_rpc.format_doc state.ocamlformat_rpc doc
+          else
+            Fiber.return (Error `NotYet)
+        in
         match res with
         | Ok res -> Fiber.return @@ Ok res
         | Error _ -> Ocamlformat.run state.ocamlformat doc
@@ -534,7 +541,9 @@ let hover server (state : State.t)
     let+ doc = query_doc doc pos
     and+ typ =
       (* We ask Ocamlformat to format this type *)
-      let* result = Ocamlformat_rpc.format_type state.ocamlformat_rpc ~typ in
+      let* result =
+        Ocamlformat_rpc.format_type state.ocamlformat_rpc_popup ~typ
+      in
       match result with
       | Ok v ->
         (* OCamlformat adds an unnecessay newline at the end of the type *)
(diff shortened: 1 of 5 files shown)


panglesd:
After a bit of thinking, it is a bit strange to have a hanging ocamlformat_rpc server started but never used. 
I will work a little bit more on this PR, sorry for the noise.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -1150,17 +1159,45 @@ let start () =
                 Server.Batch.notification batch (PublishDiagnostics d));
             Server.Batch.submit batch))
   in
-  let ocamlformat_rpc = Ocamlformat_rpc.create () in
+  let ocamlformat_rpc = Ocamlformat_rpc.create ()
+  and ocamlformat_rpc_popup = Ocamlformat_rpc.create () in
```
rgrinberg, line 1163: So we're creating two instances of the rpc just so that we can use two different config options? That seems very wrong to me. The protocol should allow overriding config it it's really necessary.

ocaml-lsp-server/src/ocamlformat_rpc.ml
```diff
@@ -46,15 +47,18 @@ end = struct
...
       | Error e -> Exn_with_backtrace.reraise e)
 
-  let configure ~logger { io_thread; client; _ } =
+  let configure ~logger ~format_style { io_thread; client; _ } =
     (* We ask for 64 columns formatting as this appear to be the maximum size of
        VScode popups. TODO We should probably allow some flexibility for other
        editors that use the server. *)
+    let options =
+      match format_style with
+      | `VSCodePopup -> [ ("module-item-spacing", "compact"); ("margin", "63") ]
```
rgrinberg, line 56: Are we pushing vscode limitations onto other editors? Again that seems very wrong.

ocaml-lsp-server/src/ocaml_lsp_server.ml
panglesd: Yes, it seems wrong. I should have directly closed the PR when I realized it, sorry for your time taken.

PR #555, opened 2021-11-24
fix: disable ocamlformat-rpc for now

rgrinberg:
cc @panglesd @voodoos


panglesd:
I am not sure, why do you disable ocamlformat_rpc also for formatting types and generated code, and not only for formatting the whole file?
The fact that generated code does not follow the options for the whole document does not seem too bad, as anyway a run of ocamlformat binary is needed. And formatted types in popup are always easier to read.

rgrinberg:
Merged in master already

PR #559, opened 2021-11-27
Fix jump to hole on destruct/construct to jump only within the new text

ulugbekna:
comes in pair with https://github.com/ocamllabs/vscode-ocaml-platform/pull/784

it would be nice if we could cut a patch release for ocaml-lsp supporting 4.12 because it's quite an annoying bug


ocaml-lsp-server/src/client.ml
```diff
@@ -7,17 +7,16 @@ module Vscode = struct
         ~command:"editor.action.triggerSuggest" ()
 
     module Custom = struct
-      let next_hole ?start_position ~notify_if_no_hole () =
+      let next_hole ?in_range ~notify_if_no_hole () =
         let arguments =
           let arg_obj_fields =
             let notif_json =
-              Some ("notify-if-no-hole", Json.bool notify_if_no_hole)
+              Some ("shouldNotifyIfNoHole", Json.bool notify_if_no_hole)
```
rgrinberg, line 14: Isn't this a breaking change? What about old clients?

ocaml-lsp-server/src/document.ml
```diff
@@ -291,3 +291,22 @@ let edit t text_edit =
     TextDocumentEdit.create ~textDocument ~edits:[ `TextEdit text_edit ]
   in
   WorkspaceEdit.create ~documentChanges:[ `TextDocumentEdit edit ] ()
+
+let new_range_on_replace (range : Range.t) new_text =
+  let lines = String.split_lines new_text in
+  match lines with
+  | [] -> { range with end_ = range.start }
+  | several_lines ->
+    let n_lines = List.length several_lines in
```
rgrinberg, line 300: This variable should be inlined. Or its scope should be reduced.

ocaml-lsp-server/src/document.mli
```diff
@@ -64,3 +64,7 @@ val close : t -> unit Fiber.t
 val get_impl_intf_counterparts : Uri.t -> Uri.t list
 
 val edit : t -> TextEdit.t -> WorkspaceEdit.t
+
+(** [new_range_on_replace range new_text] returns a new range that will be
+    occupied by [new_text] when [new_text] replaces the text content in [range] *)
+val new_range_on_replace : Range.t -> string -> Range.t
```
rgrinberg, line 70: I find this function hard to understand.  It could be my own fault, but I think we could at least it break it down somewhat. For example:

```ocaml
val extend : Range.t -> Position.t -> Range.t
val end_position : string -> Position.t
```

There could be other ways to do this, so feel free to try something else.

ocaml-lsp-server/src/client.ml
ulugbekna: Ah, indeed, that's a breaking change. When working on the PR, I thought we'd be releasing one more ocaml-lsp for 4.12, which would have the fix. I can provide backward-compatibility or we could have one more release (I can handle it if you prefer). With backward compat, users will have everything working but if they happen to have a typed hole outside the newly generated text and the newly generated text doesn't, the cursor may jump away... (sorry)

```ocaml
let a = _ (* hole here, where the jump happens *)

type r = { k : int } 
let f ( r : r ) = (* when the user destructs the value [r], they get `{ k }`, so the cursor jumps to the first line *)
  ()
```

ulugbekna:
I hope address all of the comments tomorrow.

ocaml-lsp-server/src/document.mli
ulugbekna: I moved the function to `module Range`, renamed, and put more understandable comments. Let me know if it's still unclear.

ocaml-lsp-server/src/client.ml
ulugbekna: I can preserve the current structure `{ "notify-if-no-holes": <bool>; "position": <Position.t> }` and add a field `end-position: <Position.t>` to preserve somewhat backward compat. It'll result in an ugly solution, I'm afraid. Sorry for the mess. Previously, when we introduced custom requests we never had such problems (because they had simpler param requirements), and I haven't thought about making the params more flexible for future improvements 🙏

rgrinberg: If it's too much work to maintain backwards compatibility, we don't need to bother. At the end of the day, this feature is a small part of the plugin and users can live without it until they upgrade. But at the very least we should:

* Mention the breakage somewhere.
* Make the error message explicit about the breakage. So the old method should now at least say that this "method is renamed xx and is no longer supported".

I think in this case maintaining support for the old method is easy enough.

ulugbekna:
This PR was stuck because I had a bug in VS Code (the editor itself, seemingly) which caused crashes all the time regardless of the extension or ocaml-lsp version. I moved to VS Code Insiders with which I don't have crashes so far, so will test and update this and corresponding extension PR on Friday.

ulugbekna:
Okay, this PR should be ready for a final review and merge as well as its counterpart on the vscode extension side: https://github.com/ocamllabs/vscode-ocaml-platform/pull/784

The extension PR ignores if it sees old-style arguments passed to it, which is absolutely okay because those arguments are provided for a very small feature of jumping to the next hole on destruct. (tbh I wouldn't even mind removing that feature; it's so small and not very useful)

Release plan I have in mind:

1. We can cut a release for ocaml-lsp (with this PR included) - v 1.9.2
2. We cut a release for the vscode extension (with the counterpart PR and ocaml-lsp version manager support for ocaml-lsp 1.9.2) 

Wdyt?

rgrinberg:
Do we still have the issue that upgrading the vscode extension requires an upgrade of ocamllsp?

ocaml-lsp-server/src/client.ml
rgrinberg: This issue still stands.

ulugbekna: I made the extension to drop command invocations with old-style arguments (that change is in the counterpart PR in the extension repo). 

Since the extension with the command supporting old-style arguments was never released, users who use the older ocaml-lsp (ocaml < 4.13) will simply never know this feature (jumping to a hole on destruct/construct) existed, so they don't need to be worried or notified about anything.

People on a newer ocaml-lsp will enjoy the feature as if nothing happened. 

Does this make sense?

ulugbekna:
> Do we still have the issue that upgrading the vscode extension requires an upgrade of ocamllsp?

If you mean the other way around, then yes: the vscode extension still tracks closely the ocaml-lsp version installed in the sandbox, so it needs to be updated when a new ocaml-lsp is published. 

I plan to make changes so that the extension not only warns the user about their outdated ocaml-lsp, but also proposes a button to upgrade (this can happen after the extension release since it's been postponed for a while ?)

ocaml-lsp-server/src/client.ml
rgrinberg: Okay. If it was never released, then there's no problem.

rgrinberg:
We should probably still add a capability for this `next_hole` business though. I think the current setup is offering commands that are vscode only.

ulugbekna:
> We should probably still add a capability for this next_hole business though. I think the current setup is offering commands that are vscode only.

1. Do you mean a client-side capability? 
2. If yes, if we only send this command when the client is vscode (by keeping the client info in the server state), would that be enough? I'm just afraid adding capabilities will add verbosity on both sides to not much use

rgrinberg:
1. Yes, that's what I meant.
2. If it's a lot of work, we can skip it and detection based on the client is acceptable. We should do something to make it less work however.

Let's take this code for example:
```
      let completionItem_of_constructed_expr idx expr =
        let expr_wo_parens = deparen_constr_expr expr in
        let edit = { TextEdit.range; newText = expr } in
        let command =
          Client.Vscode.Commands.Custom.next_hole
            ~in_range:(Range.resize_for_edit edit)
            ~notify_if_no_hole:false ()
        in
        CompletionItem.create ~label:expr_wo_parens ~textEdit:(`TextEdit edit)
          ~filterText:("_" ^ expr) ~kind:CompletionItemKind.Text
          ~sortText:(sortText_of_index idx) ~command ()
      in
      List.mapi constructed_exprs ~f:completionItem_of_constructed_expr
```

I don't see where we make sure that the command is only for vscode users.

ocaml-lsp-server/src/exp_client_caps.mli
```diff
@@ -0,0 +1,9 @@
+open Import
```
rgrinberg, line 1: Can this module be a submodule some other module? Maybe Client perhaps?

ocaml-lsp-server/src/state.ml
```diff
@@ -64,7 +65,16 @@ let dune t =
...
   assert (t.init = Uninitialized);
-  { t with init = Initialized { params; workspaces; dune } }
+  { t with
+    init =
+      Initialized
+        { params
+        ; workspaces
+        ; dune
+        ; exp_client_caps =
+            lazy (Exp_client_caps.of_opt_json params.capabilities.experimental)
```
rgrinberg, line 75: Doesn't seem write to delay the error checking of this field.

ocaml-lsp-server/src/client.ml
```diff
@@ -5,29 +5,28 @@ module Vscode = struct
...
-            in
-            List.filter_opt [ pos_json; notif_json ]
-          in
-          match arg_obj_fields with
-          | [] -> [] (* no arguments -- the extension uses defaults *)
-          | fields ->
-            (* the use of a (json) object as the first and single argument to
-               the command is intended *)
-            [ `Assoc fields ]
+module CustomCommands = struct
```
rgrinberg, line 11: `Custom_commands`.

ocaml-lsp-server/src/state.ml
ulugbekna: I wouldn't say we do any validation. I was under assumption we could compute whether client has experimental capabilities only when user uses ocamllsp functionality that requires those capabilities.

ocaml-lsp-server/src/client.ml
```diff
@@ -1,33 +1,45 @@
...
 
+module Experimental_capabilities = struct
+  type t = bool
+
+  let of_opt_json (json : Json.t option) =
+    match json with
+    | Some (`Assoc fields) ->
+      Json.field fields "jumpToNextHole" Json.Conv.bool_of_yojson
+      |> Option.value ~default:false
+    | _ -> false
```
rgrinberg, line 11: Seems a bit lax to me. We should error out if this is anything other than `Some Assoc` or `None`.

rgrinberg (approved):

ocaml-lsp-server/src/client.ml
ulugbekna: Crashing lsp seems somewhat drastic? I would imagine sending a showMessage notification saying that the experimental capabilities are incorrect in another PR?

rgrinberg:
I don't think it would be drastic. If anything other than an object is provided in this field, there's clearly a programming mistake on the client and there's no point proceeding. This is not a big issue though.

Rudi.

PR #560, opened 2021-11-28
Avoid unexpected shell redirection to file in make install-test-deps

Khady:
Without the quotes in bash a file `=v0.14.0` is created with the output of the command and the installation can get stuck because the Y/N prompt isn't visible.


ulugbekna:
thanks!

PR #561, opened 2021-11-28
Add a new custom command ocamllsp/hoverExtended with support for variable verbosity

Khady:
This is a proposal to emulate merlin's behavior when multiple hover commands are received in a row on a single position. Each new call should increase the verbosity counter by 1. The PR is not ready. But I'd like to know if the idea looks fine before to invest more time on it.

https://discuss.ocaml.org/t/merlin-vs-ocaml-lsp/8887

This work is sponsored by Ahrefs


ulugbekna:
Wow, this works like a charm! Thanks! 

Regarding implementation:
1. Am I missing something or could we could use a single option to represent the last hover that happened and verbosity with which it happened? 
2. Maybe reset verbosity to 0 when increasing it doesn't change the type? I'd imagine this would eliminate the need to hover something to get the initial type (with verbosity 0)

https://user-images.githubusercontent.com/16353531/143781598-f539ec34-b372-4fca-bb93-9164c65723eb.mov

ulugbekna:
There is a somewhat related PR in merlin repo btw: https://github.com/ocaml/merlin/pull/1374

rgrinberg:
I understand the appeal of doing it on the server - we do it once and it automatically works everywhere. Unfortunately, there's also a serious downside. The LSP server is a basic block for editors to build functionality on top of, and this behavior is making a pretty basic request unpredictable. It might not matter for users of vscode, but in other editors, lsp requests can be scripted to provide higher level functionality and this is a huge wrench thrown in that work.

So I suggest we leave basic requests alone and do one of:
* Create a new request that automatically handles verbosity if you insist on handling the state on the server.
* Save the state on the client side and provide verbosity as a parameter in a custom request.

Khady:
> Regarding implementation:
> 1. Am I missing something or could we could use a single option to represent the last hover that happened and verbosity with which it happened? 
> 2. Maybe reset verbosity to 0 when increasing it doesn't change the type? I'd imagine this would eliminate the need to hover something to get the initial type (with verbosity 0)

1. sure no problem
2. because the position stored is the one of the cursor, not the one of the value, I think that getting the original hover is not a real problem, it's a matter of moving the cursor by one. But I'm fine with implementing this behavior if there's a consensus on it

> Create a new request that automatically handles verbosity if you insist on handling the state on the server.

Ok I can do that. Any suggestion on the name? HoverVerbose? HoverSmart?

ulugbekna:
Regarding designing a custom hover request, I think this may be an opportunity to design the command to also support selections (ie show types for hovers) that was also discussed in the ocaml discuss thread mentioned earlier I think. Reference of Scala metals implementation if interesting: https://github.com/scalameta/metals/pull/3060

Khady:
I've started to work on adding a new command. I'm not sure exactly how much work it would require in the client to become usable. I am scared that it becomes very troublesome to use. But in the process I found that it is possible to have multiple hover providers, which could be an alternative way to have this feature. One provider could give the normal type, one could give the doc, one could give a more and more verbose answer. But I don't know if one server can have multiple hover providers.

Anyway, in the process of adding a new command I had to duplicate quite a lot of code from `ocaml_lsp_server.ml` to `req_hover_extended.ml`. Because the later is called from the former. So all the utils can't be called from the custom commands. I wonder where I should move that code to avoid duplication.

https://github.com/Khady/ocaml-lsp/blob/c2bec75749a5d2c3968d2364e86ca2fb1fa196e5/ocaml-lsp-server/src/custom_requests/req_hover_extended.ml#L31-L80

ulugbekna:
> I've started to work on adding a new command. I'm not sure exactly how much work it would require in the client to become usable. I am scared that it becomes very troublesome to use. But in the process I found that it is possible to have multiple hover providers, which could be an alternative way to have this feature. One provider could give the normal type, one could give the doc, one could give a more and more verbose answer. But I don't know if one server can have multiple hover providers.

I haven't looked into this too closely, but I was under assumption that for vscode client, we simply add our own hover provider that uses the custom hover request-response and turn off the provider that uses the default LSP request for hovers. Does this make sense?

> Anyway, in the process of adding a new command I had to duplicate quite a lot of code from ocaml_lsp_server.ml to req_hover_extended.ml. Because the later is called from the former. So all the utils can't be called from the custom commands. I wonder where I should move that code to avoid duplication.

This [PR](https://github.com/ocaml/ocaml-lsp/pull/596) should help with this.

Khady:
A few months later, I'm back with an updated version of this PR. I'm not handling selections (solving one problem at a time). And I haven't tried the change in an editor yet. Sadly it looks like my change to send the verbosity to merlin no longer works

https://github.com/ocaml/ocaml-lsp/blob/259bf6e0a016d209c719ab9adecf1f88c072e71a/ocaml-lsp-server/src/document.ml#L281-L291

It was fine it my original commit https://github.com/Khady/ocaml-lsp/commit/a25a778bad222d1db3790ff16c535451f29759e2

```
Summary of all failing tests
 FAIL  __tests__/ocamllsp-hoverExtended.ts
  ● ocamllsp/hoverExtended › 

    expect(received).toMatchInlineSnapshot(snapshot)

    Snapshot name: `ocamllsp/hoverExtended  2`

    - Snapshot  - 1
    + Received  + 1

    @@ -1,9 +1,9 @@
      Object {
        "contents": Object {
          "kind": "plaintext",
    -     "value": "int",
    +     "value": "foo",
        },
        "range": Object {
          "end": Object {
            "character": 5,
            "line": 2,

      567 |     });
      568 |
    > 569 |     expect(result1).toMatchInlineSnapshot(`
          |                     ^
      570 |       Object {
      571 |         "contents": Object {
      572 |           "kind": "plaintext",

      at __tests__/ocamllsp-hoverExtended.ts:569:21
      at fulfilled (__tests__/ocamllsp-hoverExtended.ts:28:58)

  ● ocamllsp/hoverExtended › 

    expect(received).toMatchInlineSnapshot(snapshot)

    Snapshot name: `ocamllsp/hoverExtended  3`

    - Snapshot  - 1
    + Received  + 1

    @@ -1,9 +1,9 @@
      Object {
        "contents": Object {
          "kind": "plaintext",
    -     "value": "int",
    +     "value": "foo",
        },
        "range": Object {
          "end": Object {
            "character": 5,
            "line": 2,

      592 |     });
      593 |
    > 594 |     expect(result2).toMatchInlineSnapshot(`
          |                     ^
      595 |       Object {
      596 |         "contents": Object {
      597 |           "kind": "plaintext",

      at __tests__/ocamllsp-hoverExtended.ts:594:21
      at fulfilled (__tests__/ocamllsp-hoverExtended.ts:28:58)
```

ocaml-lsp-server/src/custom_requests/req_hover_extended.ml
```diff
@@ -0,0 +1,116 @@
...
+      (* We ask Ocamlformat to format this type *)
+      let* result = Ocamlformat_rpc.format_type state.ocamlformat_rpc ~typ in
+      match result with
+      | Ok v ->
+        (* OCamlformat adds an unnecessay newline at the end of the type *)
+        Fiber.return (String.trim v)
+      | Error `No_process -> Fiber.return typ
+      | Error (`Msg _message) ->
+        (* We log OCamlformat errors and display the unformated type *)
+        (* TODO: how to obtain the server *)
```
Khady, line 91: I could add the server as an arg to all the custom commands

https://github.com/ocaml/ocaml-lsp/blob/259bf6e0a016d209c719ab9adecf1f88c072e71a/ocaml-lsp-server/src/ocaml_lsp_server.ml#L720-L725

ocaml-lsp-server/test/e2e/__tests__/ocamllsp-hoverExtended.ts
```diff
@@ -0,0 +1,613 @@
...
+            "line": 2,
+          },
+          "start": Object {
+            "character": 4,
+            "line": 2,
+          },
+        },
+      }
+    `);
+  });
```
Khady, line 612: the rest of the tests are duplicated from textDocument/hover to check that my new command works properly. They can probably be removed to keep only ones playing with the verbosity once we know that things are working decently

ulugbekna:
Since we're creating a new request there should be no harm in keeping the state on the server side. I would hate for every client to have to implement the same thing.

AFAIU, Rudi is not against it as long as we have a custom request for it:

> So I suggest we leave basic requests alone and do one of:
> 
> Create a new request that automatically handles verbosity if you insist on handling the state on the server.
> Save the state on the client side and provide verbosity as a parameter in a custom request.

ocaml-lsp-server/src/custom_requests/req_hover_extended.ml
```diff
@@ -0,0 +1,116 @@
...
+
+  let of_jsonrpc_params_exn params : t =
+    let params_spec = { Custom_request.params_schema; of_jsonrpc_params } in
+    Custom_request.of_jsonrpc_params_exn params_spec params
+end
+
+let client_capabilities (state : State.t) =
+  (State.initialize_params state).capabilities
+
+(* TODO: duplicated with Hover_req.format_contents *)
```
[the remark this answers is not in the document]
ulugbekna: I think either a new file `hover_format` containing that fn or just expose the fn and use `Hover_req.format_contents` in `req_hover_extended.ml`?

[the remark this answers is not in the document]
Khady: I was also considering completely sharing the code of hover_req (textdocument/hover) and hover_extended (ocamllsp/hoverextended) as they are the same except for an extra verbosity argument

Khady:
At this point I don't have a strong opinion anymore. As soon as it's a new request it requires some extra work in every client anyway. I guess the new request can even support both. If verbosity is passed as an argument then it is forwarded to merlin. If it is omitted some server side state is used instead. Pleases everyone?

rgrinberg:
> At this point I don't have a strong opinion anymore. As soon as it's a new request it requires some extra work in every client anyway. I guess the new request can even support both. If verbosity is passed as an argument then it is forwarded to merlin. If it is omitted some server side state is used instead. 

Sounds fine to me.

> Pleases everyone?

It's a new and custom request, and you're the one taking the initiative, so feel free to make the final call.

ocaml-lsp-server/src/custom_requests/req_hover_extended.ml
```diff
@@ -0,0 +1,70 @@
...
+let meth = "ocamllsp/hoverExtended"
+
+module Request_params = struct
+  type t =
+    { text_document : TextDocumentIdentifier.t
+    ; cursor_position : Position.t
+    ; verbosity : int Json.Nullable_option.t
+          [@default None] [@yojson_drop_default ( = )]
+    }
+  [@@deriving_inline yojson] [@@yojson.allow_extra_fields]
```
Khady, line 15: I wrote of_jsonrpc_params by hand as it was done in another custom req, but is there a reason not to use deriving instead?

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -27,15 +32,40 @@ let format_contents ~syntax ~markdown ~typ ~doc =
...
+let handle server { HoverParams.textDocument = { uri }; position; _ } mode =
   Fiber.of_thunk (fun () ->
       let state : State.t = Server.state server in
       let doc =
         let store = state.store in
         Document_store.get store uri
       in
+      let verbosity =
+        let mode =
+          match (mode, Sys.getenv_opt "OCAMLLSP_HOVER_IS_EXTENDED") with
```
Khady, line 44: I used this to quickly experiment from my editor. I know this isn't encouraged, but this is convenient to have at least while in development

Khady:
the last commit factorizing the code of the two hover requests was easy to write so I thought it was worth raising the idea. But I'm happy to drop it if we prefer to keep the code of the two requests more separated

ocaml-lsp-server/docs/ocamllsp/hoverExtended-spec.md
```diff
@@ -0,0 +1,52 @@
...
+##### Request
+
+- method: `ocamllsp/hoverExtended`
+- params:
+
+  ```json
+  {
+    "textDocument": TextDocumentIdentifier,
+    "position": Position,
+    "verbosity": integer
```
Khady, line 34: not sure what the syntax it to explain that it is optional

ocaml-lsp-server/docs/ocamllsp/hoverExtended-spec.md
```diff
@@ -0,0 +1,52 @@
...
+	/**
+	 * The hover's content
+	 */
+	contents: MarkedString | MarkedString[] | MarkupContent;
+	/**
+	 * An optional range is a range inside a text document
+	 * that is used to visualize a hover, e.g. by changing the background color.
+	 */
+	range?: Range;
+```
```
Khady, line 52: the verbosity could also be returned. My idea is that a client could rely on the server to handle the verbosity most of the time, not passing it in the request. While storing the returned verbosity. Then it could offer some additional commands like `hoverLessVerbose` (to send the stored verbosity minus one) or `hoverSameVerbose` (to request the same verbosity)

Khady:
gentle ping on this one, I'd like to put a final effort so that it can get merged

ocaml-lsp-server/src/custom_requests/req_hover_extended.ml
rgrinberg: You can use either. The workflow with promotion is quite annoying though.

ocaml-lsp-server/src/hover_req.ml
rgrinberg: Sounds fine to me.

ocaml-lsp-server/src/document.mli
```diff
@@ -74,4 +74,5 @@ type type_enclosing =
   ; doc : string option
   }
 
-val type_enclosing : t -> Msource.position -> type_enclosing option Fiber.t
+val type_enclosing :
+  ?verbosity:int -> t -> Msource.position -> type_enclosing option Fiber.t
```
rgrinberg, line 78: can you make this argument mandatory? i don't find optional arguments as useful most of the time.

ocaml-lsp-server/src/custom_requests/req_infer_intf.mli
```diff
@@ -4,4 +4,8 @@ val capability : string * Json.t
 
 val meth : string
 
-val on_request : params:Jsonrpc.Structured.t option -> State.t -> Json.t Fiber.t
+val on_request :
+     params:Jsonrpc.Structured.t option
+  -> State.t Server.t
```
rgrinberg, line 9: why do we add this unused argument everywhere?

rgrinberg:
Looks OK apart from some minor issues. By the way, we've deprecated the typescript test suite and are writing tests in OCaml from now on. Do you mind porting your tests to OCaml? There's plenty of examples in e2e-new.

ocaml-lsp-server/src/custom_requests/req_infer_intf.mli
Khady: It is necessary to be able to do some logging, similar to what is done in the hover's handler.

```
State.log_msg server ~type_:Warning ~message
```

And with the factorized version I need it to be able to call hover's handler from hoverExtend.

Also I tend to think that it is better if custom requests tend to mimic normal requests. Which means it's better if they receive `State.t Server.t` than just `State.t`. I tried to do a minimal chance. A more thorough and clean version would be

```diff
- val on_request : params:Jsonrpc.Structured.t option -> State.t -> Json.t Fiber.t
+ val on_request : params:Jsonrpc.Structured.t option -> State.t Server.t -> Json.t Fiber.t`
```

rgrinberg: If it's needed for logging, let's pass it to functions that actually log something. I see this argument being added in a few places without it actually being used.

> Also I tend to think that it is better if custom requests tend to mimic normal requests. Which means it's better if they receive State.t Server.t than just State.t. I tried to do a minimal chance. A more thorough and clean version would be

Some sort of consistency of type signature to me is a much lesser goal than making signatures as precise and as informative as possible.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -718,7 +720,7 @@ let on_request :
     | Some handler ->
       Fiber.return
         ( Reply.later (fun send ->
-              let* res = handler ~params state in
+              let* res = handler ~params rpc state in
```
rgrinberg, line 723: Instead of this change, you can just partially apply `rpc` to the requests that actually need this argument.

Khady:
I think I addressed all your comments.

For the ocaml e2e tests, I've been able to write something and get it running. But I'm a bit lost. It works fine for the default hover request, even when in extended mode thanks to the environment variable. But for hoverExtended I always receive a null answer. Am I missing some kind of special wait that must be done for custom requests only?

ulugbekna:
> But for hoverExtended I always receive a null answer. Am I missing some kind of special wait that must be done for custom requests only?

`Req_hover_extended.on_request` is returning `Null` (I put an exception in that pattern matching branch, which throws), so it must be `Hover_req.handle` that is returning `None` somehow.

ocaml-lsp-server/src/document.ml
```diff
@@ -274,9 +274,20 @@ module Merlin = struct
...
+        let pipeline =
+          match verbosity with
+          | 0 -> pipeline
+          | verbosity ->
+            let source = source doc in
+            let config = Mpipeline.final_config pipeline in
+            let config =
+              { config with query = { config.query with verbosity } }
+            in
+            Mpipeline.make config source
```
ulugbekna, line 289: I think this code needs to be adapted if this PR is merged after #904.

Khady:
Ok I figured out the problem.There was a hardcoded position that I forgot to remove...

ocaml-lsp-server/src/document.ml
Khady: that PR was merged, so is there a need for the code to be adapted?

rgrinberg: The current code looks fine to me. You aren't doing any illegal operations outside the merlin thread.

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -37,8 +42,35 @@ let handle server { HoverParams.textDocument = { uri }; position; _ } =
       match Document.kind doc with
       | `Other -> Fiber.return None
       | `Merlin merlin -> (
+        let verbosity =
+          let mode =
+            match (mode, Sys.getenv_opt "OCAMLLSP_HOVER_IS_EXTENDED") with
```
rgrinberg, line 47: You might want to hoist up getting the environment variable to the top so it's only done once.

rgrinberg (approved):

Khady:
Thanks to both of you for the help!

copy:
In case someone is wondering how to call this with nvim's lsp:

```lua
    vim.keymap.set("n", "<f2>", function()
        local params = vim.lsp.util.make_position_params()
        params["verbosity"] = 1
        vim.lsp.buf_request(0, "ocamllsp/hoverExtended", params, vim.lsp.handlers.hover)
    end, bufopts)
```

PR #562, opened 2021-11-28
Upgrade OCaml version in CI

ulugbekna:


ulugbekna:
@rgrinberg do you know off the top of your head why URI tests are failing on windows? otherwise, I can take a look

smorimoto (approved):

smorimoto:
To me, the content of the test doesn't seem appropriate in the first place.

tatchi:
From what I see in the code, a `/` is [prepended to the path](https://github.com/ocaml/ocaml-lsp/blob/master/lsp/src/uri0.ml#L52) on every platform except windows. I guess that's probably why it fails on this platform.

ulugbekna:
Windows-related failures can be fixed in another PR. 

I'm not sure, but should nix-related files be updated to the new compiler? Also can happen in another PR.

Issue #547, opened 2021-11-07
Feature Request: More folding!


cdaringe:
# Problem

Only a few things fold!

# Discussion

Let's fold _more_! It's quite common in ocaml for fn authors to inline all sorts of foldable _stuff_ inside fns. fns in fns in fns! 

Specifically, I desire support of `let ... in` expressions. Unsure about `and`. I don't see a lot of `and` out in the wild, besides in the stdlib.

![image](https://user-images.githubusercontent.com/1003261/140658303-afbaef0b-bd8d-4b02-bbac-9850da5bfa6d.png)

Seems like [a](https://github.com/ocaml/ocaml-lsp/blob/16ece09d6dc6af32b9d48dfa095cd5b249ce2162/lsp/src/types.mli#L234) & [b](https://github.com/ocaml/ocaml-lsp/blob/e58630b1077a3b5f9442993756cf2bc5d11bfa41/ocaml-lsp-server/src/ocaml_lsp_server.ml#L590) are relevant here 🤔

ulugbekna:
That's vscode right? Set `folding` to `indentation` in your vscode settings :-) 

![image](https://user-images.githubusercontent.com/16353531/140658648-0e475aba-4ee5-456d-b553-2534d8eeaca8.png)
![image](https://user-images.githubusercontent.com/16353531/140658690-583e5343-d87c-4d65-b805-cd5e2d5d4edb.png)

I'm hesitant about closing this issue, since we could improve folding support in ocaml-lsp as well.

cdaringe:
Yes, I did consider indentation as well. It felt less precise, breaking at non-logically grouped blocks

rgrinberg:
Feel free to send a PR. It should be enough to traverse the let bindings and inspect their locations to generate the folds.

tatchi:
Hey, I quickly looked at that issue and noticed that what we can back from Merlin outline request only contains the "top level bindings". For instance, the following code:

```ocaml
let a =
  let b =
    let i = 6 in
    let () = print_endline "hello" in
    let result = i + 1 in
    result
  in
  let c = b + 3 in
  c
```

produces:

```json
{
  "class": "return",
  "value": [
    {
      "start": {
        "line": 1,
        "col": 0
      },
      "end": {
        "line": 9,
        "col": 3
      },
      "name": "a",
      "kind": "Value",
      "type": "int",
      "children": [],
      "deprecated": false
    }
  ],
  "notifications": [],
  "timing": {
[...]

PR #568, opened 2021-12-22
More folding 

tatchi:
This PR implements #547

That should be easier to review it by commits, at least I tried to make it so.

* 7eb4c13 : uses `toMatchInlineSnapshot` instead of manually writing the expected result
* d87c00b: adds more test cases that will be useful later
* 19f85c9: switches to the new implementation. No tests broke
* c6d5d7c: supports more `let` folding + `letop` (let+,..) and `pexp_extension` (let%...) 

Since I believe the initial test didn't cover all the different cases, I can't guarantee any regression; i.e something that was foldable previously that's not anymore with that PR. Don't hesitate if you can think of anything I might have forgotten, it should be pretty easy to add.

This PR only focuses on adding more `let` folding. `let` inside classes are not supported (as with the current implementation). I think that we could support more stuff like `method` and `let` inside classes. `records` and function `args` might be nice too. Here's an example in javascript:

https://user-images.githubusercontent.com/5595092/147128674-fcddacff-b05a-4902-8e8b-28be84d41776.mov

I would be happy to add support for more elements, either in that PR or in a follow up one. In any case, I think the current PR set up the base code, and that should be easy to support more use cases as we discover them.


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -656,33 +656,205 @@ let text_document_lens (state : State.t)
...
-        else
-          let items = item.children @ items in
-          let range = folding_range range in
-          loop (range :: acc) items
+  let ranges = ref [] in
+  let push (range : Range.t) =
+    if range.end_.line - range.start.line >= 2 then
+      ranges := folding_range range :: !ranges
+    else
+      ()
```
ulugbekna, line 669: we don't need an else case here

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -656,33 +656,205 @@ let text_document_lens (state : State.t)
...
+        match expr.pexp_desc with
+        | Parsetree.Pexp_let (_, value_bindings, expr) ->
+          List.iter value_bindings ~f:(fun value_binding ->
+              iterator.value_binding iterator value_binding);
+          iterator.expr iterator expr
+        | Parsetree.Pexp_fun (_, _, _, expr) -> iterator.expr iterator expr
+        | Parsetree.Pexp_open (_, expr) -> iterator.expr iterator expr
+        | Parsetree.Pexp_letop letop ->
+          (* Location is not correct. It include the location of the whole
+             expression. See: https://github.com/ocaml/ocaml/pull/10682 *)
```
tatchi, line 735: Location is wrong (see link) with `letop` (let+,..). That's a bit annoying because when folding, it hides the whole expression. It works fine with regular `let` or `let%..`. Should it be better to disable support for `letop` for now ?

https://user-images.githubusercontent.com/5595092/147131513-6de7f6e1-4fdc-4ce4-a124-479b9632dfcb.mov

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -656,33 +656,205 @@ let text_document_lens (state : State.t)
...
+    | `Implementation structure -> iterator.structure iterator structure
+  in
+  let doc = Document_store.get state.store uri in
+  let+ () =
+    Document.with_pipeline_exn doc (fun pipeline ->
+        let parsetree = Mpipeline.reader_parsetree pipeline in
+        fold_over_parsetree parsetree)
   in
-  Some folds
+  Some (List.rev !ranges)
```
tatchi, line 857: Reverting the list is not necessary. I did it only to keep it in the same order as the previous implementation and avoid noises in test results.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -656,33 +656,205 @@ let text_document_lens (state : State.t)
...
+          (module_type : Parsetree.module_type) =
+        match module_type.pmty_desc with
+        | Parsetree.Pmty_ident _ -> ()
+        | Parsetree.Pmty_signature signature ->
+          iterator.signature iterator signature
+        | Parsetree.Pmty_functor (_, _) -> ()
+        | Parsetree.Pmty_with (_, _) -> ()
+        | Parsetree.Pmty_typeof _ -> ()
+        | Parsetree.Pmty_extension _ -> ()
+        | Parsetree.Pmty_alias _ -> ()
```
ulugbekna, line 705: 1. cases which do the same thing can be collapsed, eg `| Pmty_typeof _ | Pmty_extension _ -> ()` 
2. we can keep the `Parsetree.` prefix only for the first case and remove for others - it hints the compiler where to look for the value constructor

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -656,33 +656,205 @@ let text_document_lens (state : State.t)
 
 let folding_range (state : State.t)
```
ulugbekna, line 657: I feel like this is a too involved function (too large at least) and deserves to be moved to a separate file :-)

ulugbekna (commented):
a couple of quick notes; can have a full review on Friday

btw, thanks for the PR, impressive work!

tatchi:
Might be cool to be able to fold the cases of a `match ... with` expression. Seems like rust has that:

https://user-images.githubusercontent.com/5595092/147272449-0f0533a3-f030-4ac7-a80c-dbc63dcbf39d.mov

PR #570, opened 2022-01-01
refactor: functorize io parsing

rgrinberg:


rgrinberg:
Looks like the CI failure are menhir related. Though it is not our use of menhir, but merlin's.

PR #573, opened 2022-01-05
More more folding

ulugbekna:
Adds folding support for 
- record expressions and patterns
- match expressions, case patterns, case expressions

TODO: 
- [x] add tests

@tatchi if you'd like to have a look

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## Features
 
-- Add better support for code folding: more precise folds for `let` bindings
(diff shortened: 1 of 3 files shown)


ulugbekna:
> Do we want to also support Ptype_record ? Not sure if that makes sense though 🤔

Folding for `type r = ...` is enough I think

Added support for folding patterns as fn args.

Seems good to me, wdyt?

tatchi:
> Seems good to me, wdyt?

Sounds good yes 👍

PR #585, opened 2022-01-10
Add constraint to menhir as test deps

Nymphium:
Considering below logs, Menhir 20211230 may contain a bug about type inference:
https://github.com/Nymphium/ocaml-lsp/runs/4754176481?check_suite_focus=true#step:10:31

This PR adds constraint to Menhir in test dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ all:
 # results in a conflict
 .PHONY: install-test-deps
 install-test-deps:
-	opam install menhir cinaps 'ppx_expect>=v0.14.0' \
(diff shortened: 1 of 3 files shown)


rgrinberg:
Thanks. I updated the nix setup as well.

@tatchi could you investigate the failures on windows? it seems like we can finally run the tests there.

tatchi:
> Thanks. I updated the nix setup as well.
> 
> @tatchi could you investigate the failures on windows? it seems like we can finally run the tests there.

Sure, I'll investigate that 👍

PR #587, opened 2022-01-19
[Folding ranges]: support return type annotation

tatchi:
Support folding with explicit return type annotation. Fix the following case:

https://user-images.githubusercontent.com/5595092/150073804-88daba85-8fd9-4c80-829b-7d99d2afb103.mov


ulugbekna:
Thanks! 🙏

PR #589, opened 2022-01-19
[Folding ranges]: support try/with

tatchi:
It only traverses the `Pexp_try` for now. Not sure how we could support folding the `try` and `with` separately due to the way locations are reported (same as with the if/else).

We could add support for folding the entire `try/with` expression though. Do we want to support that?


ulugbekna:
I think it could be useful to be able to fold try/with expressions

PR #596, opened 2022-01-28
[refactor] extract hover request handling in separate file

ulugbekna:
Mostly moves code around to help with https://github.com/ocaml/ocaml-lsp/pull/561#issuecomment-1008457123:

> Anyway, in the process of adding a new command I had to duplicate quite a lot of code from ocaml_lsp_server.ml to req_hover_extended.ml. Because the later is called from the former. So all the utils can't be called from the custom commands. I wonder where I should move that code to avoid duplication.

### DIFF
--- a/ocaml-lsp-server/src/hover_req.mli
+++ b/ocaml-lsp-server/src/hover_req.mli
@@ -1,0 +1,8 @@
+open Import
+
+(** This module contains functionality to handle `textDocument/hover` LSP
+    request. *)
+
+(** [handle server hover_params] provides a response for LSP request
+    `textDocument/hover` *)
+val handle : State.t Server.t -> HoverParams.t -> Hover.t option Fiber.t
--- a/ocaml-lsp-server/src/document.ml
+++ b/ocaml-lsp-server/src/document.ml
@@ -241,6 +241,26 @@ let dispatch t command =
 let dispatch_exn t command =
   with_pipeline_exn t (fun pipeline -> Query_commands.dispatch pipeline command)
 
+let doc_comment doc pos =
+  let+ res =
+    let command = Query_protocol.Document (None, pos) in
+    dispatch_exn doc command
+  in
+  match res with
+  | `Found s
+  | `Builtin s ->
+    Some s
+  | _ -> None
+
+let type_enclosing doc pos =
+  let command = Query_protocol.Type_enclosing (None, pos, None) in
+  let+ res = dispatch_exn doc command in
+  match res with
+  | []
+  | (_, `Index _, _) :: _ ->
+    None
+  | (location, `String value, _) :: _ -> Some (location, value)
+
 let close t =
   match t with
   | Other _ -> Fiber.return ()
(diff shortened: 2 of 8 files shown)


ocaml-lsp-server/src/merlin_query.ml
```diff
@@ -0,0 +1,21 @@
...
+  | _ -> None
+
+let type_enclosing doc pos =
+  let command = Query_protocol.Type_enclosing (None, pos, None) in
+  let+ res = Document.dispatch_exn doc command in
+  match res with
+  | []
+  | (_, `Index _, _) :: _ ->
+    None
+  | (location, `String value, _) :: _ -> Some (location, value)
```
rgrinberg, line 21: perhaps move this to Document? This module doesn't really have a well defined purpose on its own.

ocaml-lsp-server/src/hover_req.mli
```diff
@@ -0,0 +1,3 @@
+open Import
+
+val run : State.t Server.t -> State.t -> HoverParams.t -> Hover.t option Fiber.t
```
rgrinberg, line 3: This needs a module doc strings.

rgrinberg: Do we need to pass the state? Isn't it a part of the server already.

rgrinberg: Yeah, but now we've given this function an explicit interface. It's important to keep that tidy. Whatever `later` does is less important. And `later` should probably stop passing the state altogether. It's confusing to be able to acquire it from two places.

ulugbekna: Do you propose 
option a) ignore the state passed within `later` callback and get state from the server 
option b) first send another PR where `later` doesn't pass `state` in the callback at all ? 

I'm not very confident in my understanding of how `later` works, so option b will take a couple of days. It would be nice to remove #561's blocker faster.

rgrinberg: a) is fine. So is b)

If this refactoring work for another PR you could always just include it in that PR as separate commits. What if we decide to reject the other PR? Or what if the requirements for that PR change?

ulugbekna:
> What if we decide to reject the other PR? Or what if the requirements for
that PR change?

Good point, but this PR on its own is also useful in my opinion

Ulugbek

ulugbekna:
Should be ready to go

ocaml-lsp-server/src/logger.ml
```diff
@@ -0,0 +1,7 @@
+open Import
+
+let msg server ~type_ ~message =
+  let state = Server.state server in
+  task_if_running state.State.detached ~f:(fun () ->
+      let log = LogMessageParams.create ~type_ ~message in
+      Server.notification server (Server_notification.LogMessage log))
```
rgrinberg, line 7: Move this to State perhaps? A single function module is overkill.

rgrinberg (approved):
Looks good. One last possible issue and this is ready to be squashed and merged.

PR #597, opened 2022-01-30
fix: crash when inferring interface

rgrinberg:
Do not handle when trying to inferring interfaces for mli only modules.

### DIFF
--- a/ocaml-lsp-server/src/inference.ml
+++ b/ocaml-lsp-server/src/inference.ml
@@ -39,23 +39,28 @@ let language_id_of_fname s =
 
 let force_open_document (state : State.t) uri =
   let filename = Uri.to_path uri in
-  let text = Io.String_path.read_file filename in
-  let languageId = language_id_of_fname filename in
-  let text_document =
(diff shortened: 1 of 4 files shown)


ocaml-lsp-server/src/inference.ml
```diff
@@ -39,23 +39,28 @@ let language_id_of_fname s =
...
-  let+ doc =
-    Document.make (State.wheel state) state.merlin_config
-      ~merlin_thread:state.merlin params
-  in
-  Document_store.put state.store doc;
-  doc
+  match Io.String_path.read_file filename with
+  | exception Sys_error _ ->
+    Log.log ~section:"warning" (fun () ->
+        Log.msg "Unable to open file" [ ("filename", `String filename) ]);
```
ulugbekna, line 45: is the section expected to be "warning"?

ulugbekna (approved):
Thanks, LGTM!

ocaml-lsp-server/src/inference.ml
rgrinberg: I suppose it is totally expected. I made it debug

PR #599, opened 2022-02-01
fix: async io for ocamlformat-rpc-lib

rgrinberg:
cc @panglesd

### DIFF
--- a/ocaml-lsp-server/src/ocamlformat_rpc.ml
+++ b/ocaml-lsp-server/src/ocamlformat_rpc.ml
@@ -1,21 +1,28 @@
 open Import
 open Fiber.O
 
-let await_no_cancel task =
-  let+ res = Lev_fiber.Thread.await task in
-  match res with
-  | Ok s -> s
-  | Error `Cancelled -> assert false
-  | Error (`Exn exn) -> Exn_with_backtrace.reraise exn
+module Ocamlformat_rpc = Ocamlformat_rpc_lib.Make (struct
+  type 'a t = 'a Fiber.t
+
+  let return a = Fiber.return a
+
+  let ( >>= ) x f = Fiber.bind x ~f
+
+  type ic = Lev_fiber_csexp.Session.t
+
+  type oc = Lev_fiber_csexp.Session.t
+
+  let read = Lev_fiber_csexp.Session.read
+
+  let write t s = Lev_fiber_csexp.Session.write t (Some s)
+end)
 
 module Process : sig
   type t
 
   val pid : t -> Pid.t
 
-  val thread : t -> Lev_fiber.Thread.t
-
-  val client : t -> Ocamlformat_rpc_lib.client
+  val client : t -> Ocamlformat_rpc.client
 
   val create :
        logger:(type_:MessageType.t -> message:string -> unit Fiber.t)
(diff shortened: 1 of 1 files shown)


rgrinberg:
@panglesd can you take over this and switch to v2?

panglesd:
Sure!

ocaml-lsp-server/src/ocamlformat_rpc.ml
```diff
@@ -69,26 +66,26 @@ end = struct
...
-      let stdin_i, stdin_o = Unix.pipe () in
-      let stdout_i, stdout_o = Unix.pipe () in
+    let* pid, stdout, stdin =
+      let* stdin_i, stdin_o = Lev_fiber.Io.pipe ~cloexec:true () in
+      let+ stdout_i, stdout_o = Lev_fiber.Io.pipe ~cloexec:true () in
       let pid =
-        Spawn.spawn ~prog:bin ~argv:[ bin ] ~stdin:stdin_i ~stdout:stdout_o ()
+        Spawn.spawn ~prog:bin ~argv:[ bin ]
+          ~stdin:(Lev_fiber.Fd.fd (Lev_fiber.Io.fd stdin_i))
+          ~stdout:(Lev_fiber.Fd.fd (Lev_fiber.Io.fd stdin_o))
```
panglesd, line 75: Shouldnt this be:
```suggestion
          ~stdout:(Lev_fiber.Fd.fd (Lev_fiber.Io.fd stdout_o))
```
From my understanding, and what it replaces, it seems so. However, when I make the change, the server crash with a `Sys_blocked io`...

rgrinberg: Thanks. This is fixed.

rgrinberg: How do you get the Sys_blocked io error?

panglesd: Unfortunately `make test` does not display well errors/logging, so I pinned my development version of `ocaml-lsp` in a switch, open vs-code on this switch, and hover a value or type with my mouse. Then, I can see the errors in the bottom panel, `Output` tab, selecting `OCaml Language Server` in the list.

image

panglesd: Also, lines 159 and 161 of `ocamlformat_rpc.ml` should be "inverted" to un-`Disable` the rpc, to get the error.

I once committed that (and a new test for formatted output on hover) but it got erase by one of your force-push, I don't know if that was intentional.

rgrinberg: > Also, lines 159 and 161 of ocamlformat_rpc.ml should be "inverted" to un-Disable the rpc, to get the error.

Yes indeed. I forgot about that.

> I once committed that (and a new test for formatted output on hover) but it got erase by one of your force-push, I don't know if that was intentional.

Whoops that was completely accidental. Could you send a separate PR for the test though?

rgrinberg: Okay, I think I fixed the blocking issue.

panglesd:
I pushed the switch to V2.

A comment: Unfortunately, `ocamlformat-rpc` crashes on configuration error, so for instance if we try to format the file using the rpc `v0.20.1` but the `.ocamlformat` file has `version = 0.19.0`, then the rpc crash (it is relaunched just after, and the file is still formatted using the `ocamlformat` binary, so that's mostly invisible to the user).
I can open a PR to fix this on the `ocamlformat-rpc` side.

I added a test that types are formatted (I can move it in a separate PR, but the test will require this PR to pass)

ocaml-lsp-server/test/e2e/__tests__/textDocument-hover.test.ts
```diff
@@ -35,6 +35,40 @@ describe("textDocument/hover", () => {
...
+      contents: { kind: "plaintext", value: outdent`int ->
+                                                    int ->
+                                                    int ->
+                                                    int ->
+                                                    int ->
+                                                    int ->
+                                                    int ->
+                                                    int ->
+                                                    int ->
+                                                    int` },
```
rgrinberg, line 64: @panglesd does this generated type look right? It seems like a lot of lines for a pretty simpler signature.

panglesd: The value 
``` ocaml
let f a b c d e f g h i = 1 + a + b + c + d + e + f + g + h + i
```
has type `int -> int -> int -> int -> int -> int -> int -> int -> int -> int`, which is 67 characters. But we send `ocamlformat_rpc` the option `("margin", "63")`, so it breaks every line. So this is the intended behaviour!
We could modify the value `f` to something like 
``` ocaml
let f (_: string option) (_: float -> string) (_: (string, char) result) = 1
```
 to get 
``` ocaml
string option ->
(float -> string) ->
(string, char) result ->
int
```
which is less lines.

rgrinberg:
@panglesd what version of ocamlformat-rpc do we need? It seems like 0.21.0 isn't enough

panglesd:
I'll be happy to investigate, but I tried it and it seems to work with `0.21.0` (which is the one needed, from the changelog).
What error did you encountered?

rgrinberg:
I'm referring to the error in CI:

```
File "ocaml-lsp-server/src/ocamlformat_rpc.ml", line 6, characters 25-49:
6 | module Ocamlformat_rpc = Ocamlformat_rpc_lib.Make (struct
                             ^^^^^^^^^^^^^^^^^^^^^^^^
Error: Unbound module Ocamlformat_rpc_lib.Make
```

rgrinberg:
Never mind, it seems like the test deps are downgrading the dependencies for some reason:

```
↘ downgrade ocamlformat-rpc-lib 0.21.0 to 0.20.0 [required by ocamlformat-rpc]
[33](https://github.com/ocaml/ocaml-lsp/runs/5728545902?check_suite_focus=true#step:8:33)
```

panglesd:
Ah, yes.

After dependencies are installed, `make install-test-deps` installs ocamlformat with the version found in `.ocamlformat` (that is, `0.19.0`).

For some reason, [this line](https://github.com/ocaml-ppx/ocamlformat/blob/0.21.0/ocamlformat-rpc.opam#L15) was added, making the version of `ocamlformat` and `ocamlformat-rpc` tight together, starting from `0.21.0`.
So the highest version of `ocamlformat-rpc` than can be installed with `ocamlformat.0.19.0` is `ocamlformat-rpc.0.20.0`.

But `ocamlformat-rpc.0.20.0` requires `ocamlformat-rpc-lib.0.20.0`, which is why it is downgraded...

panglesd:
I guess this would be fixed by switching to `ocamlformat.0.21.0` (see #609)

PR #609, opened 2022-02-04
Preview: upgrade ocamlformat to 0.21.0

gpetiot:
Based on #533

This is a preview of the formatting of your project codebase with the upcoming `ocamlformat.0.21.0`. Note that this package is not yet released in opam (so the linting/formatting task of your CI won't pass), the formatting might also change slightly in the released version as we try to integrate feedback from our users.

As always, we are happy to hear your feedback!  Thank you for using ocamlformat!

### DIFF
--- a/ocaml-lsp-server/src/import.ml
+++ b/ocaml-lsp-server/src/import.ml
@@ -122,8 +122,10 @@ include struct
   module MarkupContent = MarkupContent
   module MarkupKind = MarkupKind
   module MessageType = MessageType
+
   module OptionalVersionedTextDocumentIdentifier =
     OptionalVersionedTextDocumentIdentifier
+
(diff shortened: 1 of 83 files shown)


ocaml-lsp-server/src/import.ml
```diff
@@ -122,8 +122,10 @@ include struct
   module MarkupContent = MarkupContent
   module MarkupKind = MarkupKind
   module MessageType = MessageType
+
   module OptionalVersionedTextDocumentIdentifier =
     OptionalVersionedTextDocumentIdentifier
+
```
smorimoto, line 128: Is this intentional?

rgrinberg:
Upgraded manually, thanks.

PR #613, opened 2022-02-07
Various small improvements

rgrinberg:
Aimed at minimizing the use of stdune in public libraries.


ulugbekna:
What’s the goal of minimising dependence on stdune, out of curiosity? Especially, when we can’t completely remove dependency for certain modules at this point

rgrinberg:
At least we will not increase our reliance on it over time. We are fairly close to getting rid of it almost everywhere however. We should be able to do it once dune 3.0 is out.

PR #614, opened 2022-02-07
Fix Code_action_error.combine

ulugbekna:


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -331,15 +331,11 @@ module Code_action_error = struct
...
-    | Exn _, Exn _ -> y
-    | Exn _, _
-    | _, Exn _ ->
-      x
-    | (Need_merlin_extend _ as r), Initial
-    | Initial, (Need_merlin_extend _ as r) ->
-      r
+    | Initial, _ -> y (* [Initial] cedes to any *)
+    | _, Initial -> x
+    | _, Exn _ -> y (* [Exn] takes over any *)
```
rgrinberg, line 336: When we have two exceptions, we should pick the first one. It doesn't matter much in practice, but it's the behavior we prefer elsewhere.

rgrinberg (approved):

PR #616, opened 2022-02-08
Fix cancel request param (de)serialization

ulugbekna:
- fix: better missing json-field error formatting
- fix: $/cancelRequest params were incorrectly (de)serialized

Noticed this problem luckily thanks to learning how to use logging yesterday :D

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-# Unreleased 
+# Unreleased
 
 ## Features
 
@@ -16,6 +16,8 @@
 - More accurate completion kinds.
   New completion kinds for variants and fields. Removed inaccurate completion
   kinds for constructors and types. (#510)
+
+- Fix handling request cancellation (#616)
 
 # 1.9.1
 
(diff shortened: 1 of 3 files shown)


ulugbekna:
1. You mean a changelog entry saying cancellation works now? I'm not sure it does because I see a lot of responses for cancelled requests and no `RequestCancelled` errors. Or is this expected? 

Anyway, I don't mind adding a changelog entry anyway. 

2. In the vscode output channel, I see requests with the same ID (number in parentheses). Should that be possible at all? 

```
[Trace - 7:36:09 PM] Sending request 'textDocument/codeLens - (15)'.
...
[Trace - 7:38:23 PM] Sending request 'textDocument/codeAction - (15)'.
```

rgrinberg:
> You mean a changelog entry saying cancellation works now? I'm not sure it does because I see a lot of responses for cancelled requests and no RequestCancelled errors. Or is this expected?

There would be no error if cancellation failed. We'd simply do more work on the server side.

> In the vscode output channel, I see requests with the same ID (number in parentheses). Should that be possible at all?

I guess it's possible. As long as they aren't concurrent we should deal with it easily.

PR #620, opened 2022-02-12
refactor: move side-effects into fiber

rgrinberg:
If a function returns Fiber.t, it should shift all side effects into the fiber.

I've been sloppy about this before, but from now let's keep it strict. cc @ulugbekna


rgrinberg:
Note that there's usually no need to wrap the entire thing with `Fiber.of_thunk`. The pure parts can of course be done immediately. But for a first pass, I just wrapped whatever looked suspicious.

PR #630, opened 2022-02-23
Remove vendored dependencies

tmattio:
This PR removes the vendoring of `cmdliner`, `omd`, `octavius` and `uutf`.

This fixes conflicts with opam-installed libraries when using ocaml-lsp-server as a library.


rgrinberg:
Merged in #634

PR #631, opened 2022-02-23
Expose ocaml-lsp-server as a public library

tmattio:


rgrinberg:
I don't think it's a good idea to install this stuff. Can't you keep these changes private?

PR #632, opened 2022-02-23
Unvendor dune

tmattio:
This unvendors dune and uses the library published on opam instead.
Depends on https://github.com/ocaml/dune/pull/5483 because of `ocamlc_loc`


rgrinberg:
There should be opam updates to this files.

> Depends on https://github.com/ocaml/dune/pull/5483 because of ocamlc_loc

This should be moved to its own package. We don't want to depend on the entirety of dune just to use a single library.

rgrinberg:
For now, the only parts of this RP that are acceptable are the ones that are unvendoring dune-rpc, stdune, fiber, dyn.

rgrinberg:
I unvendored stdune, fiber, ordering, xdg, dyn in master. Will unvendor the rpc library before the release.

tmattio:
Replaced by #869

PR #645, opened 2022-03-31
refactor: simplify inferred doc opening

rgrinberg:
[force_open_impl] is always [true]. There's no need to pass it


ulugbekna (approved):
Looks good to me.

PR #654, opened 2022-04-05
fix(dune): deduplicate dune and merlin diagnostics

rgrinberg:
Try to deduplicate dune and merlin diagnostics. Messages with identical
locations and error messages differing only by whitespace are treated as
equivalent.

### DIFF
--- a/ocaml-lsp-server/src/diagnostics.ml
+++ b/ocaml-lsp-server/src/diagnostics.ml
@@ -19,6 +19,46 @@ end
 
 module Dune = Stdune.Id.Make ()
 
+let equal_message =
+  (* because the compiler and merlin wrap messages differently *)
+  let is_space = function
+    | ' '
+    | '\r'
+    | '\n'
+    | '\t' ->
+      true
+    | _ -> false
+  in
+  let eat_space s i =
+    while !i < String.length s && is_space s.[!i] do
+      incr i
+    done
+  in
+  fun m1 m2 ->
+    let i = ref 0 in
+    let j = ref 0 in
+    try
+      eat_space m1 i;
+      eat_space m2 j;
+      while !i < String.length m1 && !j < String.length m2 do
+        let ci = m1.[!i] in
+        let cj = m2.[!j] in
+        if is_space ci && is_space cj then (
+          eat_space m1 i;
+          eat_space m2 j
+        ) else if Char.equal ci cj then (
+          incr i;
+          incr j
+        ) else
+          raise_notrace Exit
+      done;
+      eat_space m1 i;
+      eat_space m2 j;
+      (* we make sure that everything is consumed *)
+      !i = String.length m1 && !j = String.length m2
+    with
+    | Exit -> false
+
 type t =
   { dune :
       (Dune.t, (Drpc.Diagnostic.Id.t, Uri.t * Diagnostic.t) Table.t) Table.t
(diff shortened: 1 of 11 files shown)


rgrinberg:
ping @ulugbekna

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -1,5 +1,6 @@
 open Import
 module Version = Version
+module Diagnostics = Diagnostics
```
ulugbekna, line 3: I quite liked when you put for-test-only definitions in a separate module such as `Private` or `For_tests`. 

The same applies to `Diagnostics.equal_message` but it may get verbose `Ocaml_lsp_server.Private.Diagnostics.Private.equal_message` 🧐

ocaml-lsp-server/src/range.ml
```diff
@@ -6,6 +6,13 @@ let compare (x : t) (y : t) =
   | (Lt | Gt) as r -> r
   | Ordering.Eq -> Position.compare x.end_ y.end_
 
+let dyn_of_position { Position.line; character } =
+  Dyn.record [ ("line", Dyn.int line); ("character", Dyn.int character) ]
```
ulugbekna, line 10: Why don't we put this as `to_dyn` in `ocaml_lsp_serer/src/position.ml`?

ocaml-lsp-server/src/diagnostics.ml
```diff
@@ -19,6 +19,45 @@ end
...
+      eat_space m1 i;
+      eat_space m2 j;
+      while !i < String.length m1 && !j < String.length m2 do
+        let ci = m1.[!i] in
+        let cj = m2.[!j] in
+        eat_space m1 i;
+        eat_space m2 j;
+        if (is_space ci && is_space cj) || Char.equal ci cj then (
+          incr i;
+          incr j
```
ulugbekna, line 50: when we eat all space, we set `i` and `j` to non-space chars, but then we increment over them because in the if condition we check `ci` and `cj` which are values at old `i` and `j` index values. As a result, 

```ocaml
utop # equal_message "foo bar" "foo gar";;
- : bool = true
```

ocaml-lsp-server/src/diagnostics.ml
```diff
@@ -38,31 +77,60 @@ let create send ~workspace_root =
...
-        (match which with
-        | `All -> Uri_set.empty
-        | `One uri -> Uri_set.remove t.dirty_uris uri);
-      Table.foldi pending ~init:[] ~f:(fun uri diagnostics acc ->
-          (* we don't include a version because some of the diagnostics might
-             come from dune which reads from the file system and not from the
-             editor's view*)
-          PublishDiagnosticsParams.create ~uri ~diagnostics () :: acc)
-      |> t.send)
+let send =
```
ulugbekna, line 80: I will take a closer look at this function tomorrow if that's okay.

ulugbekna (changes requested):

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: Yeah, I like that convention as well for public libraries. But in this case, ocaml-lsp-server is only a library to make it possible to add tests. So it seems a little unnecessary in this case.

ocaml-lsp-server/src/diagnostics.ml
rgrinberg: Fixed. Thanks.

ulugbekna (approved):

PR #657, opened 2022-04-06
add leading slash to Uri.to_path on windows

tatchi:
@rgrinberg as discussed on Slack. I don't see any special behaviour for the path (but only `fsPath` and `Uri.file`) on windows in the `vscode-uri` test suite: https://github.com/microsoft/vscode-uri/blob/main/src/test/uri.test.ts#L156

So I assume that's fine to add a leading `/` on windows too. I can't test it though since I don't have a windows machine.

### DIFF
--- a/lsp/src/uri0.ml
+++ b/lsp/src/uri0.ml
@@ -46,10 +46,7 @@ let to_path t =
     |> String.replace_all ~pattern:"%3D" ~with_:"="
     |> String.replace_all ~pattern:"%3F" ~with_:"?"
   in
-  if Sys.win32 then
-    path
-  else
-    Filename.concat "/" path
+  Filename.concat "/" path
 
 let of_path (path : string) =
   let path = Uri_lexer.escape_path path in


rgrinberg (approved):
Looks good. Looks like there's still some other windows issues in CI. Do you mind taking a look?

tatchi:
That's the integration tests that are failing

```
Error: Program "yarn" not found!
```

From my understanding, that's because yarn is a shell script and not an executable. See related issue: https://github.com/ocaml/dune/issues/5401

So I assume we'll need for that issue to be resolved right?

rgrinberg:
> So I assume we'll need for that issue to be resolved right?

I would say we don't need to wait that far. We can have special logic for running yarn under dune on Windows in the Makefile. Instead of relying on `dune exec` looking up things correctly, can we just point to yarn in another way?

rgrinberg:
@tatchi do you plan to merge this PR?

tatchi:
> @tatchi do you plan to merge this PR?

Yes, I just wanted to try what you suggested in your last comment to fix the CI first. Feel free to merge if you want, I can do the CI part in another PR :)

PR #670, opened 2022-04-12
fix(lsp): encode all server requests correctly

rgrinberg:
* encode unknown requests
* do not provide null for missing parameters. Just omit the field
  altogether


rgrinberg:
@xldenis related fix.

Btw, I just noticed the naming inconsistency between `UnknownRequest` and `Unknown_notification`. I'll switch `Unknown_notification` to `UnknownNotification` for consistency.

PR #671, opened 2022-04-13
Workspace symbol wrong path

### LINKED ISSUE #656: Workspace symbols have wrong file paths
Repro: 

```
git clone git@github.com:ocaml/ocaml.git
./configure
dune build @libs
```

1. Open in VSCode
2. Go to symbol in workspace
3. Type 'ulambda' 
4. The list has an item in `clambda.ml`, but opening it fails with 'File not found'. 

The response from LSP indeed returns just `clambda.ml` instead of `middle_end/clambda.ml`:
```
[Trace - 21:56:41] Received response 'workspace/symbol - (82)' in 503ms.
Result: [
    {
        "name": "ulambda_switch",
        "kind": 15,
        "deprecated": false,
        "location": {
            "uri": "file:///[PATH]/ocaml/clambda.ml",
            "range": {
                "start": {
                    "line": 87,
                    "character": 0
                },
                "end": {
                    "line": 91,
                    "character": 37
                }
            }
        }
    },
```

At the same time "Go to type definition" works fine and finds the file under `middle_end` dir; so it seems that the problem is specific to workspace symbols query.

The dune file has `copy` rules, perhaps they confuse the LSP. I see that a similar issue was reported for 'go-to-def' in #470.

tatchi:
Should fix #656 

First commit adds a test that uses a `copy_files` stanza to reproduce the issue. The second commit implements the fix.

So instead of taking the filename from the `cmt.cmt_sourcefile`, we now use the `pos_fname` from the `Mbrowse_raw.node` location.

It actually fixes also issues with `.pp.ml` and `.mll` files:

**.pp.ml (BEFORE)**

https://user-images.githubusercontent.com/5595092/163256636-7b51aec2-4cf3-4f34-915c-931b88f07386.mov

**.pp.ml (AFTER)**

https://user-images.githubusercontent.com/5595092/163256128-dd5b3b82-c031-411c-ab53-19938d03b916.mov

**.mll (BEFORE)**

https://user-images.githubusercontent.com/5595092/163256838-8decce44-397b-4f90-9a2a-626e5e1c1137.mov

**.mll (AFTER)**

https://user-images.githubusercontent.com/5595092/163256916-b63f1441-ca0a-4922-a3d4-2372a3f4960e.mov


rgrinberg (approved):
LGTM. Could you add a changes entry?

cc @danidickstein your PR https://github.com/ocaml/merlin/pull/1219 solved a similar issue I believe

@voodoos could you have a look at this as well?

voodoos:
Looks like a reasonable change.

rgrinberg:
Merged manually.

PR #675, opened 2022-04-14
ci: fix detection of ocamlformat version

tatchi:
Fixes https://github.com/ocaml/ocaml-lsp/runs/6027736384?check_suite_focus=true

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ all:
 .PHONY: install-test-deps
 install-test-deps:
 	opam install 'menhir<20211230' cinaps 'ppx_expect>=v0.14.0' \
-		ocamlformat.$$(cat .ocamlformat | grep version | cut -d '=' -f 2) ocamlformat-rpc
+		ocamlformat.$$(awk -F = '/^version=/{print $$2}' .ocamlformat) ocamlformat-rpc
 
 .PHONY: dev
 dev: ## Setup a development environment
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -17,7 +17,7 @@ all:
 .PHONY: install-test-deps
 install-test-deps:
 	opam install 'menhir<20211230' cinaps 'ppx_expect>=v0.14.0' \
-		ocamlformat.$$(cat .ocamlformat | grep version | cut -d '=' -f 2) ocamlformat-rpc
+		ocamlformat.$$(awk -F = '/^version=/{print $$2}' .ocamlformat) ocamlformat-rpc
```
rgrinberg, line 20: Shouldn't the search condition be `$$1 == "version"`?  You're already tokenizing by the `=`, so might as well use the field names everywhere.

tatchi: I might misunderstand something, but since `$1` represents the entire line I don't think it works.

tatchi: The following seems to work:

```
awk -F '=' '$1=="version" {print $2}' .ocamlformat
```

Is that better ?

rgrinberg: It works for me:

```
$ awk -F = '$1 == "version" { print $2 }' .ocamlformat
0.21.0
```

tatchi: Ah yes you're right. I think I omitted the `-F` when trying it. I'll do the change, thanks :)

rgrinberg:
Thanks.

PR #678, opened 2022-04-16
Fix lev compilation flags

nojb:
Companion to https://github.com/ocaml/ocaml-lsp/commit/3e76bbb83a1dc6d0f102908c1bd5e6fabdc12f57 and rgrinberg/lev#3.


nojb:
Thanks!

PR #680, opened 2022-04-17
[Folding ranges]: traverse Pstr_extension

tatchi:


ocaml-lsp-server/src/folding_range.ml
```diff
@@ -185,15 +185,15 @@ let fold_over_parsetree (parsetree : Mreader.parsetree) =
...
       | Pstr_open _ ->
         Ast_iterator.default_iterator.structure_item self structure_item
       | Pstr_primitive _
       | Pstr_typext _
       | Pstr_exception _
       | Pstr_class_type _
       | Pstr_include _
-      | Pstr_attribute _
-      | Pstr_extension _ -> ()
+      | Pstr_attribute _ -> ()
```
rgrinberg, line 196: What's the logic for deciding for what gets traversed? For example, why don't class types get traversed but module types do?

tatchi: That's a good question, I don't think there's any logic. We just add them as we discover use cases. @ulugbekna raised a similar point in https://github.com/ocaml/ocaml-lsp/pull/593#pullrequestreview-863814306

Do you think it would make sense to traverse all variants by default (except maybe for the ones we know it doesn't make sense, e.g `Pexp_hole `) ?

rgrinberg: I would say it makes sense to fold anything that is over 3 lines. So attributes, class types are all fair game.

ulugbekna: I agree that we need to try switching to default iterator in cases perhaps making sure when we can that the AST node spans over >= 3 lines. Otherwise, we end up with cases that we don't handle: 

image

tatchi: Makes sense, I'll try to do that :) If you don't mind, I would prefer to address that in a separate PR

rgrinberg: Okay, but can you traverse class types as well?

rgrinberg (approved):

PR #687, opened 2022-04-21
use sexplib0 v0.14.0

tatchi:
Should fix #684

PR #688, opened 2022-04-21
use ppxlib 0.26.0 (alternative to #687)

tatchi:
Alternative to #687. This one make it so that sexplib0 `0.15.0` is installed on all platforms (instead of version `0.14.0` like in #687). 

Not sure which one is the best, the most important thing is that all platforms use the same version


tatchi:
Closed in favor of https://github.com/ocaml/ocaml-lsp/pull/687

PR #689, opened 2022-04-22
test: kill child process

tatchi:
That should fix the following warning:

```
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.
```

and fix tests not finishing in CI on Linux: https://github.com/ocaml/ocaml-lsp/runs/6089594745?check_suite_focus=true


tatchi:
> I don't think this is the right fix. We should figure out why the server isn't able to terminate cleanly instead of violently shutting it down.

Oh yes, you're right. I thought that was an issue with the way tests are written. I definitely went too fast, we should rather investigate why the server doesn't terminate properly.

Issue #690, opened 2022-04-22
Dune watch mode is disabled by default


edwintorok:
https://discuss.ocaml.org/t/ann-ocaml-lsp-server-1-11-0/9677 says that a new feature is integration with dune in watch mode.
However I keep getting this error:
`ocamllsp: -32600: No dune instance found. Please run dune in watch mode for /home/edwin/o/foo/lib/dune`

I *was* running dune in watch mode, but I've seen no attempts from `ocamllsp` to connect to it:
`(export DUNE_RPC_DEBUG=1; dune build --watch --no-buffer --verbose)`

However the RPC server seemed to be working fine (`dune rpc ping` and `dune rpc status` both worked and showed up in the console of dune watch.
Tried stracing `ocamllsp` on startup, and it made no attempt to search for, or connect to the dune `.rpc`.

Looking through the source code I found this:
https://github.com/ocaml/ocaml-lsp/commit/5d60767468207c628e3cfb1a40e6fa1b683c2a30
https://github.com/ocaml/ocaml-lsp/blob/1.11.2/ocaml-lsp-server/src/dune.ml#L714-L719
https://github.com/ocaml/ocaml-lsp/blob/6a38d66c57930c6ef5ac67d740a51e2d9551f93f/ocaml-lsp-server/src/import.ml#L226-L235

Which *only* enables Dune when inside test (so the 'reenable dune' commit actually does the opposite: it disables dune support because the default for `inside_test` is `false`).

I *can* test `ocamllsp` like this though (after patching `nvim-lsp-config` to recognize `dune` as a language):
`(export OCAMLLSP_TEST=true; vi lib/dune) `

I think perhaps the `reenable dune` commit meant to also delete the `inside_test` conditional, so something like this?
```diff
diff --git a/ocaml-lsp-server/src/dune.ml b/ocaml-lsp-server/src/dune.ml
index f4f9ded23..dbfa86aa3 100644
--- a/ocaml-lsp-server/src/dune.ml
+++ b/ocaml-lsp-server/src/dune.ml
@@ -713,10 +713,8 @@ let create workspaces (client_capabilities : ClientCapabilities.t) diagnostics
 
 let create workspaces (client_capabilities : ClientCapabilities.t) diagnostics
     progress document_store ~log =
-  if inside_test then
     create workspaces client_capabilities diagnostics progress document_store
       ~log
-  else ref Closed
 
 let run_loop t =
[...]

PR #691, opened 2022-04-22
Enable dune build --watch

edwintorok:
See https://github.com/ocaml/ocaml-lsp/issues/690.

Note that `dune runtest` is failing on `types.ml.cinaps-corrected`, but it was failing prior to my changes as well:
```
dune runtest 2>&1|diffstat
 types.ml.cinaps-corrected |41081 ++++++++++------------------------------------------------------------------------------------------------------
 1 file changed, 3802 insertions(+), 37279 deletions(-)
```
There was also a formatting difference causing a test failure (similarly failing before my change, so I added that here as well).
`make test-ocaml test-e2e` does pass though.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -35,6 +35,9 @@ possible and does not make any assumptions about IO.
   dyn
   yojson
   (ppx_yojson_conv_lib (>= "v0.14"))
+  (cinaps :with-test)
+  (menhir (>= 20211230 :with-test))
+  (ppx_expect (>= v0.14.0 :with-test))
   (uutf (>= 1.0.2))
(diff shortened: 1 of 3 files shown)


edwintorok:
Interesting, in the PR it fails with the reverse diff of what I just applied to fix the test error on my own machine, is this test OCaml version dependent?

rgrinberg:
> Note that dune runtest is failing on types.ml.cinaps-corrected, but it was failing prior to my changes as well:

We have a dedicated `make test` target that avoids all the cinaps stuff.

PR #692, opened 2022-04-26
[Folding ranges]: support more AST nodes

tatchi:


tatchi:
I tried to identify more use cases where folding would make sense (together with some tests to cover them). Hopefully it does make sense.

ocaml-lsp-server/src/folding_range.ml
```diff
@@ -16,6 +16,17 @@ let fold_over_parsetree (parsetree : Mreader.parsetree) =
...
+    let type_extension (_self : Ast_iterator.iterator)
+        (typ_ext : Parsetree.type_extension) =
+      let loc = typ_ext.ptyext_path.loc in
+      let last_constr = List.last typ_ext.ptyext_constructors in
+      let loc =
+        match last_constr with
+        | None -> loc
+        | Some { pext_loc; _ } -> { loc with loc_end = pext_loc.loc_end }
+      in
+      Range.of_loc loc |> push
```
rgrinberg, line 28: I didn't check this code very closely, but it seems strange to me that we always push to this stack but never pop. could you explain why that's correct?

ulugbekna: `Folding_range.fold_over_parsetree` initializes a new stack for each request. The stack is `rev_map`ed and returned at the end of the function. Am I missing something?

ulugbekna:
@rgrinberg sure

ulugbekna (approved):
LGTM, thanks @tatchi ! :-)

ulugbekna:
re: more visual tests

I'm all for it. Maybe we could have something like this, ie we use IDs to see start and end of folds: 

```
>1     match 
           Some  "large expr" 
         with 
>2     | None -> 
<2       () 
>3     | Some _ -> 
            print_endline "foo";
<1 <3   print_endline "bar"
```

PR #701, opened 2022-05-17
fix: stop leaking fd's

rgrinberg:


rgrinberg:
ping @ulugbekna

rgrinberg:
or @tatchi. your review is welcome as well if you're around

ulugbekna:
I’m at the offsite event (tarides). The changes look good to me but I hadn’t had opportunity to run the code. Sorry for the wait. I can definitely have a closer look and run code tomorrow.

ocaml-lsp-server/src/dune.ml
```diff
@@ -366,28 +366,40 @@ end = struct
...
-        let exn = List.hd exns in
-        Format.asprintf "unable to connect to dune %s@.%a"
-          (Registry.Dune.root source)
-          Exn_with_backtrace.pp_uncaught exn
+    | Error exns -> (
+      Lev_fiber.Fd.close sock;
+      let exns =
+        List.filter exns ~f:(fun exn ->
+            match exn with
+            | { Exn_with_backtrace.exn = Unix.Unix_error (ECONNREFUSED, _, _)
```
tatchi, line 386: For my own curiosity: why do we want to filter out this error?

ocaml-lsp-server/src/dune.ml
```diff
@@ -366,28 +366,40 @@ end = struct
...
-      Error ()
+      match exns with
+      | [] -> Fiber.return (Error ())
+      | _ ->
+        let message =
+          let exn = List.hd exns in
+          Format.asprintf "unable to connect to dune %s@.%a"
+            (Registry.Dune.root source)
+            Exn_with_backtrace.pp_uncaught exn
+        in
```
tatchi, line 399: nit:

```suggestion
   | exn :: _ ->
        let message =
          Format.asprintf "unable to connect to dune %s@.%a"
            (Registry.Dune.root source)
            Exn_with_backtrace.pp_uncaught exn
        in
```

tatchi:
Ah I guess I just need to update the dependencies to have the updated `fiber` version

tatchi:
> Ah I guess I just need to update the dependencies to have the updated `fiber` version

Yes, that fixed the issue 👍

tatchi (approved):

ocaml-lsp-server/src/dune.ml
rgrinberg: The error occurs when we fail to connect to a dune instance (because it's not running) usually. It occurs too frequently, so if we log it, we end up flooding the logs with this.

ocaml-lsp-server/src/dune.ml
rgrinberg: done

PR #703, opened 2022-05-25
chore(ci): Make e2e tests to run on windows

tatchi:
Currently, the e2e tests are failing to run on Windows:

```
Run opam exec -- make test-e2e
dune build @install && cd ocaml-lsp-server/test/e2e && dune exec -- yarn test
Entering directory 'D:\a\ocaml-lsp\ocaml-lsp'
Error: Program "yarn" not found!
make: *** [Makefile:51: test-e2e] Error 1
Error: Process completed with exit code 1.
```

The issue seems to be that `dune exec -- <foo>` fails to execute to <foo> whenever <foo> is a shell script. See https://github.com/ocaml/dune/issues/5401

That's why I wanted to experiment with using a binary to run the e2e tests. I first tried by using [dum](https://github.com/egoist/dum) which is an npm script runner written in Rust. You can see the changes in that commit (from another branch): https://github.com/tatchi/ocaml-lsp/commit/773472945222a90d168d9454bf89c8d0ef0426eb
It did work well as the e2e tests ran on Windows: https://github.com/tatchi/ocaml-lsp/runs/6567142311?check_suite_focus=true

Then I wanted to try if we could not just create an executable in the project itself that would run the e2e tests (and avoid relying on an external binary). This is what this PR is about.

I'm opening that PR to get your opinion about that approach. At least the e2e tests are now running on Windows so that's already better than what we have now 😄 Is it a good enough "temporary" solution?


rgrinberg:
The workaround looks good to me. But do we really need bos to run a shell script?

Makefile
```diff
@@ -17,7 +17,7 @@ all:
 .PHONY: install-test-deps
 install-test-deps:
 	opam install 'menhir>20211230' cinaps 'ppx_expect>=v0.14.0' 'sexplib0.v0.14.0' \
-		ocamlformat.$$(awk -F = '$$1 == "version" {print $$2}' .ocamlformat) ocamlformat-rpc
+		ocamlformat.$$(awk -F = '$$1 == "version" {print $$2}' .ocamlformat) ocamlformat-rpc bos
```
rgrinberg, line 20: still some bos left overs here

tatchi: Thanks, I was about the remove the remaining leftovers, should be ok now :) I also renamed the executable to `run_test_e2e`

rgrinberg (approved):

PR #705, opened 2022-05-30
Restore reading merlin files behind a CLI flag

anmonteiro:
- Currently LSP Server needs to be started with `--read-dot-merlin`.

64c0580267c58f6f4246d30b3c3515f13d2a4f96 implements falling back to `.merlin` iff `get_config` didn't find a useful Merlin configuration, and the `--read-dot-merlin` flag was passed to the CLI.

For the fallback, I'm using `Stdlib.(=)` to compare the config against `Config.empty`. If you prefer a designated equality function, I can also write one.

### DIFF
--- a/ocaml-lsp-server/bin/main.ml
+++ b/ocaml-lsp-server/bin/main.ml
@@ -4,7 +4,7 @@ let () =
   let read_dot_merlin = ref false in
   Arg.parse
     [ ("--version", Arg.Set version, "print version")
-    ; ( "--read-dot-merlin"
+    ; ( "--fallback-read-dot-merlin"
       , Arg.Set read_dot_merlin
       , "read Merlin config from .merlin files. The `dot-merlin-reader` \
          package must be installed" )
--- a/ocaml-lsp-server/bin/main.ml
+++ b/ocaml-lsp-server/bin/main.ml
@@ -1,8 +1,13 @@
 let () =
   Printexc.record_backtrace true;
   let version = ref false in
+  let read_dot_merlin = ref false in
   Arg.parse
-    [ ("--version", Arg.Set version, "print version") ]
+    [ ("--version", Arg.Set version, "print version")
+    ; ( "--read-dot-merlin"
+      , Arg.Set read_dot_merlin
+      , "read Merlin config from .merlin files" )
+    ]
     (fun _ -> raise (Arg.Bad "anonymous arguments are not accepted"))
     "ocamllsp";
   let version = !version in
@@ -11,7 +16,10 @@ let () =
     print_endline version
   else
     let module Exn_with_backtrace = Stdune.Exn_with_backtrace in
-    match Exn_with_backtrace.try_with Ocaml_lsp_server.run with
+    match
+      Exn_with_backtrace.try_with
+        (Ocaml_lsp_server.run ~read_dot_merlin:!read_dot_merlin)
+    with
     | Ok () -> ()
     | Error exn ->
       Format.eprintf "%a@." Exn_with_backtrace.pp_uncaught exn;
--- a/ocaml-lsp-server/bin/main.ml
+++ b/ocaml-lsp-server/bin/main.ml
@@ -6,7 +6,8 @@ let () =
     [ ("--version", Arg.Set version, "print version")
     ; ( "--read-dot-merlin"
       , Arg.Set read_dot_merlin
-      , "read Merlin config from .merlin files" )
+      , "read Merlin config from .merlin files. The `dot-merlin-reader` \
+         package must be installed" )
     ]
     (fun _ -> raise (Arg.Bad "anonymous arguments are not accepted"))
     "ocamllsp";
--- a/ocaml-lsp-server/src/merlin_config.ml
+++ b/ocaml-lsp-server/src/merlin_config.ml
@@ -207,10 +207,11 @@ module Dot_protocol_io =
       let write t x = write t (Some [ x ])
     end)
 
+let should_read_dot_merlin = ref false
+
 type db =
   { running : (string, entry) Table.t
   ; pool : Fiber.Pool.t
-  ; read_dot_merlin : bool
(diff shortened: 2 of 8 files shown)


anmonteiro:
Thanks, @laaksonel, fiexd in https://github.com/ocaml/ocaml-lsp/commit/64c0580267c58f6f4246d30b3c3515f13d2a4f96

ocaml-lsp-server/src/state.mli
```diff
@@ -32,6 +32,7 @@ val create :
   -> diagnostics:Diagnostics.t
   -> symbols_thread:Lev_fiber.Thread.t Lazy_fiber.t
   -> wheel:Lev_fiber.Timer.Wheel.t
+  -> read_dot_merlin:bool
```
rgrinberg, line 35: Could you make this a global variable in `Merlin_config` so that we don't need to pass this everywhere?

ocaml-lsp-server/bin/main.ml
```diff
@@ -1,8 +1,13 @@
 let () =
   Printexc.record_backtrace true;
   let version = ref false in
+  let read_dot_merlin = ref false in
   Arg.parse
-    [ ("--version", Arg.Set version, "print version") ]
+    [ ("--version", Arg.Set version, "print version")
+    ; ( "--read-dot-merlin"
+      , Arg.Set read_dot_merlin
+      , "read Merlin config from .merlin files" )
```
rgrinberg, line 9: should also mention that this needs the `dot-merlin-reader` package installed I believe.

rgrinberg:
> get_config didn't find a useful Merlin configuration, and the --read-dot-merlin flag was passed to the CLI.

I find that a bit weird. If `--read-dot-merlin` is passed I'd expect it to be taking priority. Perhaps it should be `--fallback-read-dot-merlin` if you want it to be a backup.

ocaml-lsp-server/src/merlin_config.ml
```diff
@@ -390,16 +391,32 @@ let config (t : t) : Mconfig.t Fiber.t =
           use_entry entry
     in
     let+ dot, failures = get_config entry.process ~workdir:ctx.workdir t.path in
-    let merlin = Config.merge dot t.initial.merlin failures config_path in
-    Mconfig.normalize { t.initial with merlin }
+
+    if dot <> Config.empty then
```
rgrinberg, line 395: The code path when `.merlin` is disabled shouldn't have this check at all. I.e.

```
let merlin =
  if config_from_dot_merlin && dot = Config.empty then
     (* your stuff here *)
  else
     (* old code *)
in
Mconfig.normalize merlin
```

anmonteiro:
Thanks for the initial review, @rgrinberg. I'm working on addressing your comments.

I realize I added the fallback but didn't explain why:

- in projects with both e.g. Melange and a native backend, I expect the LSP to just work. My understanding is that LSP keeps a merlin config per source file. The fallback would allow server-side files to use the RPC call, while the Melange files would be using the `.merlin` config. Does that make sense?

anmonteiro:
Just addressed the code review items, too.

ocaml-lsp-server/src/merlin_config.ml
```diff
@@ -389,9 +391,22 @@ let config (t : t) : Mconfig.t Fiber.t =
...
+    let config_from_dot_merlin =
+      if !should_read_dot_merlin then
+        Some (Mconfig.get_external_config t.path t.initial)
+      else None
+    in
     let+ dot, failures = get_config entry.process ~workdir:ctx.workdir t.path in
-    let merlin = Config.merge dot t.initial.merlin failures config_path in
-    Mconfig.normalize { t.initial with merlin }
+
+    match config_from_dot_merlin with
```
rgrinberg, line 401: I think my original suggestion avoided running `Mconfig.get_external_config` where it wasn't unnecessary. As in the clause where `dot <> Config.empty`.

anmonteiro: My bad, I understand what you mean now.

anmonteiro: Just pushed a commit implementing your suggestion.

rgrinberg (approved):
Looks good. Could you update CHANGES?

rgrinberg:
cc @gares

anmonteiro:
> Looks good. Could you update CHANGES?

Thanks! Sure thing.

One last question: What are your thoughts on doing this behavior by default and switching the flag to _disable_ it instead?

rgrinberg:
I'd prefer not to. I get a lot of issues from people with stale merlin files. Perhaps if you can change the code to only read .merlin files from  _build/, it would be acceptable.

anmonteiro:
That's fair. I'll be happy to come up with a detailed proposal in a subsequent PR.

ocaml-lsp-server/src/merlin_config.ml
```diff
@@ -390,8 +392,12 @@ let config (t : t) : Mconfig.t Fiber.t =
           use_entry entry
     in
     let+ dot, failures = get_config entry.process ~workdir:ctx.workdir t.path in
-    let merlin = Config.merge dot t.initial.merlin failures config_path in
-    Mconfig.normalize { t.initial with merlin }
+
+    if !should_read_dot_merlin && dot = Config.empty then
+      Mconfig.get_external_config t.path t.initial
```
rgrinberg, line 397: Is it ok not to normalize here?

anmonteiro: I looked at the [source](https://github.com/rgrinberg/merlin/blob/e757fe8f9276c87154356171959fabc806f4488e/src/kernel/mconfig.ml#L232) and `get_external_config` already calls normalize.

rgrinberg:
Merged, thanks.

gares:
Thanks!

PR #706, opened 2022-05-30
Fix crash in the Merlin pipeline

anmonteiro:
[This line](https://github.com/rgrinberg/merlin/blob/e757fe8f9276c87154356171959fabc806f4488e/src/extend/extend_driver.ml#L45) in the Merlin Extend driver (used by Reason's `ocamlmerlin-reason`) can throw a `Unix.Error (ECHILD, _,_)`.

This fix avoids crashing the OCaml LSP Server when that happens.

### DIFF
--- a/ocaml-lsp-server/src/document.ml
+++ b/ocaml-lsp-server/src/document.ml
@@ -143,7 +143,14 @@ let with_pipeline (t : t) f =
     let* pipeline = Lazy_fiber.force t.pipeline in
     let* task =
       Lev_fiber.Thread.task t.merlin ~f:(fun () ->
-          Mpipeline.with_pipeline pipeline (fun () -> f pipeline))
+          match
+            Exn_with_backtrace.try_with (fun () ->
+                Mpipeline.with_pipeline pipeline (fun () -> f pipeline))
+          with
+          | Ok ret -> ret
+          | Error exn ->
+            Code_error.raise "Document.merlin"
+              [ ("exn", Exn_with_backtrace.to_dyn exn) ])
     in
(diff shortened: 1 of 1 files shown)


anmonteiro:
This just fixes the symptom of what I'm seeing. It might not be related to LSP. 

Though I'm not sure what the behavior should be here. Clearly, `Mpipeline.with_pipeline` can throw, and that's not currently being handled.

rgrinberg:
> [This line](https://github.com/rgrinberg/merlin/blob/e757fe8f9276c87154356171959fabc806f4488e/src/extend/extend_driver.ml#L45) in the Merlin Extend driver (used by Reason's ocamlmerlin-reason) can throw a Unix.Error (ECHILD, _,_).

That's indeed suspicious. The most reliable way to fix this is to rewrite the driver to use `Lev_fiber` and use Async IO and `Lev_fiber.waitpid`.

I'm not even sure how this is fixing anything. It seems like you're just wrapping the exception with Code_error?

anmonteiro:
> That's indeed suspicious. 

I believe it's some PPX doing unspeakable things (e.g. including Marshaled output in `external` payload) and killing whatever transport it passes through here. Still trying to track that down.

> The most reliable way to fix this is to rewrite the driver to use Lev_fiber and use Async IO and Lev_fiber.waitpid.

Where? In the Merlin reader code?

rgrinberg:
> Where? In the Merlin reader code?

That's right. We have a merlin fork where we apply changes like that.

rgrinberg:
See #735

PR #707, opened 2022-05-31
fix: allow cancelling formatting requests

rgrinberg:
@tatchi do you mind adding cancellation in other slow operations as well?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Features
 
+- Allow cancellation of formatting requests (#707)
+
 - Add `--fallback-read-dot-merlin` to the LSP Server (#705). If `ocamllsp` is
   started with this new flag, it will fall back to looking for Merlin
   configuration in `.merlin` files rather than calling `dune ocaml-merlin`.
--- a/ocaml-lsp-server/src/ocamlformat.ml
+++ b/ocaml-lsp-server/src/ocamlformat.ml
@@ -21,6 +21,11 @@ let run_command prog stdin_value args : command_result Fiber.t =
       Unix.close stdin_i;
       Unix.close stdout_o;
       Unix.close stderr_o;
+      let* () =
+        Server.on_cancel (fun () ->
+            Unix.kill (Pid.to_int pid) Sys.sigint;
+            Fiber.return ())
+      in
       let blockity =
         if Sys.win32 then `Blocking
         else (
@@ -114,6 +119,13 @@ let exec bin args stdin =
   let+ res = run_command refmt stdin args in
   match res.status with
   | Unix.WEXITED 0 -> Result.Ok res.stdout
+  | Unix.WSIGNALED s
+    when s = Sys.sigint
+         (* TODO we should really make sure the cancellation was triggered *) ->
+    let e =
+      Jsonrpc.Response.Error.make ~code:RequestCancelled ~message:"cancelled" ()
+    in
+    raise (Jsonrpc.Response.Error.E e)
   | _ -> Result.Error (Unexpected_result { message = res.stderr })
 
 let run doc : (TextEdit.t list, error) result Fiber.t =


tatchi:
Do you know how we can manually test that code? I.e how can we cancel the `format` request?

> @tatchi do you mind adding cancellation in other slow operations as well?

Happy to try at least :) Do you have any operation in mind I could start with (something that shouldn't be too hard ideally)?

tatchi:
Do you know how we can manually test that code? I.e how can we cancel the `format` request?

> @tatchi do you mind adding cancellation in other slow operations as well?

Happy to try at least :) Do you have any operation in mind I could start with (something that shouldn't be too hard ideally)?

ulugbekna:
> how can we cancel the format request?

maybe if you edit the document before formatting response is returned for it, the request will get cancelled. Formatting of large files (eg lsp/types.ml) takes quite some time

tatchi:
I created a test that reproduces the issue with the `cancel token` that was not propagated in a `Reply.later`. I added it to the current test. Basically, two requests are now being sent, one that's cancelled and the other not. Let me know if you prefer them to be in separate tests.

I also made another commit that fixes the test, hopefully the right way.

lsp-fiber/src/rpc.ml
```diff
@@ -190,19 +184,31 @@ struct
...
+            Table.remove t.pending req.id;
+            Jsonrpc_fiber.Reply.now (to_response r)
+          | Reply.Later k ->
+            let f send =
+              Fiber.finalize
+                (fun () ->
+                  Fiber.Var.set cancel_token cancel (fun () ->
+                      k (fun r -> send (to_response r))))
+                ~finally:(fun () ->
+                  Table.remove t.pending req.id;
```
tatchi, line 208: Removing that line here and also in the `Reply.now` case still makes the tests pass. Should we try to have a test to cover that we correctly remove the token from the `pending` table? If so, can we do it easily with the current public API?

lsp-fiber/src/rpc.ml
```diff
@@ -190,19 +184,31 @@ struct
...
+        let cancel = Fiber.Cancel.create () in
         let+ response, state =
-          let cancel = Fiber.Cancel.create () in
-          Fiber.finalize
-            (fun () ->
-              Fiber.Var.set cancel_token cancel (fun () ->
-                  Table.replace t.pending req.id cancel;
-                  h_on_request.on_request t r))
-            ~finally:(fun () ->
-              Table.remove t.pending req.id;
```
rgrinberg, line 231: I think this cleanup is still needed if `h.on_request.on_request` raises instead of producing a value `Now` or `Later`.

We need to wrap this with a `Fiber.map_reduce_errors` and clean up `t.pending` in case of an error.

lsp-fiber/test/lsp_fiber_test.ml
```diff
@@ -104,21 +130,42 @@ module End_to_end_server = struct
...
+        match command with
+        | "cmd_cancel" ->
+          let+ () = Lev_fiber.Timer.sleepf 0.2 in
+          ( Rpc.Reply.later (fun k ->
+                let* cancel = Rpc.Server.cancel_token () in
+                (* Make sure that we can access the cancel token in a Reply
+                   response *)
+                (match cancel with
+                | None -> assert false
+                | Some _ -> assert true);
```
rgrinberg, line 160: `assert (Option.is_some cancel);`

lsp-fiber/test/lsp_fiber_test.ml
```diff
@@ -104,21 +130,42 @@ module End_to_end_server = struct
...
+                k result)
+          , state )
+        | _ ->
+          (* Make sure that we can access the cancel token in a Now response *)
+          (match cancel with
+          | None -> assert false
+          | Some _ -> assert true);
```
rgrinberg, line 167: ditto

lsp-fiber/src/rpc.ml
rgrinberg: I don't how to do it. Even if we don't cleanup the table, all we're doing is wasting memory so it isn't the biggest deal. I'd say not to worry about it for now.

lsp-fiber/src/rpc.ml
tatchi: I didn't use the `map_reduce_errors` function because it returns a `result` type and I don't know what that function should return in case of an `Error`. Perhaps the proper thing to do is for this function to also return a `result` but then we would have to handle the `Error` case somewhere else and that would require more changes.

Instead, I used the `map_and_reraise` function to clean up the `t.pending` and just re-raise the exception.

Does that make sense?

tatchi: Ah interesting, I hadn't thought about that but it definitely makes sense. Thanks :)

lsp-fiber/src/rpc.ml
```diff
@@ -277,9 +277,10 @@ struct
       Fiber.Cancel.with_handler cancel
         ~on_cancel:(fun () -> on_cancel jsonrpc_req.id)
         (fun () ->
-          let+ resp =
-            Session.request_with_cancel (Fdecl.get t.session) cancel jsonrpc_req
+          let _, req_f =
```
tatchi, line 280: I'm not sure if we need to use and fire the `cancel` in the `on_cancel`? The `lsp-fiber` cancellation test works in both cases..

jsonrpc-fiber/test/jsonrpc_fiber_tests.ml
```diff
@@ -267,3 +267,97 @@ let%expect_test "test from jsonrpc_test.ml" =
...
+      (fun () ->
+        Fiber.fork_and_join_unit
+          (fun () -> Jrpc.stopped client)
+          (fun () -> Jrpc.stopped server))
+      (fun () -> all)
+  in
+  Fiber_test.test Dyn.opaque run;
+  (* Ensure that server still responds even if the request was cancelled.
+     Required by the lsp spec *)
+  List.rev !responses
```
tatchi, line 348: Is it the correct way of testing that?

lsp-fiber/src/rpc.ml
rgrinberg: Using the `cancel` will allow the request to proceed without waiting for a response from the server. So it's a matter if want to trust the server to send the RequestCancelled response or we just ignore it and move on. I elected to go for the conservative route here for now and just wait for the cancellation response.

PR #713, opened 2022-06-10
Add `positionEncoding` capability

TOTBWF:
## Patch Description

Prior to LSP 3.17, all positions were specified in UTF-16 codeunits. This was somewhat suboptimal, so version 3.17 of the protocol introduced the [PositionEncodingKind](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#positionEncodingKind) type, and the associated `positionEncodings` and `positionEncoding` client and server capabilities, resp.

This PR adds the requisite types and capabilities. I didn't want to open the can of worms that is actually _supporting_ these capabilities in the server implementations, so that's work for a different day 🙂 

##  Notes

I wasn't sure if `PositionEncodingKind.t` should be a string or a sum type. The spec suggests that it's some random string with some predefined constants, so I just followed that, but this does feel a bit clunky.

### DIFF
--- a/lsp/src/types.ml
+++ b/lsp/src/types.ml
@@ -12363,6 +12363,8 @@ module ClientCapabilities = struct
           [@default None] [@yojson_drop_default ( = )]
     ; markdown : MarkdownClientCapabilities.t Json.Nullable_option.t
           [@default None] [@yojson_drop_default ( = )]
+    ; positionEncodings : PositionEncodingKind.t list option
+          [@default None] [@yojson_drop_default ( = )]
     }
   [@@deriving_inline yojson] [@@yojson.allow_extra_fields]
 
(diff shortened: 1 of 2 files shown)


TOTBWF:
Oh jeez, I hadn't noticed that `ocp-indent` decided to format all of types.mli`, I'll fix that ASAP!

rgrinberg:
> I wasn't sure if PositionEncodingKind.t should be a string or a sum type. The spec suggests that it's some random string with some predefined constants, so I just followed that, but this does feel a bit clunky.

The way we represent other ad hoc constants like that is with:
```
type t =
  | Foo
  | Bar
  | Other of string
```
So let's use that here as well.

rgrinberg:
Thanks.

PR #720, opened 2022-06-13
Show parsed ocamlformat syntax error message

ulugbekna:
The biggest quality of life improvement in editing ocaml code on vscode achievable in 8 lines IMHO

before: 

image
image

after: 

image

---

Better solutions exist where we 
- add a custom notification if client supports to add a button to the pop-up to jump to the error 
- add a status bar item in vscode that shows the formatting didn't succeed and clicking on the item, one can jump to the place of error

### DIFF
--- a/ocaml-lsp-server/src/ocamlformat.ml
+++ b/ocaml-lsp-server/src/ocamlformat.ml
@@ -77,7 +77,15 @@ let message = function
   | Unknown_extension uri ->
     Printf.sprintf "Unable to format. File %s has an unknown extension"
       (Uri.to_path uri)
-  | Unexpected_result { message } -> message
+  | Unexpected_result { message } -> (
+    match String.split message ~on:'\n' with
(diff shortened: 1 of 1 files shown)


rgrinberg:
It's an improvement, but it's a bit of a shame it has to be done this way. Have you considered the following solution:

* Set the env var `INSIDE_OCAMLLSP` when running ocamlformat
* Modify ocamlformat to output decent errors when this happens.

ulugbekna:
> It's an improvement, but it's a bit of a shame it has to be done this way. Have you considered the following solution:

Do you mean submitting a patch to ocamlformat to have better errors?

rgrinberg:
Yeah, exactly.

ulugbekna:
Is it worth merging this PR before we make a PR to ofmt and they release a new version? 

It's a big QoL improvement imho

rgrinberg:
Tbh, it never bothered me. I'm also worried that we're going to mangle some potential errors if they change some things. Since we need to support more than one version of ocamlformat.

PR #722, opened 2022-06-16
feature(jsonrpc): add batching to types

rgrinberg:


tatchi (approved):
Code looks good :) Would you mind giving some explanations as to why you added batching types? Is it something you're planning to implement soon?

rgrinberg:
I don't have anything in mind soon, but here's what I have in mind:

* It's good to have good error messages for a feature we explicitly don't support for now.
* LSP might add bathing support one day and this would be useful.
* Our jsonrpc library isn't all that LSP specific. It could be used in other contexts as well.

PR #723, opened 2022-06-16
test: add e2e test written in ocaml

rgrinberg:
to eventually move from typescript

### DIFF
--- a/ocaml-lsp-server/test/e2e-new/start_stop.ml
+++ b/ocaml-lsp-server/test/e2e-new/start_stop.ml
@@ -1,0 +1,120 @@
+open Stdune
+open Fiber.O
+
+let _PATH =
+  Bin.parse_path (Option.value ~default:"" @@ Env.get Env.initial "PATH")
+
+let bin = Bin.which "ocamllsp" ~path:_PATH |> Option.value_exn |> Path.to_string
+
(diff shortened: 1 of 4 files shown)


rgrinberg:
Honestly it's very hard to benchmark anything automatically. So unfortunately I don't see it helping much. Good profiling support is the way to go here.

ocaml-lsp-server/test/e2e-new/start_stop.ml
```diff
@@ -0,0 +1,128 @@
...
+    in
+    Fiber.fork_and_join_unit run_client (fun () ->
+        run
+        >>> Fiber.all_concurrently_unit
+              [ Lsp_fiber.Fiber_io.close io `Read
+              ; Lsp_fiber.Fiber_io.close io `Write
+              ; Client.stop client
+              ])
+  in
+  let waiting = ref true in
```
ulugbekna, line 63: `waiting` doesn't seem to be read anywhere?

ulugbekna (approved):

ocaml-lsp-server/test/e2e-new/start_stop.ml
rgrinberg: it's gone.

PR #725, opened 2022-06-16
Finish implementing the LSP 3.16 methods

ddickstein:
This commit extends the existing client/server request/notification
modules to support the all of the methods that have not yet been
implemented. This makes the LSP library usable for implementations other
than ocaml-lsp.


rgrinberg:
Cherry picked onto master after a few finishing touches. Thanks.

Issue #685, opened 2022-04-19
Support inlay hints


ulugbekna:
Inlay hints support was added to LSP 3.17:

- spec: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_inlayHint
- related issue in LSP repo: https://github.com/microsoft/language-server-protocol/issues/956

### What's an inlay hint? 

See https://code.visualstudio.com/docs/languages/typescript#_inlay-hints 

### What's in it for us? 

Having type annotations for local definitions as inlay hints would be very nice.

It's not an urgent but quite useful feature, IMO.

### References: 

- Rust analyzer PR switching to LSP inlay hints: https://github.com/rust-lang/rust-analyzer/pull/11935/files#diff-4bec6bfceb4323ffd0d2ff1ad25070234f084f31ddd5577bbdb510de878a3811

cemerick:
2¢ after seeing what the inlay hint support looks like via #1159:

* it really will be great to (probably only occasionally as needed) quickly toggle inlay hints as a way to peek at the types of *locals* without navigating about. Big :+1: there
* The loss of what I'll call "aggregated" top-level signature hints (as provided by code lenses [until recently :skull_and_crossbones:](#1134)) is a big blow IMO. Having now experienced not having those hints on top levels, I realize just how much I've come to rely on them, and I don't think that inlays will be a good replacement. Most of us are really well-trained to quickly grok function signatures (e.g. `foo -> bar -> string -> 'b` and so on), and disaggregating them makes inlays relatively less useful in that function context IMO.

Just as an example, the screenshot from @ https://github.com/ocaml/ocaml-lsp/pull/1159#issuecomment-1614773432 demonstrates what inlays will look like for a top-level function definition:

![image](https://github.com/ocaml/ocaml-lsp/assets/47489/cbe31c38-043a-49a7-8ebf-751808aa895c)

Comparing what needs to be parsed for the inlay approach (including all of the always-variable parameter names, etc):

```run rpc: State.t Server.t doc: Document.t ...```

seems obviously more difficult to parse than the always-familiar notation for OCaml function signatures, ```State.t Server.t -> Document.t -> ...```.

It feels like there's a best of both worlds to be had here: inlays for locals, and the familiar format and display for all function definitions (even local functions, perhaps using line-level lenses?).

(I guess that was more like $2 than 2¢, alas 😳)

jfeser:
[...]

PR #728, opened 2022-06-17
feature(lsp): add metamodel parser

rgrinberg:
In preparation for 3.17, we should be able to replace the typescript parser with a code generator that depends on this json file.

### DIFF
--- a/lsp/bin/dune
+++ b/lsp/bin/dune
@@ -15,9 +15,17 @@
  (modules gen)
  (libraries lsp_gen dyn stdune))
 
+(test
+ (name test_metamodel)
+ (modules test_metamodel metamodel)
+ (libraries stdune yojson)
+ (deps metaModel.json)
+ (action
+  (run ./test_metamodel.exe %{deps})))
+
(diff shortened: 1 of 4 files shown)


ulugbekna:
Interesting
1. is the plan to go from vscode.d.ts -> metamodel.json -> Metamodel.t -> generated ocaml code? 
2. I don't see code generation, will the ocaml code generation come in another PR? 
3. how do you generate this metamodel.json? 

Also, during my work on semantic highlighting I changed `data` from `int list` to `int array` because it's more efficient, especially when we compute diff between `data` from previous response and the currently computed `data`. I wonder whether this work is a good place to discuss when we should treat json array as a list and when as an array.

ulugbekna (approved):

ulugbekna:
I'd love to work on inlay hints when we upgrade to 3.17. I'd like to show types as inlay hints for symbols #685

rgrinberg:
Okay, let me start pushing the remaining of what I have.

PR #729, opened 2022-06-18
Add support for semantic highlighting

ulugbekna:
## Notes

There're still improvements to be made but wanted to get this public in case I'll get some feedback. 

Having some tests would be good. 

Would love some feedback on editing latency -- if you feel experience worsened. 

We can make this feature opt-in if people prefer that, at least until the feature is more stable. 

`Tokens` module could use a vector implementation rather than a `list ref` to be more efficient, I think. 

----

### Before 

image

image

### After 

image

image

----

### DIFF
--- a/ocaml-lsp-server/src/import.ml
+++ b/ocaml-lsp-server/src/import.ml
@@ -88,7 +88,12 @@ module Ast_iterator = Ocaml_parsing.Ast_iterator
 module Asttypes = Ocaml_parsing.Asttypes
 module Cmt_format = Ocaml_typing.Cmt_format
 module Ident = Ocaml_typing.Ident
-module Loc = Ocaml_parsing.Location
+
+module Loc = struct
+  include Ocaml_parsing.Location
+  include Ocaml_parsing.Location_aux
+end
+
 module Longident = Ocaml_parsing.Longident
 module Parsetree = Ocaml_parsing.Parsetree
 module Path = Ocaml_typing.Path
--- a/ocaml-lsp-server/src/import.ml
+++ b/ocaml-lsp-server/src/import.ml
@@ -214,6 +214,13 @@ include struct
   module RenameParams = RenameParams
   module SelectionRange = SelectionRange
   module SelectionRangeParams = SelectionRangeParams
+  module SemanticTokens = SemanticTokens
+  module SemanticTokensEdit = SemanticTokensEdit
+  module SemanticTokensLegend = SemanticTokensLegend
+  module SemanticTokensDelta = SemanticTokensDelta
+  module SemanticTokensDeltaParams = SemanticTokensDeltaParams
+  module SemanticTokensOptions = SemanticTokensOptions
+  module SemanticTokensParams = SemanticTokensParams
   module ServerCapabilities = ServerCapabilities
   module Server_notification = Lsp.Server_notification
   module SetTraceParams = SetTraceParams
--- a/ocaml-lsp-server/src/import.ml
+++ b/ocaml-lsp-server/src/import.ml
@@ -2,7 +2,29 @@
    listed alphabetically. Try to keep the order. *)
 include struct
   open Stdune
-  module Array = Array
+
+  module Array = struct
+    include Array
+    module View = Array_view
+
+    let split_into_subs arr ~at =
+      if at < 0 then
+        invalid_arg "split_into_subs: [~at] argument must be non-negative";
+      if at < Array.length arr then
+        let left = View.make arr ~pos:0 ~len:at () in
+        let right = View.make arr ~pos:at () in
+        (left, right)
+      else invalid_arg "split_into_subs: [~at] argument out of bounds of array"
+
+    let common_prefix_len ~equal (a : 'a array) (b : 'a array) : int =
+      let i = ref 0 in
+      let min_len = min (Array.length a) (Array.length b) in
+      while !i < min_len && equal (Array.get a !i) (Array.get b !i) do
+        incr i
+      done;
+      !i
+  end
+
   module Code_error = Code_error
   module Comparable = Comparable
   module Exn_with_backtrace = Exn_with_backtrace
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -86,6 +86,14 @@ let initialize_info : InitializeResult.t =
         ~commands:(view_metrics_command_name :: Dune.commands)
         ()
     in
+    let semanticTokensProvider =
+      Option.map (Sys.getenv_opt "OCAMLLSP_SEMANTIC_HIGHLIGHTING") ~f:(fun v ->
+          let delta = String.equal v "full/delta" in
+          let full = `Full (SemanticTokensOptions.create_full ~delta ()) in
+          `SemanticTokensOptions
+            (SemanticTokensOptions.create ~legend:Semantic_highlighting.legend
+               ~full ()))
+    in
     ServerCapabilities.create ~textDocumentSync ~hoverProvider:(`Bool true)
       ~declarationProvider:(`Bool true) ~definitionProvider:(`Bool true)
       ~typeDefinitionProvider:(`Bool true) ~completionProvider
@@ -94,7 +102,8 @@ let initialize_info : InitializeResult.t =
       ~documentFormattingProvider:(`Bool true)
       ~selectionRangeProvider:(`Bool true) ~documentSymbolProvider:(`Bool true)
       ~workspaceSymbolProvider:(`Bool true) ~foldingRangeProvider:(`Bool true)
-      ~experimental ~renameProvider ~workspace ~executeCommandProvider ()
+      ?semanticTokensProvider ~experimental ~renameProvider ~workspace
+      ~executeCommandProvider ()
   in
   let serverInfo =
     let version = Version.get () in
@@ -771,6 +780,8 @@ let on_request :
       ; (Req_infer_intf.meth, Req_infer_intf.on_request)
       ; (Req_typed_holes.meth, Req_typed_holes.on_request)
       ; (Req_wrapping_ast_node.meth, Req_wrapping_ast_node.on_request)
+      ; ( Semantic_highlighting.Debug.meth_request_full
+        , Semantic_highlighting.Debug.on_request_full )
       ]
       |> List.assoc_opt meth
     with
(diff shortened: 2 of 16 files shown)


ocaml-lsp-server/src/semantic_highlighting.ml
```diff
@@ -0,0 +1,828 @@
...
+
+(* TODO:
+
+   - [x] support textDocument/semanticTokens/full
+
+   - [x] support textDocument/semanticTokens/full/delta
+
+   - [ ] support textDocument/semanticTokens/range *)
+
+module Token_type = struct
```
rgrinberg, line 12: Shouldn't this be in `Lsp.Types`?

ocaml-lsp-server/src/semantic_highlighting.ml
```diff
@@ -0,0 +1,828 @@
...
+    State.t -> SemanticTokensParams.t -> SemanticTokens.t option Fiber.t =
+ fun state params ->
+  Fiber.of_thunk (fun () ->
+      let store = state.store in
+      let uri = params.textDocument.uri in
+      let doc = Document_store.get store uri in
+      let+ tokens =
+        Document.with_pipeline_exn doc (fun pipeline ->
+            let parsetree = Mpipeline.reader_parsetree pipeline in
+            let module Fold = Parsetree_fold () in
```
rgrinberg, line 748: You don't want to do unnecessary work in the merlin thread. The fold should be safe outside as the parsetree is immutable.

ulugbekna: Right, forgot to fix this after prototyping. Fixed now.

ocaml-lsp-server/src/semantic_highlighting.ml
ulugbekna: It could be but given the design requirements for token types, I'm not sure that's achievable in a desired manner.

- A client should be able to add their own token types
- Encoding of a token type (into an `int`) happens based on the position of the token type represented in the token type legend that the server sends to client

I'm not sure how to add token types to `Lsp.Types` such that they'd be useful

ocaml-lsp-server/src/document.mli
```diff
@@ -45,6 +45,10 @@ val source : t -> Msource.t
 
 val with_pipeline_exn : t -> (Mpipeline.t -> 'a) -> 'a Fiber.t
 
+(** [get_parsetree_exn doc] returns a fiber with [doc]'s parsetree. Keep in
+    mind, this function is run on the merlin thread *)
+val get_parsetree_exn : t -> Mreader.parsetree Fiber.t
```
rgrinberg, line 50: This function is one line long, only used once, and doesn't offer much abstraction. I'd just inline it in the caller.

ocaml-lsp-server/src/semantic_highlighting.ml
rgrinberg: > A client should be able to add their own token types

They can always do so by adding their own token type:

```
type my_token = Standard of Lsp.token | Mine of token
```

> Encoding of a token type (into an int) happens based on the position of the token type represented in the token type legend that the server sends to client

I suppose we don't need to add the encoding function then.

> I'm not sure how to add token types to Lsp.Types such that they'd be useful

Yes, it's only a little bit of value. Better than nothing I suppose.

ocaml-lsp-server/src/import.ml
```diff
@@ -2,7 +2,82 @@
...
+        let left = Sub.make arr ~start:0 ~length:at () in
+        let right = Sub.make arr ~start:at () in
+        (left, right)
+      else invalid_arg "split_into_subs: [~at] argument out of bounds of array"
+
+    let common_prefix_len (a : 'a array) (b : 'a array) : int =
+      let i = ref 0 in
+      while
+        !i < Array.length a
+        && !i < Array.length b
```
rgrinberg, line 73: A little cleaner would be to compute `min (length a) (length b)` and only compared it to that.

ocaml-lsp-server/src/import.ml
```diff
@@ -2,7 +2,82 @@
...
+        && Array.get a !i = Array.get b !i
```
rgrinberg, line 74: Let's avoid polymorphic equality.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -70,6 +70,13 @@ let initialize_info : InitializeResult.t =
     let executeCommandProvider =
       ExecuteCommandOptions.create ~commands:Dune.commands ()
     in
+    let semanticTokensProvider =
+      let full =
+        `Full (Lsp.Types.SemanticTokensOptions.create_full ~delta:true ())
+      in
+      Lsp.Types.SemanticTokensOptions.create
```
rgrinberg, line 77: We shouldn't need the `Lsp.Types` qualifier here.

ocaml-lsp-server/src/import.ml
```diff
@@ -2,7 +2,82 @@
    listed alphabetically. Try to keep the order. *)
 include struct
   open Stdune
-  module Array = Array
+
+  module Array = struct
```
rgrinberg, line 6: Could you move this to a separate file?

ocaml-lsp-server/src/semantic_highlighting.ml
```diff
@@ -0,0 +1,827 @@
...
+end = struct
+  type token =
+    { start : Position.t
+    ; length : int
+    ; token_type : int
+    ; token_modifiers : int
+    }
+
+  type t =
+    { mutable tokens : token list
```
rgrinberg, line 164: how big are these token lists in practice? also, can they have duplicate token types?

ulugbekna: > how big are these token lists in practice?

For `document_store.ml` in this PR, the `data` array is ~2575 integers long. This means there are 515 tokens. 

> can they have duplicate token types? 

If you really mean token types, then yes, because almost every symbol whether it's a variable name or a function name gets a token type, so duplication is inevitable. Duplications I guess do happen but I don't know whether we could do some sort of caching. 

One perf improvement perhaps would be to not use a list of token modifiers but use raw integers hidden by opaque type, e.g., something like `[Definition; Abstract]` could become `Token_modifiers.(definition % abstract)` which behind the scenes would simply do `lor`s

rgrinberg: > One perf improvement perhaps would be to not use a list of token modifiers but use raw integers hidden by opaque type, e.g., something like [Definition; Abstract] could become Token_modifiers.(definition % abstract) which behind the scenes would simply do lors

That sounds like the right way to go to me.

ocaml-lsp-server/src/semantic_highlighting.ml
```diff
@@ -88,68 +88,93 @@ module Token_type = struct
...
-    | Readonly
-    | Static
-    | Deprecated
-    | Abstract
-    | Async
-    | Modification
-    | Documentation
-    | Default_library
+module Token_modifiers : sig
+  type t
```
rgrinberg, line 92: I would rather organize it like this:

```
module Token_modifier : sig
  type t
  ..
  module Set  : sig
    type elt := set
    type t
    val empty
    val add : t -> elt -> t
  end
end
```

ocaml-lsp-server/src/subarray.mli
```diff
@@ -0,0 +1,14 @@
+type 'a t
+
+val make : 'a array -> start:int -> ?length:int -> unit -> 'a t
```
rgrinberg, line 3: Could you could it `pos` and `len` as customary. Also, can we document the behavior when `len` is absent?

ocaml-lsp-server/src/semantic_highlighting.ml
ulugbekna: What you propose is more intuitive indeed but also more verbose. The current solution we have reminds of a monoid where one can `append` values to each other and we have an identity element `empty`. I can use `<+>` instead of `%%` for more monoid-like notation and maybe can add a comment. Can also rewrite to you version but seems a little verbose

rgrinberg: I looked at it again and you can leave it as is. I don’t see how one can do something wrong even if they mix up sets and their elements with the current interface. Can you switch %%% to `union` though? Verbose or not, I much prefer using words over operators.

ulugbekna:
Highlighting is mostly good. There are still several problems, which need to be fixed for the feature to become enabled by default: 

- there seems to be a bug with delta computation -- I see sometimes incorrect highlighting: 

image

- I also notice editing become quite lagging. This is fixed when I restart the server, which is weird. Is it delta computations acting up? 

- Add tests

- Add micro-benchmarks 

- Improve performance. I wonder whether we could cache parts of the tree and avoid them again

rgrinberg:
I'd say let's proceed as follows:

* Add some tests (in OCaml please)
* Clean up the history
* Hide the feature behind an environment variable

Then you can iterate on improving performance without endless rebasing and conflict resolution.

ulugbekna:
I need to add some tests. I will add some e2e tests, but I would also want to add expect tests that work on some internal data structures for semantic tokens because they contain non-delta symbol positions, which should be much easier to debug. My understanding is to expose those internal structures in `Ocaml_lsp_server` just as `Diagnostics.equal_message` and  add `ocaml-lsp-server/tests/semantic_highlighting_tests.ml`

ulugbekna:
Two interesting things about the new test: 

1. It's super hard to debug in case of failure, so it's definitely worth adding tests I mentioned in the comment above where we can have a better look at positions, symbols and their token info 
2. There's epipe error because of some operation sequence problem perhaps. Need to figure this out

rgrinberg:
Can we start with some smaller tests? Or is this the only source where you're able to reproduce the issue.

ulugbekna:
> Can we start with some smaller tests? Or is this the only source where you're able to reproduce the issue.

Oh, I haven't thought that the source size could be playing any role here, but indeed if I put `fiber.ml` sources, for example, or simply `let a = ()`, I don't have that error.

If I keep the `ocaml_lsp_server.ml` source but remove shutdown request (`let* () = Client.request client Shutdown in`), the error is gone. 

----

Unrelatedly, we forgot to come back to the promotion bug we discussed during the pair programming session.

https://user-images.githubusercontent.com/16353531/179863969-7aa91261-7829-4e7d-b15d-54e9973b3c0f.mov

rgrinberg:
There's a debug variable in lev that allows you debug these issues in an easier way: `LEV_DEBUG=1`.

I can also improve it to include which part of the stream failed to write.

ulugbekna:
> There's a debug variable in lev that allows you debug these issues in an easier way: LEV_DEBUG=1.

Passing it with `extra_env` doesn't seem to change the output

rgrinberg:
One second, let me push something to improve this. Note that this will simply add what was unflushed to the error.

Rudi.

rgrinberg:
In any case, don't let this issue block you. Just add some smaller tests that pass and this should be good.

ulugbekna:
there's one basic test; I'll try adding more but would prefer to merge this PR and changes to happen in following PRs

Merging is extra safe because the feature is behind a flag

rgrinberg:
CI is failing with a good error message now. The issue seems to be that you're dropping the diagnostics on the floor. I suggest you wait for the diagnostics to come before closing the connection.

ulugbekna:
Great, thanks!

ocaml-lsp-server/test/e2e-new/semantic_hl_data.ml
```diff
@@ -0,0 +1,95 @@
+let src0 =
+  {|open Import
```
rgrinberg, line 2: Could you just write some dummy code here instead? E.g. create some simple source that uses one of every syntactical construct. The problem with reusing actual source code here is that you will trip up grep for everyone.

ocaml-lsp-server/src/document.mli
rgrinberg: This function is still useless

rgrinberg (approved):

PR #730, opened 2022-06-19
feat(folding range): support `include`

ulugbekna:
example:

```ocaml
include struct
  let u = ()
end
```


tatchi (approved):

rgrinberg:
@ulugbekna could you update the change log?

PR #739, opened 2022-06-23
Improve uri parsing

tatchi:
This PR changes the Uri module implementation to match the "official" one from [vscode-uri](https://github.com/microsoft/vscode-uri/blob/main/src/uri.ts). I tried to keep the code as close as possible to the vscode-uri to avoid silly mistakes from my end. 

The first commit adds more tests taken from vscode-uri. The second commit is the one that changes the implementation.

Now that we have a "correct" implementation with more test cases, I'm happy to refactor it if you prefer. Keeping it closer to the base implementation might make updating it easier (although the base implementation doesn't seem to be changed often).

Four e2e tests are failing on windows on the master branch:

* FAIL __tests__/textDocument-diagnostics.ts
* FAIL __tests__/ocamllsp-switchImplIntf.ts
* FAIL __tests__/workspace-symbol.test.ts
* FAIL __tests__/textDocument-codeAction.test.ts

This PR fixes the first two. The workspace symbol test failing is probably due to how we print the path on the test side. It shouldn't be hard to fix.
For the code action, I'll have to investigate.


.github/workflows/build-and-test.yml
```diff
@@ -64,6 +64,7 @@ jobs:
 
       - name: Run the unit tests
         run: opam exec -- make test-ocaml
+        continue-on-error: true
```
tatchi, line 67: Should be removed. I added that to not being stuck by the unit test failing on windows

lsp/src/uri0.ml
```diff
@@ -1,31 +1,137 @@
...
+    else
+      let c0 = path.[0] in
+      let c1 = path.[1] in
+      let c2 = path.[2] in
+      if
+        c0 = '/'
+        && ((c1 >= 'A' && c1 <= 'Z') || (c1 >= 'a' && c1 <= 'z'))
+        && c2 = ':'
+      then
+        String.make 1 (Char.lowercase_ascii c1)
```
tatchi, line 65: Is it a good way to go from `char` to `string` ?

lsp/src/uri0.ml
```diff
@@ -1,31 +1,137 @@
...
+        ^ String.sub path ~pos:2 ~len:(String.length path - 2)
+      else path
+  in
+  if !Private.win32 then slash_to_backslash path else path
+
+let of_string s =
+  let re =
+    Re.Perl.re "^(([^:/?#]+?):)?(\\/\\/([^/?#]*))?([^?#]*)(\\?([^#]*))?(#(.*))?"
```
tatchi, line 73: I copied this regular expression from the upstream implementation. If you prefer, I can try to create a parse with menhir

rgrinberg: Why do we need menhir? It seems like it should be easy enough to convert this regex to ocamllex.

lsp/src/uri0.ml
```diff
@@ -1,31 +1,137 @@
...
+    |> Re.compile
+  in
+  let res = Re.exec re s in
+  let scheme = Re.Group.get_opt res 2 |> Option.value ~default:"file" in
+  let group re n = Re.Group.get_opt re n |> Option.value ~default:"" in
+  let authority = group res 4 |> Uri.pct_decode in
+  let path =
+    let path = group res 5 |> Uri.pct_decode in
```
rgrinberg, line 81: We are pulling the uri package only for percent decoding, right? Doesn't seem like it's worth it as we can do percent encoding quite easily ourselves.

lsp/src/uri0.ml
```diff
@@ -1,31 +1,137 @@
...
-         Buffer.add_string b s;
-         Buffer.add_char b ':');
-  Buffer.add_string b "//";
-  Buffer.add_string b authority;
-  if not (String.is_prefix path ~prefix:"/") then Buffer.add_char b '/';
-  Buffer.add_string b path;
-  Buffer.contents b
+  let res = ref "" in
+
+  if not (String.is_empty scheme) then res := scheme ^ ":";
```
rgrinberg, line 98: You should use buffers instead of concatenating to `res`.

lsp/src/uri0.ml
tatchi: No, it's also being used to `encode`/`decode` non-ASCII characters. For instance, when you convert a Uri to a string you should encode all non-ASCII characters.

```ocaml
Uri.of_path "/Users/jrieken/Code/_samples/18500/Mödel + Other Thîngß/model.js" |> Uri.to_string
```

If you don't encode, the string representation is: `file:///Users/jrieken/Code/_samples/18500/Mödel + Other Thîngß/model.js` which is not correct. It should be instead: `file:///Users/jrieken/Code/_samples/18500/M%C3%B6del%20%2B%20Other%20Th%C3%AEng%C3%9F/model.js`

tatchi: Ah sorry, I think I misunderstood your question. Yes, that's correct, it's only being used for percent encoding.

tatchi: Ok, so I see that there's a reference table that shows the percent encoded value for some characters: https://www.w3schools.com/tags/ref_urlencode.asp. It could be easily implemented without the `uri` package.

But what about characters that are outside of that table? For instance, with the `vscode-uri` package:

```js
URI.file("/Users/mod❤.js").toString() // file:///Users/mod%E2%9D%A4.js
```

lsp/src/uri0.ml
tatchi: Oh yes that's what I meant sorry I got confused

rgrinberg: The performance should be essentially the same. It's just that with ocamllex we don't need to read line noise and the capturing groups are named rather than numbered.

lsp/src/uri0.ml
```diff
@@ -1,31 +1,143 @@
...
+        let c1 = path.[1] in
+        let c2 = path.[2] in
+        if
+          c0 = '/'
+          && ((c1 >= 'A' && c1 <= 'Z') || (c1 >= 'a' && c1 <= 'z'))
+          && c2 = ':'
+        then (
+          Buffer.add_char buff (Char.lowercase_ascii c1);
+          Buffer.add_string buff
+            (String.sub path ~pos:2 ~len:(String.length path - 2)))
```
rgrinberg, line 75: You can use `Buffer.add_substring` here.

lsp/src/uri0.ml
```diff
@@ -1,31 +1,143 @@
...
+        else Buffer.add_string buff path);
+      Buffer.contents buff
+  in
+  if !Private.win32 then slash_to_backslash path else path
+
+let of_string s =
+  let Uri_lexer.{ scheme; authority; path } = Uri_lexer.of_string s in
+  let scheme = scheme |> Option.value ~default:"file" in
```
rgrinberg, line 83: This looks much better with ocamllex now. What about pushing this `scheme`, `authority`, and `path` post processing there as well?

lsp/src/uri0.ml
rgrinberg: > But what about characters that are outside of that table? For instance, with the vscode-uri package:

I would assume it's just a percent encoding of the special characters in a particular encoding. utf8 most likely?

rgrinberg:
Great PR @tatchi. Let's try to get this polished and release 1.12.3 with it. Users are going to be very appreciative of this.

lsp/src/uri_lexer.mll
```diff
@@ -1,38 +1,50 @@
...
-and uri1 scheme = parse
-| "//" ([^ '/']* as authority) { uri2 scheme authority lexbuf }
-| "" { uri2 scheme "" lexbuf }
-and uri2 scheme authority = parse
-| "" { { scheme ; authority ; path = path lexbuf } }
+and path = parse
+| "" { { scheme = "file"; authority = ""; path = "/" } }
+| "//" ([^ '/']* as authority) (['/']_* as path) { { scheme = "file"; authority; path } }
+| "//" ([^ '/']* as authority) { { scheme = "file"; authority; path = "/" } }
+| _* as path
```
rgrinberg, line 33: What do we need this case for?

tatchi: To cover the case when the string is not empty and doesn't start with `//`. That's the equivalent to https://github.com/ocaml/ocaml-lsp/pull/739/commits/786178025795fd09f95cde61a93e27fc185c05b4#diff-4e2de9533ce99abc9d651d3c346c050c2e1923251c6ad49528deb91519d8748fL48

Also, if I remove that line I get the following error:

```
File "lsp/src/uri_lexer.mll", line 49, character 0: syntax error.
```

Do I misunderstand something?

rgrinberg: no, I think you got it. Can you split it into two cases though? One with the `/` prefix and one without the `/`. Should remove the needed for any helper functions.

lsp/src/import.ml
```diff
@@ -18,6 +18,8 @@ module String = struct
 
   let index = index_opt
 
+  let is_empty s = length s = 0
```
rgrinberg, line 21: Btw, this function can also be written as `= ""` with equivalent performance. Additionally `not (is_empty s)` is just `<> ""`.

lsp/src/uri0.ml
```diff
@@ -1,31 +1,112 @@
...
+   supports scheme, authority and path. Query, port and fragment are not
+   implemented *)
+
 open Import
 
 module Private = struct
   let win32 = ref Sys.win32
 end
 
-type t = Uri_lexer.t =
```
rgrinberg, line 7: if you keep this type equality, you will not need to do the conversion from `Uri_lexer.t` to `Uri.t`

lsp/src/uri0.ml
```diff
@@ -1,31 +1,112 @@
...
+    | c -> c)
+
+let slash_to_backslash =
+  String.map ~f:(function
+    | '/' -> '\\'
+    | c -> c)
+
+let of_path path =
+  let path = if !Private.win32 then backslash_to_slash path else path in
+  let Uri_lexer.{ scheme; authority; path } = Uri_lexer.of_path path in
```
rgrinberg, line 30: As mentioned above, you won't need this back and forth conversion if you keep the equality in the type def.

lsp/src/import.ml
tatchi: Interesting, that's how I initially wrote it and then I saw the `stdune` implementation: https://github.com/ocaml/dune/blob/main/otherlibs/stdune/string.ml#L43

Good to know that it doesn't make any difference! :)

lsp/src/uri0.ml
tatchi: Nice! 👌

lsp/src/uri_lexer.mll
tatchi: Clever. Amazing how this `of_path` function is now way simpler! 🤩

tatchi:
I should have addressed all the comments. I still have to write a change entry and eventually add some tests.

rgrinberg (approved):
Looks good.

tatchi:
Closing it as it was committed in https://github.com/ocaml/ocaml-lsp/commit/3ea0bb30ba32bbf0b79934d06266b16f873169c1

PR #752, opened 2022-06-27
chore: document store state transitions

rgrinberg:

ps-id: ccc7c797-ef50-44f3-ace9-2adbacfbb62b


ulugbekna (approved):
thanks heaps!

PR #755, opened 2022-06-29
git subrepo pull submodules/lev

rgrinberg:
subrepo:
  subdir:   "submodules/lev"
  merged:   "043bd6a6"
upstream:
  origin:   "https://github.com/rgrinberg/lev"
  branch:   "master"
  commit:   "043bd6a6"
git-subrepo:
  version:  "0.4.1"
  origin:   "???"
  commit:   "???"

ps-id: 41887970-b768-4965-90f9-a0d90cc0a207


ulugbekna:
If this changes how we work with vendored dependencies, would you mind also changing the dev setup instructions in README? 

Also, what are the advantages for our use case of subrepos over submodules? I am so used to the current workflow and know what problems there are and how to pick specific changes and all.

rgrinberg:
Yeah, sure. I'll write something up. It's still a little experimental, but there's enough for us to at least give this a try. There's a few problems with submodules:

* Normal git operations become painful. Branching, bisecting etc.
* Tags are not as useful without tagging the corresponding submodule tags. Something that is quite easy to do.
* Code review with changes in vendored repos is hardly readable. Let's take this PR as an example. You can actually read the diff and get an idea for what changed.

All in all, there's little friction for day to day work. There's a bit more friction for when editing upstream repos, but that occurs much less often comparatively.

rgrinberg:
> If this changes how we work with vendored dependencies, would you mind also changing the dev setup instructions in README?

One of the nice things about this change is that casual developers don't need to worry about any of this at all. They can just clone the repo and immediately start working. The dev instructions will only be useful to people interested in making changes to upstream code.

PR #762, opened 2022-06-30
test: disable broken tests on windows

rgrinberg:

ps-id: 4c2ff6f5-c5ce-4352-9b13-5d2cfc106967


tatchi:
Thanks, that definitely makes sense.

Looks like we need to upgrade to `dune 3.0` though:

```
# Error: 'enabled_if' is only available since version 3.0 of the dune language.
```

tatchi (approved):

PR #765, opened 2022-07-01
fix(merlin): disable promotion rules

rgrinberg:
We can experiment with adding changes we don't plan to upstream directly in this repo now.


ulugbekna:
Ohhh, love this PR! it looks correct to me!

PR #771, opened 2022-07-04
test(uri): add an e2e test for uri

tatchi:
This adds an e2e test to demonstrate that paths with special characters (space, accents, ...) are not well handled. I used a `DidOpenTextDocumentNotification` but it doesn't matter, any request/notification that sends a `uri` will have the same issue.

### DIFF
--- a/ocaml-lsp-server/test/e2e-new/uri.ml
+++ b/ocaml-lsp-server/test/e2e-new/uri.ml
@@ -1,0 +1,44 @@
+open Test.Import
+
+let%expect_test "uri" =
+  let notif_received = Fiber.Ivar.create () in
+  let handler =
+    let on_notification _ (notification : Lsp.Server_notification.t) =
+      (match notification with
+      | PublishDiagnostics d ->
+        print_endline "client: received publish diagnostics notification";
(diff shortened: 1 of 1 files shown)


ocaml-lsp-server/test/e2e/__tests__/uri.test.ts
```diff
@@ -0,0 +1,48 @@
...
+        ),
+      );
+    openDocument(outdent`
+      let () =
+        let x = 123 in
+        ()
+    `);
+    const params = await receivedDiganostics;
+
+    expect(params.uri).toBe(uri);
```
tatchi, line 46: It fails with:

```js
Expected: "file:///src/n%C3%A9w/M%C3%B6del%20%2B%20Other%20Th%C3%AEng%C3%9F/test.ml"
Received: "file:///src/n%C3%A9w/M%C3%B6del %2B Other Th%C3%AEng%C3%9F/test.ml"
```

https://github.com/ocaml/ocaml-lsp/pull/739 should fix it but I'm not sure if we really need an e2e test for that since we already have OCaml tests for the URI parsing?

rgrinberg:
Please re-open if this still relevant

PR #773, opened 2022-07-05
feature(olsp): warning for unused symbols are marked using `DiagnosticTag.Unnecessary`

### LINKED ISSUE #754: Use `Unnecessary` Diagnostic tag for unused symbols
If you have an unused variable in python, you get 

image

We should also be able to do so by using `DiagnosticTag` `Unnecessary`, see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticTag

There're various unused diagnostics possible, see #700

ulugbekna:
image

image

Fixes #754

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -265,7 +265,10 @@ let set_diagnostics rpc doc =
                                  DiagnosticRelatedInformation.create ~location
                                    ~message)) )
                     in
-                    create_diagnostic ?relatedInformation ~range ~message
+                    let tags =
+                      Diagnostics.tags_of_message ~src:`Merlin message
+                    in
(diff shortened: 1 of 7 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -245,7 +245,12 @@ let set_diagnostics rpc doc =
                                  DiagnosticRelatedInformation.create ~location
                                    ~message)) )
                     in
-                    create_diagnostic ?relatedInformation ~range ~message
+                    let tags =
```
rgrinberg, line 248: can we keep the tags logic in one place? It will be a bit messy because the check is slightly different for dune and merlin (but we'll fix this soon enough)

rgrinberg (approved):
LGTM

rgrinberg:
@ulugbekna this needs a CHANGES entry.

PR #775, opened 2022-07-07
refactor(uri): get rid of uri dependency

tatchi:
Extracted the relevant `encode`/`decode` functions from `ocaml-uri`


rgrinberg:
Merged

PR #776, opened 2022-07-07
refactor(uri): refactor uri impl

tatchi:


lsp/src/uri0.ml
```diff
@@ -29,26 +29,22 @@ let of_path path =
   let path = if !Private.win32 then backslash_to_slash path else path in
   Uri_lexer.of_path path
 
+let is_letter c = (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z')
```
ulugbekna, line 32: just fyi, this could be `let is_letter = function 'a'..'z' | 'A'..'Z' -> true | _ -> false`

tatchi: Oh yes, I like it more. Thanks 🤗

lsp/src/uri0.ml
```diff
@@ -29,26 +29,22 @@ let of_path path =
...
     let len = String.length path in
     if len = 0 then "/"
     else
       let buff = Buffer.create 64 in
-      (if (not (String.is_empty authority)) && len > 1 && scheme = "file" then (
-       Buffer.add_string buff "//";
-       Buffer.add_string buff authority;
-       Buffer.add_string buff path)
+      (if (not (String.is_empty authority)) && len > 1 && scheme = "file" then
+       Buffer.add_string buff (Printf.sprintf "//%s%s" authority path)
```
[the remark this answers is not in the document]
rgrinberg: sprintf will allocate an intermediate string that will immediately be thrown away. it's less efficient for sure, but it's likely not to matter much.

rgrinberg:
Please re-open when the PR is updated.

PR #777, opened 2022-07-07
feat: allow cancellation of workspace symbols requests

tatchi:


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -660,17 +660,38 @@ let workspace_symbol server (state : State.t) (params : WorkspaceSymbolParams.t)
...
+            match cancel with
+            | None ->
+              fun f ->
+                let+ res = f () in
+                (res, Fiber.Cancel.Not_cancelled)
+            | Some token ->
+              let on_cancel () = Lev_fiber.Thread.cancel task in
+              fun f -> Fiber.Cancel.with_handler token ~on_cancel f
+          in
+          maybe_cancel @@ fun () -> Lev_fiber.Thread.await task
```
tatchi, line 680: Might be a bit too early to say but I've the impression that this code will be often used with cancellation. Would it make sense to extract it? Something like:

```ocaml
let make_with_cancellation cancel ~on_cancel =
  match cancel with
  | None ->
    fun f ->
      let+ res = f () in
      (res, Fiber.Cancel.Not_cancelled)
  | Some token ->
    fun f -> Fiber.Cancel.with_handler token ~on_cancel f
```

Which could be used like so:

```ocaml
let* res, cancel =
  match task with
  | Ok task ->
    let* cancel = Server.cancel_token () in
    let on_cancel () = Lev_fiber.Thread.cancel task in
    let with_cancellation = make_with_cancellation cancel ~on_cancel in
    with_cancellation @@ fun () -> Lev_fiber.Thread.await task
  | Error `Stopped -> Fiber.never
```

rgrinberg: If you can clean up the code with this then sure.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -660,17 +660,38 @@ let workspace_symbol server (state : State.t) (params : WorkspaceSymbolParams.t)
     let+ symbols_results =
       let task =
         Lev_fiber.Thread.task thread ~f:(fun () ->
+            let+ () = Fiber.return () in
```
rgrinberg, line 663: why do we need this line?

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -660,17 +660,38 @@ let workspace_symbol server (state : State.t) (params : WorkspaceSymbolParams.t)
...
+        | Ok task ->
+          let* cancel = Server.cancel_token () in
+          let maybe_cancel =
+            match cancel with
+            | None ->
+              fun f ->
+                let+ res = f () in
+                (res, Fiber.Cancel.Not_cancelled)
+            | Some token ->
+              let on_cancel () = Lev_fiber.Thread.cancel task in
```
rgrinberg, line 677: That looks good, but I think we can do even better. Why don't we check if the cancellation token has been fired before reading every cmt file? If it has been fired, we can terminate the work early.

ocaml-lsp-server/src/ocaml_lsp_server.ml
tatchi: I'll eventually do that in another PR :)

rgrinberg (approved):

rgrinberg:
Merged.

PR #790, opened 2022-07-16
fix(ocamllsp): make more errors non-fatal

rgrinberg:
ocamlformat-rpc, merlin, and various background errors would make the
server stop executing. After this commit, these errors will be logged
and the server will continue

ps-id: 01867186-666a-4e47-9ff9-baff803bf5ba

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -1021,11 +1021,45 @@ let start () =
     Fdecl.get server
   in
   let state = Server.state server in
+  let with_log_errors what f =
+    let+ (_ : (unit, unit) result) =
+      Fiber.map_reduce_errors
+        (module Monoid.Unit)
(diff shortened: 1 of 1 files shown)


ulugbekna (approved):

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -1021,11 +1021,46 @@ let start () =
...
+      Fiber.map_reduce_errors
+        (module Monoid.Unit)
+        f
+        ~on_error:(fun exn ->
+          Format.eprintf "%s: %a@." what Exn_with_backtrace.pp_uncaught exn;
+          Fiber.return ())
+    in
+    match res with
+    | Ok () -> ()
+    | Error () -> ()
```
ulugbekna, line 1035: this also looks very verbose for very innocuous thing, IMHO. I would create a util function in `Import.Fiber`

PR #795, opened 2022-07-20
feature: open related code actions

rgrinberg:
We can now open documents on the lsp side, which means that we can
implement ml/mli toggling entirely in LSP

- [x] Add a create .mli code action if one doesn't exist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Features
+
+- Code actions for jumping to related files (`.ml`, `.mli`, etc.) (#795)
+
 # 1.12.4
 
 - Allow cancellation of workspace symbols requests (#777)
--- a/ocaml-lsp-server/src/code_actions.ml
+++ b/ocaml-lsp-server/src/code_actions.ml
@@ -26,8 +26,8 @@ end
 
 let compute server (params : CodeActionParams.t) =
   let state : State.t = Server.state server in
+  let uri = params.textDocument.uri in
   let doc =
-    let uri = params.textDocument.uri in
     let store = state.store in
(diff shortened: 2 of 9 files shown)


rgrinberg:
ping @ulugbekna

ulugbekna:
Is it okay if I review it on Monday?

rgrinberg:
Sure, next week is fine.

Rudi.

ocaml-lsp-server/src/code_actions/open_related.mli
```diff
@@ -0,0 +1,7 @@
+open Import
```
ulugbekna, line 1: The current convention for code action-containing files is to have prefix `action_`. 

I find it verbose, but it makes looking the files up easier (cmd + p and `action_`). 

We can remove prefixes in other files or add it here.

ocaml-lsp-server/src/code_actions/open_related.ml
```diff
@@ -0,0 +1,45 @@
+open Import
+open Fiber.O
+
+let command_name = "ocamllsp/open-related-source"
+
+let command_run server (params : ExecuteCommandParams.t) =
+  let uri =
+    match params.arguments with
+    | None -> assert false
```
ulugbekna, line 9: more like `Invalid_argument` ? the same for `Some _`

ulugbekna (approved):
I like this PR a lot!

ocaml-lsp-server/src/code_actions/open_related.mli
rgrinberg: The directory already includes the suffix `_action`. that's probably enough. I won't touch the convention here though.

PR #827, opened 2022-08-01
feature(lsp_gen): use metamodel

rgrinberg:
ps-id: 4afba4fe-227c-4cc4-964d-f8ff3ca1c5d3


rgrinberg:
@ulugbekna this should be enough to get started with inlay hints.

PR #844, opened 2022-08-21
fix(ocamllex): do not raise exceptions on signatureHelp

edwintorok:
As soon as I open a .mll file (with filetype set to ocamllex in NeoVim) the LSP
server sends a continuous stream "Document.dune" exceptions.
I currently have to turn off Lsp (using :LspStop) to be able to keep
editing.

Instead of raising an exception just return an empty result for
filetypes where signatureHelp is not implemented yet.

It'd probably be better to not advertise the capability, but not sure if
that is doable by filetype...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@
 - Respect `showDocument` capabilities. Do not offer commands or code actions
   that rely on this request without client support. (#836)
 
+- Fix signatureHelp on .mll files: avoid "Document.dune" exceptions
+
 # 1.13.1
 
 ## Fixes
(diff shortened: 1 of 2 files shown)


edwintorok:
For some reason this is only a problem with `.mll` files, `.mly` files don't have this problem.

edwintorok:
FWIW editing this file using ocaml-lsp itself I found another bug: it continously raises some exceptions , e.g. on line 479, so there is probably more that could be done to make `signature_help` more robust against failures, but that is unrelated to fixing `ocamllex` here.

rgrinberg:
Thanks for the fix. Could you add a CHANGES entry?

PR #860, opened 2022-09-13
Use Merlin Enclosing command instead of Shape to answer SelectionRange queries 

voodoos:
The enclosing command is the dedicated Merlin command to provide meaningful growing and shrinking of the selection.

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -538,39 +538,19 @@ let selection_range (state : State.t)
   match Document.kind doc with
   | `Other -> Fiber.return []
   | `Merlin merlin ->
-    let selection_range_of_shapes (cursor_position : Position.t)
-        (shapes : Query_protocol.shape list) : SelectionRange.t option =
-      let rec ranges_of_shape parent s =
-        let selectionRange =
-          let range = Range.of_loc s.Query_protocol.shape_loc in
-          { SelectionRange.range; parent }
-        in
-        match s.Query_protocol.shape_sub with
-        | [] -> [ selectionRange ]
-        | xs -> List.concat_map xs ~f:(ranges_of_shape (Some selectionRange))
-      in
-      (* try to find the nearest range inside first, then outside *)
-      let nearest_range =
-        let ranges = List.concat_map ~f:(ranges_of_shape None) shapes in
-        List.min ranges ~f:(fun r1 r2 ->
-            let inc (r : SelectionRange.t) =
-              Position.compare_inclusion cursor_position r.range
-            in
-            match (inc r1, inc r2) with
-            | `Outside x, `Outside y -> Position.compare x y
-            | `Outside _, `Inside -> Gt
-            | `Inside, `Outside _ -> Lt
-            | `Inside, `Inside -> Range.compare_size r1.range r2.range)
-      in
-      nearest_range
+    let rec selection_range_of_enclosings = function
+      | hd :: tl ->
+        let parent = selection_range_of_enclosings tl in
+        Some { SelectionRange.range = Range.of_loc hd; parent }
+      | [] -> None
     in
     let+ ranges =
       Fiber.sequential_map positions ~f:(fun x ->
-          let+ shapes =
-            let command = Query_protocol.Shape (Position.logical x) in
+          let+ enclosings =
+            let command = Query_protocol.Enclosing (Position.logical x) in
             Document.Merlin.dispatch_exn merlin command
           in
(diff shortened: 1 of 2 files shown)


rgrinberg:
Does this still work when the module fails to type check?

Out of curiosity, what are "Shapes" useful for if not for this?

Please update CHANGES.

voodoos:
The "shape" command of Merlin has nothing to do with the "shapes" that we use to jump to definitions.
It is an old command that is not even use by the Emacs and Vim modes... In fact the vim mode even relies on the `type_enclosing` command instead of the more specialized one `enclosing`.

I just tried and it appears to work when the buffer does not type but I did not test it extensively.

rgrinberg:
I'm still curious why do we have Shape in merlin at all then?

By the way, I would expect expanding selection wouldn't always select an expression or a type. For example, given (bar signifies the cursor, `[` and `]` demonstrate the selection)

```
let ..|.. in
```

would expand to

```
let [ ... ] in
```

and then to

```
[let ... in]
```

I'm not sure the current shapes command respects that either, but it would be good if a proper solution for expand selection did.

rgrinberg:
Finally, does expand selection work inside comments?

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -518,40 +518,20 @@ let rename (state : State.t)
...
   let+ ranges =
     Fiber.sequential_map positions ~f:(fun x ->
-        let+ shapes =
-          let command = Query_protocol.Shape (Position.logical x) in
+        let+ enclosings =
+          let command = Query_protocol.Enclosing (Position.logical x) in
           Document.dispatch_exn doc command
         in
-        selection_range_of_shapes x shapes)
+        selection_range_of_enclosings enclosings)
```
ulugbekna, line 534: TODO(ulugbekna): we should batch these merlin queries into one

ulugbekna (approved):

ulugbekna:
> Finally, does expand selection work inside comments?

No, the only range returned is the whole document range, but, for example, vscode is smart about it and expands selection gradually: word, parenthesized expression, line, then the whole document. 

```
[Trace - 11:25:34 PM] Sending request 'textDocument/selectionRange - (400)'.
Params: {
    "textDocument": {
        "uri": "file:///Users/ulugbekna/code/ocaml-lsp/ocaml-lsp-server/src/semantic_highlighting.ml"
    },
    "positions": [
        {
            "line": 947,
            "character": 63
        }
    ]
}


[Trace - 11:25:34 PM] Received response 'textDocument/selectionRange - (400)' in 1ms.
Result: [
    {
        "range": {
            "end": {
                "character": 72,
                "line": 1027
            },
            "start": {
                "character": 0,
                "line": 0
            }
        }
    }
]
```

What would you expect from expanding selection in a comment? Do you mean the selections could work in pre-formatted code block within a comment `{| |}` ?

voodoos:
> I'm not sure the current shapes command respects that either, but it would be good if a proper solution for expand selection did.

Yes, both the `shapes`and `enclosing` commands use the same data which is the typedtree to provide the selections so the results should be mostly the same and in both case they only return selections that corresponds to valid typedtree nodes.

It looks like the kind of selection you describe would be more "parsetree" based.

> I'm still curious why do we have Shape in merlin at all then?

It looks like a refinement over the `outline` command that returns only the outline for the current branch of the tree. I guess it was meant to allow easy navigation to parent binders like (|⁰, |¹ are successive positions of the cursor):
```
|² let x = 3 in„
  |¹ let y = 4 in
     |⁰ x + y
```

rgrinberg:
> It looks like the kind of selection you describe would be more "parsetree" based.

That's exactly what I would expect for this command.

> This command can be used to assist navigation in a source code buffer. It returns a tree of all relevant locations around the cursor. It is similar to outline without telling any information about the entity at a given location.

At least from this description, it seems like Shape fits better fitted to play the role of "expand selection" than the enclosing type.

ulugbekna:
Do we plan to merge this? @rgrinberg 

Would be great to merge before there're more conflicts. I can fix the conflict if @voodoos is busy

voodoos:
Rebased.

> At least from this description, it seems like Shape fits better fitted to play the role of "expand selection" than the enclosing type.

Right, and it does look like the code on Merlin's side is doing some interesting selection to reduce noise, however the result are still bad. Maybe it is worth to have a look and fix it. But it is quite low on my priority list.

This PR would make the behavior consistent with how Merlin behave in Emacs in Vim, but this might not be "better". I will try to find the type to write some test that highlight differences between the two commands.

PR #870, opened 2022-10-11
Add extract code actions

jfeser:
Adds two new code actions: `Extract local` and `Extract function`. 

Extract local transforms:
```
let f x =  $x+1$ + 2
```
to 
```
let f x = 
  let new_var = x + 1 in 
  new_var + 2
```
The new variable gets placed at the tightest enclosing binder of the extracted expression.

 Extract function would transform the same code to:
```
let new_fun x = x + 1
let f x = new_fun x + 2
```
The new function will be inserted directly before the enclosing structure item. Any free variables in the expression will be passed as arguments.

Just wanted to get opinions on whether this is a useful feature, and whether these behaviors make sense.

TODO:
- Should extract local move the expression to the loosest binder rather than the tightest?
- Extract function doesn't handle free variables with nontrivial `Longident`s.
- Add tests.


rgrinberg:
The feature looks useful to me.

> Should extract local move the expression to the loosest binder rather than the tightest?

I would expect the tightest. But perhaps you could have two extract actions for the different behaviors.

Or perhaps we could have different refactorings for hoisting bindings up.

> The new function will be inserted directly before the enclosing structure item.

I would expect it to stay close to the scope where it's being used personally.

rgrinberg:
Btw, can you try splitting this code action into a discovery/resolve phase and evaluate if it helps performance a little. I'm getting a little worried about all these code actions clogging up the server.

jfeser:
I have a branch (https://github.com/jfeser/ocaml-lsp/tree/tracing) with some extra tracing. From the traces I've looked at, code actions aren't a performance concern. How have you been evaluating performance/where have you seen performance problems?

ocaml-lsp-server/test/e2e-new/code_actions.ml
```diff
@@ -599,3 +599,98 @@ let f (x : t) = x
...
+      cnum := 0)
+    else incr cnum
+  done;
+  Position.create ~character:!cnum ~line:!lnum
+
+let parse_selection src =
+  let start_pos =
+    match String.index src '$' with
+    | Some x -> x
+    | None -> failwith "expected a selection opening mark"
```
rgrinberg, line 619: Is it possible the user can encounter this error?

1. If yes, then it should be a Jsonrpc error.
2. if not, it should it be a code error.

jfeser: This is part of the test harness. It gets raised when the test input doesn't have $ markers for the selection. Do you want me to use a different exception?

rgrinberg: Oh never mind. It's fine.

rgrinberg (approved):

3Rafal:
@rgrinberg , unit tests for this PR are broken. CI fails on 4/7 checks

rgrinberg:
@jfeser could you take a look?

PR #873, opened 2022-10-13
Add Jest to the list of tools required for tests

cuihtlauac:


cuihtlauac:
Jest gets installed automatically. No manual operation required. No need to make it appear.

PR #878, opened 2022-10-15
Do not rely on global yarn

cuihtlauac:
Install yarn (https://yarnpkg.com) in ocaml-lsp-server/test/e2e using npm --no-save --no-package-lock. Change Makefile targets and run_test_e2e in order to use project yarn. Introduce Makefile dependencies making sure yarn installation takes place before targets requiring yarn.

In Ubuntu and Debian, package cmdtest installs unrelated (python) /usr/bin/yarn and /bin/yarn which conflicts with globally installed (nodejs) yarn. Using a project-installed yarn resolves this issue. Incidentally, it reduces the contributor-facing dependencies to Node.js/npm.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,10 @@
 
 TEST_E2E_DIR = ocaml-lsp-server/test/e2e
 
-.PHONY: yarn-install
-yarn-install:
-	cd $(TEST_E2E_DIR) && yarn --frozen-lockfile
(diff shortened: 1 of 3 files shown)


rgrinberg:
Unfortunately this breaks my nix setup.

What about just making your yarn installed npm available in PATH?

rgrinberg:
It should be enough to do $ nix develop followed by running the tests

Rudi.

rgrinberg:
please re-open once the issues are fixed

PR #881, opened 2022-10-19
Remove omd < 2.0.0~alpha1

cuihtlauac:


ulugbekna:
Hi! This breaks the build because omd 2 has quite different API from omd 1. We can move to omd 2 when it becomes more stable.

PR #891, opened 2022-10-21
add info about deprecated alternatives

ShalokShalom:


README.md
```diff
@@ -174,8 +174,7 @@ the lsp protocol covers a wider scope than merlin.
 
 ## Comparison to other LSP Servers for OCaml
 
-Note that the comparisons below makes no claims of being objective and may be
-entirely out of date:
+Note that the comparisons below makes no claims of being objective. Both are deprecated.
```
ulugbekna, line 177: Thanks!

Subsumed by #895 ; I think this particular wording is ambiguous - "both" can be attributed to the comparisons rather than servers. The first server also is not clearly deprecated.

PR #899, opened 2022-10-26
fix(go-to-def): report error in response

ulugbekna:
fix(go-to-def): send a response with an error for failing go-to-def/go-to-decl/go-to-type-def
     rather than a notification afterwards with the error

also, should `definition_query`'s body be wrapped in `Fiber.of_thunk` ?

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -615,7 +615,7 @@ let definition_query kind (state : State.t) uri position =
       in
       Jsonrpc.Response.Error.raise
         (Jsonrpc.Response.Error.make
-           ~code:Jsonrpc.Response.Error.Code.InternalError
+           ~code:Jsonrpc.Response.Error.Code.RequestFailed
            ~message:(sprintf "Request \"Jump to %s\" failed." kind)
            ~data:
              (`String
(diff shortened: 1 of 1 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -601,11 +601,12 @@ let definition_query server (state : State.t) uri position merlin_request =
     match location_of_merlin_loc uri result with
     | Ok s -> Fiber.return s
     | Error message ->
-      let+ () =
-        let message = sprintf "Locate failed. %s" message in
-        State.log_msg server ~type_:Error ~message
-      in
-      None)
+      let message = sprintf "Merlin Locate failed: %s" message in
```
rgrinberg, line 604: can we move the message to the extra data field?

rgrinberg:
> also, should definition_query's body be wrapped in Fiber.of_thunk ?

Yes it does, good catch. However, I prefer `let* () = Fiber.return () in` since it doesn't push indentation to the right.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -601,11 +601,12 @@ let definition_query server (state : State.t) uri position merlin_request =
...
+      Jsonrpc.Response.Error.raise
+        (Jsonrpc.Response.Error.make
+           ~code:Jsonrpc.Response.Error.Code.InternalError
```
rgrinberg, line 607: I don't think this is the right error code. The error could easily occur due to user error - for example, the project isn't built yet. Does `RequestFailed` make more sense?

rgrinberg (approved):

PR #901, opened 2022-10-30
fix: jump to def/decl/type-def don't stop working after the first jump

### LINKED ISSUE #898: Locate failing with "not in environment"
## Problem

Go-to-def, go-to-declaration stop working after the first time going to the file of interest. 

## How to reproduce:

1. open `src/import.ml` in vscode_ocaml_platform project. 
2. trigger go-to-def on `Ocaml_version` module name line 210. This opens `.../ocaml_version.ml` 
3. switch back to file `src/import.ml` 
4. trigger again go-to-def on `Ocaml_version` → locate fails with message “not in environment: Ocaml_version” 

## Interesting observations:

- I thought whether opening `ocaml_version.ml` is causing this problem, so I put code in `ocaml_lsp_server.ml` (ocaml-lsp project) that would ignore the notification `textDocument/didOpen` if the file name is `ocaml_version.ml` . Indeed! After adding this code, I don’t observe this problem, so opening the file must be causing the problem.
    - either olsp opens a file and that triggers some problem in environments
    - or merlin is getting incorrect environment when a file is opened

- [https://github.com/ocaml/ocaml-lsp/blob/2cca06fa208318dadadcce05b60438f86f4cad63/ocaml-lsp-server/vendor/merlin/src/analysis/locate.ml#L765](https://github.com/ocaml/ocaml-lsp/blob/2cca06fa208318dadadcce05b60438f86f4cad63/ocaml-lsp-server/vendor/merlin/src/analysis/locate.ml#L765) is the first place that’s causing the error. `Env_lookup.in_namespaces nss ident env` returns `None`. The function that it invokes is `Env.find_module_by_name` that uses `Location.(!input_name)` that's equal to `import.ml` on the first run of go-to-def and `ocaml_version.ml` on the second run. Changing it on `Query_commands.dispatch` as such `Location.(input_name := (Mpipeline.final_config pipeline).Mconfig.query.filename);` before the main pattern match doesn't help.

- Invoking go-to-def on some other symbol in `import.ml` and coming back to `Ocaml_version` symbol and then doing go-to-def seems to fix it. 

----

Something similar was reported here before - https://github.com/ocamllabs/vscode-ocaml-platform/issues/970
The error message isn't exactly the same however

ulugbekna:
Fixes https://github.com/ocaml/ocaml-lsp/issues/898

It seems like we shouldn't reuse a pipeline if a pipeline has been constructed for another file.

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -824,7 +824,22 @@ let on_request :
   | TextDocumentDeclaration { textDocument = { uri }; position } ->
     later (fun state () -> definition_query `Declaration state uri position) ()
   | TextDocumentDefinition { textDocument = { uri }; position; _ } ->
-    later (fun state () -> definition_query `Definition state uri position) ()
+    let doc =
+      Document_store.change_document store uri ~f:(fun doc ->
+          (* we need [update_text] with no changes to get a new merlin pipeline;
+             otherwise the diagnostics don't get updated *)
+          Document.update_text doc [])
+    in
+    let+ (), state =
+      Fiber.fork_and_join
+        (fun (* TODO: should we really be updating the diagnostics here? *)
+               () -> set_diagnostics state.detached state.diagnostics doc)
+        (fun () ->
+          later
+            (fun state () -> definition_query `Definition state uri position)
+            ())
+    in
+    state
   | TextDocumentTypeDefinition { textDocument = { uri }; position; _ } ->
     later
       (fun state () -> definition_query `Type_definition state uri position)


rgrinberg:
> It seems like we shouldn't reuse a pipeline if a pipeline has been constructed for another file.

But how are we reusing a pipeline used for another file? We create a pipeline for every document.

rgrinberg:
In any case, it does seem sketchy what we're doing. We can't stop reusing pipelines altogether, but how about we only save the last used pipeline and reuse it for subsequent requests to the same document?

ulugbekna:
> But how are we reusing a pipeline used for another file? We create a pipeline for every document.

AFAIU, a pipeline creation invalidates all previously created pipelines, due the creation manipulating global variables such as `Location.input_name` in `setup_reader_config` fn in `mocaml.ml`. 

> how about we only save the last used pipeline and reuse it for subsequent requests to the same document?

Yes, that's what I intend to do.

rgrinberg:
Okay, carry on then. I misunderstood the PR's description then.

ulugbekna:
On second thought, we have `Mpipeline.with_pipeline` that implies that there should be more than one pipeline existing. That function invokes `Mocaml.with_state` that allows to set a set of global refs to a snapshot captured for this particular pipeline. `Env.Current_unit_name.t` and `Location.input_name` mutable vars do not seem to be snapshotted.

Also, there's `query` field in `Mconfig.t`, which allows specifying file and its directory, so could a pipeline be reused for different files? 

I'll try looking at how emacs/vim use merlin's API tomorrow.

ulugbekna:
I'll make another PR when I figure things out.

PR #904, opened 2022-11-04
fix: correctly use merlin's pipeline

rgrinberg:
do not reuse pipelines unless we are using the same pipeline multiple
times in a row

@ulugbekna let's get a quick fix in and work on optimizing in #901 parallel. This bug is likely very annoying to users.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Fixes
 
+- Fix random requests failing after switching documents (#904, fixes #898)
+
 - Do not offer related diagnostic information unless the user enables in client
   capabilities (#905)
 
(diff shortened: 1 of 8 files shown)


CHANGES.md
```diff
@@ -1,3 +1,9 @@
+# 1.14.2
+
+## Fixes
+
+- Fix random requests failing after switching documents (#901, fixes #898)
```
ulugbekna, line 5:
```suggestion
- Fix random requests failing after switching documents (#904, fixes #898)
```

ulugbekna (changes requested):
Looks correct. Indeed fixes the go-to-definition for me.

Noteworthy: there's `Mpipeline.make` used in `action_type_annotate.ml`. Could this trip up `Single_pipeline`? I couldn't reproduce the initial bug anyway.

rgrinberg:
>  there's Mpipeline.make used in action_type_annotate.ml. Could this trip up Single_pipeline? I couldn't reproduce the initial bug anyway.

It should be fine b/c it creates the pipeline inside the merlin thread.

ulugbekna:
Shall we merge this?

rgrinberg:
Let me try and fix the tests.

Rudi.

PR #920, opened 2022-11-16
re-enable ocamlformat-rpc

ulugbekna:
1) I think it could be enabled now (the disabling PR for ref: https://github.com/ocaml/ocaml-lsp/pull/555) 

2) I think that given ocamlformat-rpc is in ocamlformat now, **(option a)** we could completely remove the message when `ocamlformat-rpc` binary is not found. Otherwise, **(option b)** we can check ocamlformat version and, if it's <= 0.21.0, then we show the message. 

2.1) If we keep the message, should we make it a warning rather than an info message.


ulugbekna:
I'm not too happy with our tests depending on how ocamlformat formats things. Do you think I should disable ocamlformat-rpc for most tests and only keep a small number of tests that particularly test that ocamlformat-rpc works? If yes, should I go for an env variable or remove it from PATH for tests?

rgrinberg:
I don't particularly care since I pin the deps with nix.

rgrinberg (approved):
CHANGES should  be updated.

PR #921, opened 2022-11-16
feature(ofmt): much cleaner error reporting

ulugbekna:
Alternative to #720 

Turns out, ofmt simply prints a line of their own error message then compiler-reported error, so we can parse the error message using Ocamlc_loc and have clean error messages! 

image
image

image
image

- [x] handle several errors in one message (eg syntax error + warning 50) (will happen in another PR) 
- [x] write tests? 
- [x] add buttons that allow to jump to these locations? (via showMessageRequest?) (will happen in another PR)

### DIFF
--- a/ocaml-lsp-server/src/ocamlformat.ml
+++ b/ocaml-lsp-server/src/ocamlformat.ml
@@ -87,7 +87,62 @@ type error =
   | Unexpected_result of { message : string }
   | Unknown_extension of Uri.t
 
-let message = function
+let err_msg ~fallback_msg ~compiler_err =
+  let print_loc ppf ({ lines; chars; path = _ } : Ocamlc_loc.loc) =
+    let print_line ppf lines =
+      match (lines : Ocamlc_loc.lines) with
+      | Single line -> Format.fprintf ppf "line %d" line
+      | Range (l0, l1) -> Format.fprintf ppf "lines %d-%d" l0 l1
+    in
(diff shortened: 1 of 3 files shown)


ulugbekna:
This should be ready. 

I added tests to show how current error reporting looks like for ocamlformat. 

This PR improves on majority of cases when there's a single error being reported by ocamlformat. I will improve several-error case in a later PR. 

We should also note how we don't send any info when there's ocamlformat but we don't format the project because there's no .ocamlformat. In an ideal case, we should send a showMessageRequest with two options "never show this again for this project" and "create .ocamlformat". For the first option to work, we should maintain some state - either a global one or within the project itself. I prefer the former to avoid .gitignore changes, but I'm not sure where it is appropriate to keep this global state. Would love advice.

ulugbekna:
It's such such a shame that we show such an error message when we definitely could easily do better:

rgrinberg:
It's a no, sorry. Tools should produce error messages that can be consumed in different contexts, and it's not good for the project to accumulate a pile of hacks to make up for them not doing so. Even if this hack was small, it would still be of dubious value given that the context is this tiny issue.

Here's another approach that is at least reliable: given a formatting request, we can use merlin to check if the document has syntax errors. If it does, we can reject the formatting request or print the syntax errors again if the request fails. No parsing of anything required and we get essentially the same benefits.

PR #924, opened 2022-11-17
refactor: optimize content changes

rgrinberg:
only apply version at the end

ps-id: 440db09b-e2b5-44e9-9f71-2e1e8067282f

### DIFF
--- a/lsp/src/text_document.ml
+++ b/lsp/src/text_document.ml
@@ -131,8 +131,7 @@ let version (t : t) = t.document.version
 
 let languageId (t : t) = t.document.languageId
 
-let apply_content_change ?version (t : t)
-    (change : TextDocumentContentChangeEvent.t) =
+let apply_change (t : t) (change : TextDocumentContentChangeEvent.t) =
   let document =
     match change.range with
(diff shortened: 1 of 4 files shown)


lsp/src/text_document.ml
```diff
@@ -155,9 +154,10 @@ let apply_content_change ?version (t : t)
...
-    match version with
-    | None -> document
-    | Some version -> { document with version }
-  in
   { t with document }
+
+let apply_content_changes ?version t changes =
+  let t = List.fold_left apply_change t changes in
+  match version with
+  | None -> t
```
ulugbekna, line 162: I don't see where we're incrementing the version? 

```suggestion
  | None ->
    { t with document = { t.document with version = t.document.version + 1 } }
```

ulugbekna (changes requested):

lsp/src/text_document.ml
rgrinberg: I don't think it's needed. The spec says the versions are provided by the client to the server.

ulugbekna (approved):

PR #925, opened 2022-11-17
fix: respect completion item resolve capability

rgrinberg:

ps-id: 1286e94f-b84b-49a7-94f3-4a14af371361

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 - Fix document syncing for ranges that span an entire line (#927)
 
+- Respect the client's completion item resolve capability (#925)
+
 ## Features
 
 - Re-enable `ocamlformat-rpc` for code formatting (#920)
--- a/ocaml-lsp-server/src/compl.ml
+++ b/ocaml-lsp-server/src/compl.ml
@@ -148,7 +148,7 @@ module Complete_by_prefix = struct
         (* Without this field the client is not forced to respect the order
            provided by merlin. *)
       ~sortText:(sortText_of_index idx)
-      ~data:compl_params
+      ?data:compl_params
       ~textEdit
       ()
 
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+## Fixes
+
+- Respect the client's resolve completion item resolve capability (#925)
```
ulugbekna, line 5:
```suggestion
- Respect the client's completion item resolve capability (#925)
```

ocaml-lsp-server/src/compl.ml
```diff
@@ -259,13 +264,26 @@ let complete (state : State.t)
...
+            let open Option.O in
+            let* td = capabilities.textDocument in
+            let* compl = td.completion in
+            let* item = compl.completionItem in
+            item.resolveSupport
+          with
+          | None -> false
+          | Some { properties } ->
+            List.mem properties ~equal:String.equal "documentation"
+        in
```
ulugbekna, line 279: would it make sense to have a module-wide state - a lazy value - that would compute this value once instead of doing it on every completion

ulugbekna (commented):
1. Shouldn't we include the information that would be resolved by the resolve request right away if the client doesn't support resolve requests?

2. Do you also feel like `compl.ml` has become hard to read? (because of me)

ocaml-lsp-server/src/compl.ml
rgrinberg: Sure, it would make sense. We have a `Diganostics.t` that stores the results of computing capabilities. It will make sense here too.

rgrinberg:
> Shouldn't we include the information that would be resolved by the resolve request right away if the client doesn't support resolve requests?

Better not because resolving documentation for every completion candidate can be slow.

> Do you also feel like compl.ml has become hard to read? (because of me)

I suppose we can make some improvements. It wasn't hard for me to fix the capabilities issues though.

ulugbekna:
Hm, ok, let’s omit the documentation then. I imagine most editors should support resolve anyway

ulugbekna:
Hm, ok, let’s omit the documentation then. I imagine most editors should support resolve anyway

ulugbekna (approved):

Issue #929, opened 2022-11-18
Segfault on windows


mlasson:
# Description 

There's a "looping thread" that tries to repetitively read the file statistics (ie. Unix.stat) of the file "C:\Users\marcl\AppData\Local\dune/rpc" in order to communicate with running instances of dune watch (are there users of dune watch on windows yet ?).

When the file is absent, you'll get in the output:
```
failed to poll dune registry. Unix.Unix_error(Unix.ENOENT, "stat", "C:\\Users\\marcl\\AppData\\Local\\dune/rpc")
```
which is fine... until the OCaml heap gets corrupted which may change the error message into 
```
failed to poll dune registry. Unix.Unix_error(Unix.ENOENT, "stat", "RANDOM BINARY STRING")
```
or crash with a segfault or with "fatal: out of memory". Nasty stuff. 

In practice, ocaml-lsp is not usable on a large codebase (ocaml platform refuse to restart the server after consecutives crashes). 

I have not been able to reproduce this problem on a simple example (do not hesitate to add a smiley if you experienced the bug). 

But there is a fix: https://github.com/ocaml/ocaml/pull/11737.

The problem seems to come from a non "thread-safe" implementation of Unix.stat on windows.

Should we turn this loop off while the fix has not been integrated into ocaml ?

nojb:
> are there users of dune watch on windows yet ?

The feature is not yet supported (natively) on Windows, but there is an open PR for it: https://github.com/ocaml/dune/pull/6087. I hope to get it into a mergeable state soon-ish.

nojb:
> Should we turn this loop off while the fix has not been integrated into ocaml ?

Anoter possibility is to protect the call `Unix.stat` in a critical section.

rgrinberg:
> Should we turn this loop off while the fix has not been integrated into ocaml ?

If this is done only on windows then fine.

> Anoter possibility is to protect the call Unix.stat in a critical section.

Also sounds good.

mlasson:
> Anoter possibility is to protect the call `Unix.stat` in a critical section.

It is not completely trivial because the `Unix.stat` call in `ocaml-lsp-server/src/dune.ml` is only half of the problem. The other half in a `Unix.stat` in an unvendored dependency (merlin or dune).

nojb:
> The other half in a `Unix.stat` in an unvendored dependency (merlin or dune).

[...]

PR #935, opened 2022-11-21
Mitigate ocaml/ocaml#11737 (windows-only)

mlasson:
# Description 

 See #929. 

With this PR, we do not start the inifinite loop that is watching the "dune/rpc" file on windows. 

This loop exists to communicate with dune in watch mode... which is not yet supported on windows (although it should be supported "soonish" when https://github.com/ocaml/dune/pull/6087 will be merged). So, in practice, on windows, this loop is basically an infinite thread doing nothing but failing calls to `Unix.stat`. That is exactly what's needed to trigger ocaml/ocaml#11737.

Note: even with this PR, the ocaml bug could still lead to a segfault each time we do a Unix.stat on non-existing file. But 
with this PR, it should be harder to trigger. 

This PR should be reverted as soon as all supported compilers have ocaml/ocaml#11737 (> 4.14.0 ?) which, hopefully, should happen before #6087 is merged (or we will need another way to avoid this problem).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@
 
 - Respect the client's completion item resolve capability (#925)
 
+- Disable polling for dune's watch mode on Windows and OCaml 4.14.0 (#935)
+
 ## Features
 
 - Semantic highlighting support is enabled by default (#TODO)
--- a/ocaml-lsp-server/src/dune.ml
+++ b/ocaml-lsp-server/src/dune.ml
@@ -94,12 +94,19 @@ module Poll =
               | exception Sys_error _ -> Ok []
               | exception exn -> Error exn))
 
+      let is_414 = String.is_prefix Sys.ocaml_version ~prefix:"4.14.0"
+
       let stat s =
-        Fiber.of_thunk (fun () ->
-            Fiber.return
-              (match Unix.stat s with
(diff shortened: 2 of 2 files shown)


mlasson:
Alternatively, maintainers could chose to close this PR without merging. 
I have no idea if the "user base" of ocaml-lsp on windows is big enough to justify doing something special here.

rgrinberg:
Thanks for the fix. I do try to keep windows support working. I don't have time to test things thoroughly so your help is much appreciated.

mlasson:
commit [ca809dc](https://github.com/ocaml/ocaml-lsp/pull/935/commits/ca809dc8895a4d970087d6f4dda0650d0ed0b7a5) is an alternative way to avoid triggering the seg fault and will still allow to make dune watch work when it will be available !

rgrinberg:
> commit [ca809dc](https://github.com/ocaml/ocaml-lsp/pull/935/commits/ca809dc8895a4d970087d6f4dda0650d0ed0b7a5) is an alternative way to avoid triggering the seg fault and will still allow to make dune watch work when it will be available !

Thanks. I see you guys have already merged it to the 4.14 branch so I tightened the condition to disable it only for 4.14.0

PR #937, opened 2022-11-22
fix: respect deprecated completion capability

rgrinberg:
ps-id: 49a952ff-4c54-4210-8795-6a44d4c38311

### DIFF
--- a/ocaml-lsp-server/src/compl.ml
+++ b/ocaml-lsp-server/src/compl.ml
@@ -137,14 +137,14 @@ let sortText_of_index idx = Printf.sprintf "%04d" idx
 
 module Complete_by_prefix = struct
   let completionItem_of_completion_entry idx
-      (entry : Query_protocol.Compl.entry) ~compl_params ~range =
+      (entry : Query_protocol.Compl.entry) ~compl_params ~range ~deprecated =
     let kind = completion_kind entry.kind in
     let textEdit = `TextEdit { TextEdit.range; newText = entry.name } in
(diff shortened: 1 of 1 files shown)


ocaml-lsp-server/src/compl.ml
```diff
@@ -282,8 +282,15 @@ let complete (state : State.t)
...
+          let deprecated =
+            Option.value
+              ~default:false
+              (let open Option.O in
+              let* item = completion_item_capability in
+              item.deprecatedSupport)
+          in
           if not (Typed_hole.can_be_hole prefix) then
-            Complete_by_prefix.complete merlin prefix pos ~resolve
+            Complete_by_prefix.complete merlin prefix pos ~resolve ~deprecated
```
ulugbekna, line 293: I'm growing to believe that we should have a new type (record) that would contain all completion-related capabilities that is lazily evaluated and passed as a single argument to functions rather than several somewhat confusing named arguments

ulugbekna (approved):

ocaml-lsp-server/src/compl.ml
rgrinberg: Sure, I don't mind that. I'm not sure I'll get around to doing it soon though. Why does it matter if it's lazy or not?

ulugbekna: I'll work on that when I have time then. Lazy - just to not to recompute the bool value from the option on every completion if that makes sense.

PR #939, opened 2022-11-24
disable ofmt rpc for files

ulugbekna:
- fix(ofmt-rpc): disable ofmt-rpc for formatting files because it doesn't respect missing .ocamlformat and formats projects using default config
- refactor(formatter): cleaner error handling


rgrinberg (approved):
Needs to update the CHANGES entry

PR #942, opened 2022-12-01
chore: update merlin

rgrinberg:
ps-id: ea784131-dd41-4181-add7-fd3e5e304b00

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -1105,6 +1105,7 @@ let start () =
   Metrics.with_metrics metrics run
 
 let run ~read_dot_merlin () =
+  Merlin_utils.Lib_config.set_program_name "ocamllsp";
   Merlin_config.should_read_dot_merlin := read_dot_merlin;
   Unix.putenv "__MERLIN_MASTER_PID" (string_of_int (Unix.getpid ()));
   Lev_fiber.run ~sigpipe:`Ignore start |> Lev_fiber.Error.ok_exn
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,9 @@
 
 - Support utf-8 position encoding clients (#919)
 
+- Upgrade to merlin 4.7 and use merlin's `verbosity=smart` by default, which
+  allows unwrapping module alias types (#942)
+
 ## Fixes
 
 - Fix semantic highlighting of "long identifiers," e.g., `Foo.Bar.x` (#TODO)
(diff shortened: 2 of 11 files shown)


ulugbekna:
There should be breaking changes in the latest merlin, so we can either merge with errors and I fix them in a follow-up PR or I can try upgrading merlin and fixing errors in another PR

rgrinberg:
I made it build at least. A few test failures remaining

ulugbekna (approved):

rgrinberg:
Sure that would be appreciated

ulugbekna:
The test failures are caused by a regression in merlin - https://github.com/ocaml/merlin/issues/1540

Should we wait for fix (which is likely an easy one and we should be able to rebase and update merlin in olsp again) or I can update our tests to fetch documents for identifiers outside let-definition, which fixes our tests?

rgrinberg:
No opinion. Shall we ask the merlin team what's the best course of action?

Rudi.

ulugbekna:
I’ll just update the tests then

rgrinberg:
If we update the tests, how would we make sure there are no regressions with this issue?

ulugbekna:
I’ll add a test with a fixme that we’ll update as soon as we get merlin with this issue fixed. Hope this makes sense. This issue is a very minor one, given the documentation isnt shown only for values in let-definitions, meaning the documentation doesn’t need to appear on hover because it’s right there near cursor anyway

rgrinberg:
Sound good to me.

rgrinberg:
The windows tests are still failing. Any idea why?

rgrinberg:
We should be able to see better errors if we set `OCAMLLSP_TEST_DEBUG`

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -1105,6 +1105,7 @@ let start () =
   Metrics.with_metrics metrics run
 
 let run ~read_dot_merlin () =
+  Merlin_utils.Lib_config.set_program_name "OCaml-LSP";
```
rgrinberg, line 1108: How about just using the binary name? `ocamllsp`.

rgrinberg:
Shall we merge this and address the windows stuff separately?

ulugbekna:
> Shall we merge this and address the windows stuff separately?

Sure

ocaml-lsp-server/src/ocaml_lsp_server.ml
ulugbekna: fixed

PR #944, opened 2022-12-05
unite all olsp env var handling

ulugbekna:
The commit comment explains why. 

Ideally, we would write a config file for env vars, from which we could generate both `Env_vars` module and documentation for README, but that seems unnecessary right now. 

I can also shadow `Sys.getenv` and `Sys.getenv_opt` with empty functions, if we want to ensure people use `Env_vars` rather than those by `Sys` module.


rgrinberg (approved):
Seems harmless

PR #945, opened 2022-12-08
test: edit diffing

rgrinberg:

ps-id: 20262bc0-07a2-4234-ab4c-97cfc3f802d4

### DIFF
--- a/ocaml-lsp-server/test/ocaml_lsp_tests.ml
+++ b/ocaml-lsp-server/test/ocaml_lsp_tests.ml
@@ -17,3 +17,75 @@ let%expect_test "eat_message tests" =
   [%expect {| [PASS] |}];
   test "foo bar" "foo Bar" false;
   [%expect {| [PASS] |}]
+
+let%test_module "diff" =
+  (module struct
+    let test ~from ~to_ =
+      let edits = Ocaml_lsp_server.Diff.edit ~from ~to_ in
+      let json = `List (List.map Lsp.Types.TextEdit.yojson_of_t edits) in
+      Yojson.Safe.pretty_to_string ~std:false json |> print_endline
+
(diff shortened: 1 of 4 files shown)


ulugbekna (approved):
The result in `change a character` test seems less than ideal. 

Interestingly, looks like this diff fn could be generalized to be applicable to semantic tokens full/delta

ocaml-lsp-server/test/ocaml_lsp_tests.ml
```diff
@@ -17,3 +17,75 @@ let%expect_test "eat_message tests" =
...
+
+    let%expect_test "from empty" =
+      test ~from:"" ~to_:"foobar";
+      [%expect
+        {|
+        [
+          {
+            "newText": "foobar\n",
+            "range": {
+              "end": { "character": 0, "line": 0 },
```
rgrinberg, line 40: Should line be 1 here perhaps?

ulugbekna: Hm, your comment prompted me to think more about the edits. 

I think that this particular edit looks correct because it doesn't include a newline. According to the spec: 

> If you want to specify a range that contains a line including the line ending character(s) then use an end position denoting the start of the next line.

Also, editing from an empty file to `foobar\n` in vscode, sends an edit that looks like this

```
 "contentChanges": [
        {
            "range": {
                "start": {
                    "line": 0,
                    "character": 0
                },
                "end": {
                    "line": 0,
                    "character": 0
                }
            },
            "rangeLength": 0,
            "text": "foobar\n"
        }
    ]
```

However, with this logic, test `to empty` should replace range `([0, 0], [0, 7])`, where braces denote `[line, char]`. Similar logic in `multiline`

PR #946, opened 2022-12-11
feature: channel argument support

rgrinberg:
I haven't added real support for this, but at least we recognize the common flags now so that users will get decent error messages.

### DIFF
--- a/ocaml-lsp-server/bin/main.ml
+++ b/ocaml-lsp-server/bin/main.ml
@@ -1,16 +1,35 @@
+open Stdune
+module Cli = Lsp.Cli
+
 let () =
   Printexc.record_backtrace true;
   let version = ref false in
   let read_dot_merlin = ref false in
-  Arg.parse
+  let arg = Lsp.Cli.Arg.create () in
+  let spec =
     [ ("--version", Arg.Set version, "print version")
     ; ( "--fallback-read-dot-merlin"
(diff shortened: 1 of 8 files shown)


ocaml-lsp-server/bin/main.ml
```diff
@@ -1,16 +1,61 @@
...
-    (fun _ -> raise (Arg.Bad "anonymous arguments are not accepted"))
-    "ocamllsp";
+  in
+  let usage = "ocamllsp [ stdio ]" in
+  Arg.parse spec (fun arg -> anon_args := arg :: !anon_args) usage;
+  let anon_args = List.rev !anon_args in
+  let channel =
+    match
+      match anon_args with
+      | [] -> Ocaml_lsp_server.Stdio
```
ulugbekna, line 26: we now only support stdio, but I think we should be checking here whether `!pipe` or `!port` is Some, right? was this omitted intentionally?

rgrinberg: Do you mean is None? Those arguments are useless with Stdio.

rgrinberg: I fixed the argument parsing a little.

rgrinberg:
I misinterpreted how the flags were passed. This is now fixed and I've added support for all the other modes as well.

ulugbekna (approved):

PR #950, opened 2022-12-16
revert releases

ulugbekna:
- Revert "chore: prepare 1.15.0"
- Revert "upgrade to merlin with support for ocaml 5, upgrade olsp opam file to ocaml 5"
- Revert "Add missing declaration for functions"

### DIFF
--- a/submodules/lev/.github/workflows/workflow.yml
+++ b/submodules/lev/.github/workflows/workflow.yml
@@ -15,7 +15,7 @@ jobs:
           - ubuntu-latest
           - windows-latest
         ocaml-compiler:
-          - 5.0.x
+          - 4.14.x
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -19,7 +19,7 @@ jobs:
           - ubuntu-latest
           - windows-latest
         ocaml-compiler:
-          - 5.0.x
+          - 4.14.x
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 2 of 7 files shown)


submodules/lev/lev/src/lev_stubs.c
```diff
@@ -57,9 +57,6 @@
     CAMLreturn(Val_long(__value));                                             \
   }
 
-CAMLextern int caml_convert_signal_number(int);
-CAMLextern int caml_rev_convert_signal_number(int);
```
rgrinberg, line 61: This commit can stay. We just need it in lev as well.

ulugbekna: remove the revert of that particular commit; here's the PR to lev: https://github.com/rgrinberg/lev/pull/86

submodules/lev/.github/workflows/workflow.yml
```diff
@@ -15,7 +15,7 @@ jobs:
           - ubuntu-latest
           - windows-latest
         ocaml-compiler:
-          - 5.0.x
+          - 4.14.x
```
rgrinberg, line 18: I'm surprised about this change. We should be testing on 4.14 and 5.0. Could you add an appropriate change in lev for this as well?

ulugbekna: https://github.com/rgrinberg/lev/pull/87

I expect something like this?

rgrinberg (approved):

ulugbekna:
should I squash merge?

rgrinberg:
Individual commits are fine here

PR #971, opened 2022-12-26
fix(lsp): text edit application

rgrinberg:
Do not eat newlines when applying edits past the newline character.

Any edit where the character position goes beyond the newline char
should be truncated to right before the newline.

ps-id: ae5e0adc-18a7-450d-be78-589c4995e3d6

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,6 +30,9 @@
 
 - Fix semantic highlighting of "long identifiers," e.g., `Foo.Bar.x` (#932)
 
+- Fix syncing of document contents. Previously, whole line edits would
+  incorrectly eat the newline characters. (#971)
+
 # 1.14.2
 
 ## Fixes
--- a/lsp/test/text_document_tests.ml
+++ b/lsp/test/text_document_tests.ml
@@ -107,7 +107,7 @@ let%expect_test "replace first line" =
 
 let%expect_test "beyond max char" =
   let range = tuple_range (0, 0) (0, 100) in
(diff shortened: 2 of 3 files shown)


lsp/test/text_document_tests.ml
```diff
@@ -103,12 +103,12 @@ let%expect_test "replace first line" =
 
 let%expect_test "beyond max char" =
   let range = tuple_range (0, 0) (0, 100) in
-  test "foo\nbar\n" range ~change:"baz\n";
+  test "foo\nbar\n" range ~change:"baz";
   [%expect {|
     UTF16:
-    baz\nbar\n
+    bazbar\n
```
ulugbekna, line 109: According to my understanding of the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range), the edited text should be `baz\nbar\n`, meaning that a range spanning beyond max char should not include the line ending `\n` (?)

rgrinberg: ah yes, my bad. pushing a fix.

CHANGES.md
```diff
@@ -30,6 +30,9 @@
 
 - Fix semantic highlighting of "long identifiers," e.g., `Foo.Bar.x` (#932)
 
+- Fix syncing of document contents. Previously, whoel line edits would
```
ulugbekna, line 33:
```suggestion
- Fix syncing of document contents. Previously, whole line edits would
```

ulugbekna (commented):
LGTM

PR #972, opened 2022-12-26
refactor: semantic highlighting

rgrinberg:
shrink the api

ps-id: 005c5dda-4f84-4e81-aba1-43aac46dee11

### DIFF
--- a/ocaml-lsp-server/src/semantic_highlighting.ml
+++ b/ocaml-lsp-server/src/semantic_highlighting.ml
@@ -10,142 +10,80 @@ open Fiber.O
    - [ ] support textDocument/semanticTokens/range *)
 
 module Token_type : sig
-  type t = private int
+  type t
+
+  val of_builtin : SemanticTokenTypes.t -> t
 
   val module_ : t
 
-  val type_ : t
-
-  val class_ : t
-
-  val enum : t
-
   val module_type : t
 
-  val record : t
-
-  val type_parameter : t
-
-  val parameter : t
-
-  val variable : t
-
-  val property : t
-
-  val enum_member : t
-
-  val event : t
-
-  val function_ : t
-
-  val method_ : t
-
-  val macro : t
-
-  val keyword : t
-
-  val modifier : t
-
-  val comment : t
-
-  val string : t
-
-  val number : t
-
-  val regexp : t
-
-  val operator : t
-
-  val decorator : t
-
-  val list : string list
+  val to_int : t -> int
 
   val to_legend : t -> string
-end [@warning "-32"] = struct
-  type t = int
 
-  let module_ = 0
-  (* the number is determined by the index in [list] defined below; e.g., we use
-     [0] here because we'd like to use "namespace" token type for module
-     symbols *)
+  val tokenTypes : string list
+end = struct
+  type t = SemanticTokenTypes.t
 
-  let type_ = 1
+  let of_builtin t = t
 
-  let class_ = 2
+  let module_ = SemanticTokenTypes.Namespace
 
-  let enum = 3
+  let module_type = SemanticTokenTypes.Interface
 
-  let module_type = 4
-
-  let record = 5
-
-  let type_parameter = 6
-
-  let parameter = 7
-
-  let variable = 8
-
-  let property = 9
-
-  let enum_member = 10
-
-  let event = 11
-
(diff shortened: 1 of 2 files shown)


ocaml-lsp-server/src/semantic_highlighting.ml
```diff
@@ -10,142 +10,94 @@ open Fiber.O
...
+         List.iteri legend ~f:(fun data key -> Table.add t ~key ~data);
+         t)
+    in
+    fun t ->
+      Table.find
+        (Lazy.force table)
+        (match t with
+        | Module -> Namespace
+        | Module_type -> Interface
+        | Builtin s -> s)
```
ulugbekna, line 89: This could be `(simplify t)`

ulugbekna (approved):
I went with an int-based token types initially because I assumed it would be faster. I guess this implementation is fine as well. 

Since we keep `Token_type.module_` and `Token_type.module_type`, we could use those variables as aliases to `Namespace` and `Interface` rather than have a separate variant for them. Not a strong opinion

ocaml-lsp-server/src/semantic_highlighting.ml
rgrinberg: Done. I simplified the representation a bit more.

rgrinberg:
> Since we keep Token_type.module_ and Token_type.module_type, we could use those variables as aliases to Namespace and Interface rather than have a separate variant for them. Not a strong opinion

I did it. It shortens the code and makes things a little more efficient.

ulugbekna (approved):

ocaml-lsp-server/src/semantic_highlighting.ml
```diff
@@ -10,142 +10,80 @@ open Fiber.O
...
-  val decorator : t
-
-  val list : string list
+  val to_int : t -> int
 
   val to_legend : t -> string
-end [@warning "-32"] = struct
-  type t = int
-
-  let module_ = 0
```
ulugbekna, line 67: I think we could also just make this code more self-evident by replacing a magic int with something like `List.findi legend "namespace"` and not bother with having `of_builtin` everywhere

You may also be bothered by large modules - I guess this could be better just moved into a separate file

ocaml-lsp-server/src/semantic_highlighting.ml
```diff
@@ -10,142 +10,80 @@ open Fiber.O
...
+
+  let to_int =
+    let module Table = MoreLabels.Hashtbl in
+    let table =
+      lazy
+        (let t = Table.create (List.length legend) in
+         List.iteri legend ~f:(fun data key -> Table.add t ~key ~data);
+         t)
+    in
+    fun t -> Table.find (Lazy.force table) t
```
ulugbekna, line 75: we could replace this with a pattern match - should be much more efficient without hashing? (although I don't know how ocaml hashes constructors - just integers essentially - may be a cheap thing) (again would introduce magic ints, however)

rgrinberg: The ints aren't the problem, it's just that this large pattern match is going to add a lot of code for performance gains that are very likely to be superficial.

ulugbekna: Sounds fair

ocaml-lsp-server/src/semantic_highlighting.ml
rgrinberg: > You may also be bothered by large modules - I guess this could be better just moved into a separate file

Yes, I'm more bothered by the amount of code this adds. I'd rather us focus towards adding a proper semantic tokens api to the `lsp` module and just using whatever it provides. Once we do that, we can focus on optimizing and improving the API.

ocaml-lsp-server/src/semantic_highlighting.ml
rgrinberg: I'd suggest we focus on optimizing once we have a nice and general purpose API for semantic tokens in the lsp library.

ocaml-lsp-server/src/semantic_highlighting.ml
ulugbekna: `lsp` could use extensible variants nicely for this, I think

PR #985, opened 2022-12-30
refactor(diff): calculate string map once

rgrinberg:


rgrinberg:
Merging as i'd like to start testing this on real projects and get some decent performance metrics. The optimization is trivial and well covered by the tests.

PR #990, opened 2022-12-31
refactor(lsp): remove allocation in diffs

rgrinberg:
This PR is just an optimization that cuts a few allocations. The fix will come in another PR.


rgrinberg:
More trivial optimizations to give the GC some relief.

PR #995, opened 2023-01-02
feature(lsp): TextDocumentEdit support

rgrinberg:
Add a function to apply text document edits and use it to test diff
generation

### DIFF
--- a/lsp/src/text_document.ml
+++ b/lsp/src/text_document.ml
@@ -1,5 +1,7 @@
 open Types
 module String = StringLabels
+module List = ListLabels
+module Map = MoreLabels.Map
 
 exception Invalid_utf8
 
@@ -41,7 +43,7 @@ let find_utf8_pos =
     assert (Uutf.decoder_line dec = 1);
     find_pos newline character dec + start
 
-let find_offset_8 ~utf8 ~utf8_range:range =
+let find_offset_8 ~utf8 range =
   let { Range.start; end_ } = range in
   let start_line_offset = find_nth_nl utf8 ~nth:start.line ~start:0 in
   let end_line_offset =
@@ -62,7 +64,7 @@ let find_offset_8 ~utf8 ~utf8_range:range =
   in
   (start_offset, end_offset)
 
-let find_offset_16 ~utf8 ~utf16_range:range =
+let find_offset_16 ~utf8 range =
   let dec =
     Uutf.decoder ~nln:(`ASCII newline) ~encoding:`UTF_8 (`String utf8)
   in
@@ -138,8 +140,8 @@ let apply_change encoding text (change : TextDocumentContentChangeEvent.t) =
     let start_offset, end_offset =
       let utf8 = text in
       match encoding with
-      | `UTF16 -> find_offset_16 ~utf8 ~utf16_range:range
-      | `UTF8 -> find_offset_8 ~utf8 ~utf8_range:range
+      | `UTF16 -> find_offset_16 ~utf8 range
+      | `UTF8 -> find_offset_8 ~utf8 range
     in
     String.concat
       ~sep:""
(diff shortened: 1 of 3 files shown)


rgrinberg:
Note that this function is not used in ocamllsp. It's only used in the tests but is likely useful to anyone developing an lsp client. `TextEdit.t` lists are interpreted differently than change notifications, so it's useful to have this.

lsp/src/text_document.ml
```diff
@@ -159,3 +164,59 @@ let apply_content_changes ?version t changes =
     | Some version -> { document with version }
   in
   { t with document }
+
+type change =
+  { start : int
+  ; stop : int
```
ulugbekna, line 170: mention `start` and `stop` are offsets would be useful

lsp/src/text_document.ml
```diff
@@ -159,3 +164,59 @@ let apply_content_changes ?version t changes =
...
+
+module Edit_map = Map.Make (struct
+  type t = int
+
+  let compare = Int.compare
+end)
+
+let add_edit map change =
+  (* TODO check non overlapping property *)
+  Edit_map.update map ~key:change.start ~f:(function
```
ulugbekna, line 182: how is it that we expect non-overlapping ranges but keep a _multi_-map where the key is the same start offset?

lsp/src/text_document.ml
```diff
@@ -159,3 +164,59 @@ let apply_content_changes ?version t changes =
...
+    simplify_changes acc text find_position changes
+
+let apply_changes text encoding changes =
+  let find_position =
+    match encoding with
+    | `UTF8 -> find_offset_8
+    | `UTF16 -> find_offset_16
+  in
+  let simplified = simplify_changes Edit_map.empty text find_position changes in
+  let pos = ref 0 in
```
ulugbekna, line 202: I think this keeps an index into `text`? (a comment would be helpful) 

I just don't understand the comment "if this is an insert, it's allowed to increase the position" - if `pos` is an index in `text`, shouldn't it remain unchanged on an insertion?

Also, in `pos := max !pos stop` - can `stop` be ever less than `pos` ?

ulugbekna (commented):
I'm a bit lost at the goal of the PR:

Do I understand correctly that the main goal of this PR is to support edits without constructing `TextDocumentContentChangeEvent`s? (modulo it has its own edit application)

Do you think it's a good idea to have two edit application functions - `appy_content_changes` vs `apply_text_document_edits` - and test the latter, while the former is used by ocamllsp? Do we plan to transition to the latter in ocamllsp?

lsp/src/text_document.ml
rgrinberg: We can change the map to use `(start * stop)` as the key.

rgrinberg:
> Do I understand correctly that the main goal of this PR is to support edits without constructing TextDocumentContentChangeEvents? (modulo it has its own edit application)

The relevant goal for us is to test the `TextEdit.t list` we produce for diffs. But as I said, it's a generally useful function for people writing LSP clients (not us) and our lsp library is supposed to be general purpose.

> Do you think it's a good idea to have two edit application functions - appy_content_changes vs apply_text_document_edits - and test the latter, while the former is used by ocamllsp?

Isn't that what we have? 

> Do we plan to transition to the latter in ocamllsp?

Nope, but we will hopefully find these functions more and more useful as we test workspace edits the server generates more thoroughly.

lsp/src/text_document.ml
rgrinberg: I'll send another PR to refactor this stuff.

lsp/src/text_document.ml
rgrinberg: Even though I wrote this code only 2 days ago, I'm already not particularly confident in the implementation.  Feel free to improve it and add tests.

>  if pos is an index in text, shouldn't it remain unchanged on an insertion?

If it's an insert, then we have `start = stop`, which means that `pos := max !pos stop` will not advance past the start position of the edit. This should answer your question.

> Also, in pos := max !pos stop - can stop be ever less than pos ?

Only if we provide an invalid range where `start > end_`

PR #996, opened 2023-01-03
refactor(lsp): less string allocation on splits

rgrinberg:


rgrinberg:
More simple deforestation that replaces copies with views

PR #997, opened 2023-01-04
refactor(lsp): do not copy substrings to concatenate them

ulugbekna:


rgrinberg (approved):

ulugbekna:
I'm not sure we have this covered by tests, so this change may not be tested. Do you think this should go in this or next release?

rgrinberg:
Pretty sure it's well tested but you can some more unit tests to substring if you want.

PR #999, opened 2023-01-05
prepare 1.15.0

ulugbekna:
Cleans up CHANGES and updates version number

@rgrinberg we have dune polling disabled for ocaml 4.14. Do I understand correctly that we don't need it disabled for ocaml 5 since the upstream issue is fixed?

also, perhaps, it makes sense to have release version in CHANGES set to 1.15.0-4.14 on master (vs. just 1.15.0) and 1.15.0-5.0 on 500?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,26 +2,36 @@
 
 ## Features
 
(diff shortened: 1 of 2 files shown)


rgrinberg:
Yes, dune polling is only disabled on 4.14 + windows.

PR #1004, opened 2023-01-10
refactor(lsp): batch updates with a zipper

### LINKED ISSUE #1003: ocaml-lsp reports bogus erros on large files when editor autosaving is enabled
I can reproduce the following issue on OCaml 4.14 with ocaml-lsp-server 1.13.1 and 1.14.2, and OCaml 5.0 with ocaml-lsp-server pinned to current `#500` (176d387926dfb0a0b2b0196395ad38000395c049).

It's a bit tricky to reproduce, but the following seems to work relatively reliably.

```
% opam source core
Successfully extracted to /tmp/core.v0.15.1
% dune build -w core/src/core.cmxs
```

Make some quick changes in large file with auto-saving enabled. I use the following vimrc:

```
command! AutoWrite autocmd TextChanged,InsertLeave <buffer> try | undojoin | silent! update | catch /^Vim\%((\a\+)\)\=:E790/ | finally | silent! update | endtry

augroup autowrite
    autocmd!
    autocmd Filetype ocaml AutoWrite
augroup END
```

Edit `core/src/command.ml` and repeatedly paste `open Sys` at the top. After a few pastes, lsp reports a bogus syntax error, while the dune watch command doesn't print anything of the sort. 

Here is a recording: https://asciinema.org/a/gQqsK784tKJYwPxtB7wBn6I61
And the lsp logfile: [lsp.log](https://github.com/ocaml/ocaml-lsp/files/10376336/lsp.log)

rgrinberg:
Neovim is particularly guilty of sending a ton of change notifications - essentially one for every character change. The end result is that editing large documents slows the lsp server to a crawl. The issue is somewhat mitigated by debouncing, but the delays that introduces are rather annoying.

We fix the issue by batching the updates in a single notification. We move along the WIP document using a zipper to insert the edits. In the end, we flatten everything into a string without allocating large intermediate strings. Insertions near the zipper location are fast.

@copy no more debouncing should be necessary in neovim.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Fixes
+
+- Fix a document syncing issue when utf-16 is the position encoding (#1004)
+
 # 1.15.1
 
 ## Fixes
@@ -6,8 +12,6 @@
   running ([#1005](https://github.com/ocaml/ocaml-lsp/pull/1005), fixes
   [#941](https://github.com/ocaml/ocaml-lsp/issues/941),
   [#1003](https://github.com/ocaml/ocaml-lsp/issues/1003))
-
-# 1.15.0
 
 ## Features
 
--- a/lsp/src/import.ml
+++ b/lsp/src/import.ml
@@ -1,6 +1,8 @@
 module List = Stdlib.ListLabels
 module Option = Stdlib.Option
 module Array = Stdlib.ArrayLabels
+module Bytes = Stdlib.BytesLabels
+module Map = Stdlib.MoreLabels.Map
 
 module Result = struct
   include Stdlib.Result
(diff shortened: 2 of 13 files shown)


rgrinberg:
@ulugbekna in addition to code review, it would be nice if you could test this in day to day usage. I'm doing so already, but I'm on neovim.

ulugbekna:
I will finish the review tomorrow. Sorry for the wait

rgrinberg:
Thanks. Take your time. This is a sensitive PR so let's make sure to get it right.

ulugbekna (approved):
I rebased and made a couple of small changes. Hope that's okay. Let me know if not. 

Looks good, great work

PR #1005, opened 2023-01-10
refactor: simplify merlin diagnostics

rgrinberg:
Updating the document with no changes is no longer needed because we
don't reuse the pipeline.

### DIFF
--- a/ocaml-lsp-server/src/document.ml
+++ b/ocaml-lsp-server/src/document.ml
@@ -128,22 +128,15 @@ module Single_pipeline : sig
     -> f:(Mpipeline.t -> 'a)
     -> ('a, Exn_with_backtrace.t) result Fiber.t
 end = struct
-  type t =
-    { thread : Lev_fiber.Thread.t
-    ; mutable last : (Text_document.t * Mconfig.t * Mpipeline.t) option
-    }
+  type t = { thread : Lev_fiber.Thread.t } [@@unboxed]
 
-  let create thread = { thread; last = None }
+  let create thread = { thread }
 
   let use t ~doc ~config ~f =
     let* config = Merlin_config.config config in
     let make_pipeline =
-      match t.last with
-      | Some (doc', config', pipeline) when doc' == doc && config == config' ->
-        fun () -> pipeline
-      | _ ->
-        let source = Msource.make (Text_document.text doc) in
-        fun () -> Mpipeline.make config source
+      let source = Msource.make (Text_document.text doc) in
+      fun () -> Mpipeline.make config source
     in
     let task =
       match
@@ -152,7 +145,7 @@ end = struct
             let pipeline = make_pipeline () in
             let res = Mpipeline.with_pipeline pipeline (fun () -> f pipeline) in
             let stop = Unix.time () in
-            (res, pipeline, start, stop))
+            (res, start, stop))
       with
       | Error `Stopped -> assert false
       | Ok task -> task
@@ -160,7 +153,7 @@ end = struct
     let* res = await task in
     match res with
     | Error exn -> Fiber.return (Error exn)
-    | Ok (res, pipeline, start, stop) ->
+    | Ok (res, start, stop) ->
       let event =
         let module Event = Chrome_trace.Event in
         let dur = Event.Timestamp.of_float_seconds (stop -. start) in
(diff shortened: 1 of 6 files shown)


ocaml-lsp-server/src/document_store.ml
```diff
@@ -190,19 +190,12 @@ let get_semantic_tokens_cache : t -> Uri.t -> semantic_tokens_cache option =
...
-  in
-  Fiber.parallel_iter all ~f:(fun (uri, doc) ->
-      let+ doc =
-        match doc.document with
-        | None -> Fiber.return doc
-        | Some document ->
-          let+ document = f document in
-          { doc with document = Some document }
-      in
-      Table.set t.db uri doc)
```
rgrinberg, line 205: @ulugbekna this code was evil. We're updating a value in the hash table after waiting for the termination of `f`, but the value might be out of date by then due to a concurrent document change notification.

ocaml-lsp-server/src/document.ml
```diff
@@ -128,22 +128,15 @@ module Single_pipeline : sig
     -> f:(Mpipeline.t -> 'a)
     -> ('a, Exn_with_backtrace.t) result Fiber.t
 end = struct
-  type t =
-    { thread : Lev_fiber.Thread.t
-    ; mutable last : (Text_document.t * Mconfig.t * Mpipeline.t) option
-    }
+  type t = { thread : Lev_fiber.Thread.t }
```
ulugbekna, line 131: could use `[@@unboxed]` perhaps

ocaml-lsp-server/src/document.ml
```diff
@@ -128,22 +128,15 @@ module Single_pipeline : sig
     -> f:(Mpipeline.t -> 'a)
     -> ('a, Exn_with_backtrace.t) result Fiber.t
 end = struct
-  type t =
-    { thread : Lev_fiber.Thread.t
-    ; mutable last : (Text_document.t * Mconfig.t * Mpipeline.t) option
```
ulugbekna, line 133: does this mean that we don't try reusing the last pipeline if it's the same document?

rgrinberg: Yes, the optimization is gone. I don't think it ever worked anyway given that it relied on physical equality of the `Mconfig.t` being return.

ocaml-lsp-server/src/dune.ml
```diff
@@ -263,15 +263,11 @@ end = struct
...
+                          | `Other -> Fiber.return ()
                           | `Merlin merlin ->
-                            let doc = Document.update_text doc [] in
-                            let+ () =
-                              Diagnostics.merlin_diagnostics diagnostics merlin
-                            in
-                            doc)
+                            Diagnostics.merlin_diagnostics diagnostics merlin)
                     in
                     Diagnostics.send diagnostics `All
```
ulugbekna, line 272: Does this re-make pipelines and send diagnostics to all files on a dune build completion (when dune is in watch mode)? Sounds super expensive

ulugbekna (approved):

ocaml-lsp-server/src/dune.ml
rgrinberg: Wait until you see how expensive is the build itself :)

But more seriously, of course it would be better to recompute only the documents which were invalidated by the build. Unfortunately that would be far more difficult.

ulugbekna: > the documents which were invalidated by the build. Unfortunately that would be far more difficult.

But dune knows what those documents are, right? (because of the incremental re-building it has) 

If yes, that becomes a matter of extending the dune rpc?

ulugbekna: > the documents which were invalidated by the build. Unfortunately that would be far more difficult.

actually, yes, I started noticing how my laptop fans go off often if I have dune running in watch mode

rgrinberg: Yes, that would be enough. In particular, the rule would be that a merlin document should be invalidated whenever:

* Any .cmi file in its `-I` directories has been invalidated
* Any of the preprocessing binaries has been invalidated
* Any of the runtime dependencies of the preprocessing binaries has been invalidated

rgrinberg: > actually, yes, I started noticing how my laptop fans go off often if I have dune running in watch mode

Thankfully you can limit the concurrency with `-j`. That calms down the fans considerably.

PR #1006, opened 2023-01-10
feature: add command to dump a document

rgrinberg:
@copy would you mind reproducing the bug with this PR and running:

```
:lua = vim.lsp.buf.execute_command({
    command = "ocamllsp/show-document-text",
    arguments = {
      vim.uri_from_bufnr(vim.api.nvim_get_current_buf())
    },
})
```

This command should open a file with the contents of the currently opened document as the lsp server sees it. If the result is different, it's a syncing issue. Otherwise, it must be something else.

### DIFF
--- a/ocaml-lsp-server/src/document_text_command.ml
+++ b/ocaml-lsp-server/src/document_text_command.ml
@@ -1,0 +1,28 @@
+open Import
+open Fiber.O
+
+let command_name = "ocamllsp/show-document-text"
+
+let command_run server store args =
+  let uri =
+    match args with
+    | Some [ `String arg ] -> arg
+    | _ -> assert false
+  in
+  let doc =
+    Document_store.get store (Uri.t_of_yojson (`String uri)) |> Document.text
+  in
+  let uri, chan =
+    Filename.open_temp_file
+      (sprintf "ocamllsp-document.%d" (Unix.getpid ()))
+      (Filename.extension uri)
+  in
+  output_string chan doc;
+  close_out_noerr chan;
+  let req =
+    let uri = Uri.of_path uri in
+    Server_request.ShowDocumentRequest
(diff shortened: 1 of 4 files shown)


copy:
Yes, there is a difference between the the buffer and the tempfile returned by ocamllsp/show-document-text (and the syntax error fits the latter):

```
% diff /tmp/core.v0.15.1/core/src/command.ml /tmp/ocamllsp-document*
5,8d4
< open Sys
< open Sys
< open Sys
< open Sys
10c6,7
< module Stable = struct
---
> module S
> open Sys
```

Buffer:

```
open Sys
open Sys
open Sys
open Sys
open Sys
open Sys
open Sys
open Sys

module Stable = struct
  module Shape = Command_shape.Stable
end
```

ocamllsp:

```
open Sys
open Sys
open Sys
open Sys

module S
open Sys
  module Shape = Command_shape.Stable
end
```

(minor note: the lua you posted prints: `ocamllsp: -32601: MethodNotFound`, but I did see the request in the lsp log and found the lsp buffer in /tmp/ocamllsp-document...ml)

rgrinberg:
@copy Awesome. This is indeed a syncing issue. I will prepare a special build to diagnose the concrete change notifications causing the issue.

> (minor note: the lua you posted prints: ocamllsp: -32601: MethodNotFound, but I did see the request in the lsp log and found the lsp buffer in /tmp/ocamllsp-document...ml)

Might be because your neovim is too old. Opening documents in the editor from the server needs the master branch IIRC.

@ulugbekna yes, I will add something to help diagnose this issue.

rgrinberg:
This command is still useful and it adds no overhead or requires additional book keeping. Will keep it around for the next time the syncing bug strikes.

PR #1007, opened 2023-01-11
debug changes

rgrinberg:
@copy could you try reproducing the issue with this PR?

You need to set the log file with an environment variable: `OCAMLLSP_DEBUG_CHANGES=ocamllsp.log`. Then, just reproduce the bug and paste the first change set that introduces the incorrect document.

### DIFF
--- a/lsp/src/text_document.ml
+++ b/lsp/src/text_document.ml
@@ -133,10 +133,24 @@ let version (t : t) = t.document.version
 
 let languageId (t : t) = t.document.languageId
 
+let debug =
+  match Sys.getenv_opt "OCAMLLSP_DEBUG_CHANGES" with
+  | None -> None
+  | Some f -> Some (open_out f)
+
+let () =
+  at_exit (fun () ->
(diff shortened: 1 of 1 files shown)


ulugbekna:
I can reproduce but seemingly not if I rebase onto #1005 

I have a suspicion that we may be changing the document in the store with an older version of it.

ulugbekna:
I also think that when auto-save is on, we try sending in a `task_if_running` diagnostics a lot more often, it could be possible that we compute diagnostics for a version of document that's older than the one in doc store?

rgrinberg:
> I also think that when auto-save is on, we try sending in a task_if_running diagnostics a lot more often, it could be possible that we compute diagnostics for a version of document that's older than the one in doc store?

I don't think it would matter. An additional save in the end would fix the situation with the diagnostics. While @copy was telling us that doing additional saves at the end does not change anything.

rgrinberg:
the bug is fixed

PR #1011, opened 2023-01-12
refactor(ocamllsp): move definition to own module

rgrinberg:
Move goto-definition to own module

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -361,33 +361,6 @@ module Formatter = struct
         Some (Diff.edit ~from:(Document.text doc) ~to_))
 end
 
-let location_of_merlin_loc uri : _ -> (_, string) result = function
-  | `At_origin -> Ok None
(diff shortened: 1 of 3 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -663,12 +604,16 @@ let on_request :
   | TextDocumentHighlight req -> later highlight req
   | DocumentSymbol { textDocument = { uri }; _ } -> later document_symbol uri
   | TextDocumentDeclaration { textDocument = { uri }; position } ->
-    later (fun state () -> definition_query `Declaration state uri position) ()
+    later
+      (fun state () -> Definition_query.run `Declaration state uri position)
```
ulugbekna, line 608: `Go_to_def.query` seems a bit more self-explanatory than `Definition_query.run` but not a strong opinion

rgrinberg: I think I prefer to just follow the names of lsp requests themselves.

PR #1019, opened 2023-01-19
refactor(lsp): share invalid utf8 exceptions

rgrinberg:

### DIFF
--- a/lsp/src/text_document.ml
+++ b/lsp/src/text_document.ml
@@ -9,7 +9,11 @@ include struct
   module TextEdit = TextEdit
 end
 
-exception Invalid_utf8
+type invalid_utf = String_zipper.invalid_utf =
+  | Malformed of string
+  | Insufficient_input
+
+exception Invalid_utf = String_zipper.Invalid_utf
 
 let newline = Uchar.of_char '\n'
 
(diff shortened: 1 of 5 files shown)


ocaml-lsp-server/src/document.ml
```diff
@@ -222,14 +222,15 @@ let make wheel config pipeline (doc : DidOpenTextDocumentParams.t)
...
-  | exception Text_document.Invalid_utf8 ->
+  | exception Text_document.Invalid_utf message ->
     Log.log ~section:"warning" (fun () ->
         Log.msg
           "dropping update due to invalid utf8"
           [ ( "changes"
             , Json.yojson_of_list
                 TextDocumentContentChangeEvent.yojson_of_t
                 changes )
+          ; ("message", `String message)
```
ulugbekna, line 233: I wonder whether this log message could be a bit more helpful if we reported the context around the exception - on what document, after/on which changeEvent and with what doc state the error occurred

lsp/src/text_document.ml
```diff
@@ -32,8 +32,9 @@ let find_utf8_pos =
     if char = 0 then Uutf.decoder_byte_count dec
     else
       match Uutf.decode dec with
-      | `Malformed _ | `Await -> raise Invalid_utf8
-      | `End -> assert false
+      | `Malformed m -> raise (Invalid_utf m)
```
ulugbekna, line 35: When `Uutf.decode` returns `Malformed s`, what's `s` here? Just bytes read with `Uutf.Manual.src` call?

ulugbekna (approved):

lsp/src/text_document.ml
rgrinberg: The malformed byte sequence.

lsp/src/text_document.ml
```diff
@@ -32,8 +32,9 @@ let find_utf8_pos =
...
+      | `Await -> raise (Invalid_utf "not enough input")
+      | `End -> raise (Invalid_utf "not enough input")
```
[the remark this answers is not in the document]
rgrinberg: I'm not even sure how to trigger these errors at the moment, so I hate to spend time improving them until I'm certain they have some use.

rgrinberg:
I improved the messages somewhat. Feel free to improve things further.

PR #1026, opened 2023-01-25
refactor(lsp): do not generate texts eagerly

rgrinberg:
Instead of generating the text string immediately, we wait until it's
actually requested by the caller.

The end result is that subsequent edits will not force additional
strings needlessly

### DIFF
--- a/lsp/src/text_document.ml
+++ b/lsp/src/text_document.ml
@@ -3,6 +3,7 @@ open Import
 include struct
   open Types
   module DidOpenTextDocumentParams = DidOpenTextDocumentParams
+  module DocumentUri = DocumentUri
   module Range = Range
   module TextDocumentItem = TextDocumentItem
   module TextDocumentContentChangeEvent = TextDocumentContentChangeEvent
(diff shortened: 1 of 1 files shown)


rgrinberg:
In particular, our debouncing of diagnostics will take advantage of this optimization

ulugbekna (approved):
I like the goal of the PR, and the changes look correct.

The fact that the TextDocumentItem's fields are inlined seems to make it a bit harder to read, as now Text_document.t takes on more things to care about.

> our debouncing of diagnostics will take advantage of this optimization

Would you mind explaining a bit how?

rgrinberg:
> The fact that the TextDocumentItem's fields are inlined seems to make it a bit harder to read, as now Text_document.t takes on more things to care about.

Yes, but it's inevitable because we only want the text field to be mutable - not the other fields.

> Would you mind explaining a bit how?

If we have more than 1 content change notifications arriving within our debouncing interval (0.25 seconds?), then those change notifications will not create intermediate strings for `text`. Only the final `text` that will be used to compute the diagnostics will be created.

PR #1037, opened 2023-03-02
Disable redundant type annotation

### LINKED ISSUE #1036: Redundant "Type-annotate" action
After running type annotation once, the code action is not disabled:
Screenshot 2023-03-02 at 16 13 08

This encourages user to create a nested type annotation:
Screenshot 2023-03-02 at 16 13 21

I think it would be a good idea to disable this action for code that is already annotated.

3Rafal:
This fixes #1036 

The constrained pattern from example code
```
let my_fun (x : int) = x
```
is represented as below typedtree:
```
  structure_item (scratch.ml[1,0+0]..scratch.ml[1,0+24])
    Tstr_value Nonrec
    [
      <def>
        pattern (scratch.ml[1,0+4]..scratch.ml[1,0+10])
          Tpat_var \"my_fun/273\"
        expression (scratch.ml[1,0+11]..scratch.ml[1,0+24]) ghost
          Texp_function
          Nolabel
          [
            <case>
              pattern (scratch.ml[1,0+12]..scratch.ml[1,0+13])
                extra
                  Tpat_extra_constraint
                  core_type (scratch.ml[1,0+16]..scratch.ml[1,0+19])
                    Ttyp_constr \"int/1!\"
                    []
                Tpat_alias \"x/275\"
                pattern (scratch.ml[1,0+12]..scratch.ml[1,0+13])
                  Tpat_any
              expression (scratch.ml[1,0+23]..scratch.ml[1,0+24])
                Texp_ident \"x/275\"
          ]
    ]
]
```

Which means that we have to explicitly check for nested patterns of any and alias.


rgrinberg:
LGTM. A test would be nice as well.

3Rafal:
@rgrinberg, yes. I made this as a draft because there was no tests and changelog description. Just added them and now I think it's good. :)

ocaml-lsp-server/test/e2e/__tests__/textDocument-codeAction.test.ts
```diff
@@ -478,6 +478,76 @@ let f (x : t) = x
     `);
   });
 
+  it("does not type-annotate already annotated argument", async () => {
```
rgrinberg, line 481: Could you try porting this test to OCaml in e2e-new?

The typescript suite is our legacy test suite. We haven't yet ported it to all to OCaml, but we'd like to stop adding new tests at least.

3Rafal: Sure thing.
The funny thing is, that with the new test harness I get weird "destruct" action.
For argument x in function `let f (x : int) = 1` it wants to destruct it into:
`((0 as x) : int) | ((_ as x) : int`
It's unrelated to my PR and I've managed to reproduce this behavior on master branch:
image

I'm not sure why there is a difference between results for e2e and e2e-new, but I guess it's something we should investigate.

3Rafal: @rgrinberg , sure! I have ported them. After this PR gets merged I will port other type-annotate as well, so that they're all in one place. I already have the filtering functionality, so it should be a simple refactor.

Issue #594, opened 2022-01-28
Add code action "Remove type annotation" 


ulugbekna:
We have a code action to add type annotation, it would be useful to have a dual code action "remove type annotation". 

One can often find themselves annotating code when first prototyping something; the annotations are unnecessary afterwards, but removing them manually isn't convenient, especially for function arguments.

PR #1039, opened 2023-03-06
feat(ocamllsp): Code action "Remove type annotation"

3Rafal:
As specified in #594

### DIFF
--- a/ocaml-lsp-server/src/code_actions/action_remove_type_annotation.ml
+++ b/ocaml-lsp-server/src/code_actions/action_remove_type_annotation.ml
@@ -38,9 +38,11 @@ let get_source_text doc (loc : Loc.t) =
 
 let code_action_of_type_enclosing uri doc (loc, constr_loc) =
   let open Option.O in
-  let+ my_text = get_source_text doc loc in
+  let+ src_text = get_source_text doc loc in
   let edit : WorkspaceEdit.t =
-    let textedit : TextEdit.t = { range = Range.of_loc (Loc.union loc constr_loc); newText = my_text } in
+    let textedit : TextEdit.t =
+      { range = Range.of_loc (Loc.union loc constr_loc); newText = src_text }
+    in
     let version = Document.version doc in
     let textDocument =
       OptionalVersionedTextDocumentIdentifier.create ~uri ~version ()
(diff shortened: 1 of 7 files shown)


ocaml-lsp-server/src/code_actions/action_remove_type_annotation.ml
```diff
@@ -0,0 +1,77 @@
...
+    | Typedtree.Texp_constraint _, loc, _ -> Some loc
+    | Typedtree.Texp_coerce (Some { ctyp_loc; _ }, _), _, _ -> Some ctyp_loc
+    | _ -> None
+  in
+  let is_pat_constrained = function
+    | Typedtree.Tpat_constraint _, loc, _ -> Some loc
+    | _ -> None
+  in
+  let is_valid loc p extras =
+    match extras |> List.rev |> List.find_map ~f:p with
```
rgrinberg, line 19: Could you leave a comment as to why we're searching from the end?

ocaml-lsp-server/test/e2e-new/code_actions.ml
```diff
@@ -153,3 +159,152 @@ let f x = (1 : int :> int)
...
+type t = Foo of int | Bar of bool
+let f (x : t) = Foo x
+|ocaml}
+  in
+  let range =
+    let start = Position.create ~line:2 ~character:7 in
+    let end_ = Position.create ~line:2 ~character:8 in
+    Range.create ~start ~end_
+  in
+  print_code_actions source range ~filter:find_remove_annotation_action;
```
rgrinberg, line 175: An idea that's unrelated to this PR:

It would be nice to actually apply the code actions and print the contents of the document. It would be much easier to verify the actions are correct this way.

ocaml-lsp-server/src/code_actions/action_remove_type_annotation.ml
3Rafal: Sure. I hope it's clear enough. :)

ocaml-lsp-server/test/e2e-new/code_actions.ml
3Rafal: Yes, I was thinking about this too while writing these tests. Maybe we should create an issue describing this idea?

rgrinberg: Sounds good.

Issue #802, opened 2022-07-26
Failed to poll dune registry


mattiasdrp:
I don't know where I should report this issue but everytime I start the OCaml lsp server with emacs I have the following error:

```
failed to poll dune registry. Unix.Unix_error(Unix.ENOENT, "stat", "/home/mattias/.local/share/dune/rpc")
```

rgrinberg:
It's expected. If you don't use dune in watch mode, you'll get an error like this. It's harmless and ocamllsp is working as intended.

mattiasdrp:
Thanks for the answer. Is there a way to use it in watch mode ? Something that should be activated from emacs lsp mode? If it's not bothering you and you have them already, would you be able to point me to resources explaining how to do it?

rgrinberg:
It's just a matter of running dune with the `-w` flag. For example `$ dune build @check -w`.

mattiasdrp:
But is that call made by the ocaml-lsp-server or the client (in my case lsp-mode)?

rgrinberg:
What do you mean by call?

mattiasdrp:
Okay so currently there is no way for my lsp client to benefit from dune running in watch mode because this will be a process running in my terminal with no connection to my lsp client, if I understand correctly

rgrinberg:
There is a benefit for your lsp client because the connection between dune and ocamllsp will be established for you automatically no matter what terminal you will run dune in watch mode.

mattiasdrp:
Oh so I can start dune watch process in background when I start ocamllsp?

rgrinberg:
Sure. You can start it in in the background, foreground, turn dune on and off, and everything should work automatically.

mattiasdrp:
Great, thanks a lot :-)

askvortsov1:
@rgrinberg Would you be open to adding "Maybe you are not running `dune` in watch mode?" to the error on https://github.com/ocaml/ocaml-lsp/blob/a2fe6ffa00c422610d693ef761d92c769a139b27/ocaml-lsp-server/src/dune.ml#L646? I ran into this while looking into another issue, and it isn't immediately clear that this is expected behavior.

I'd be happy to send a quick PR.

rgrinberg:
Sure, that makes sense.

PR #1041, opened 2023-03-06
feat: improve error message when can't poll dune registry

askvortsov1:
The most common cause of this issue is not running `dune` in watch mode; therefore, there is no registry to be polled. The current error message doesn't make it clear that this behavior is expected.

We probably don't want to make polling conditional on existence of the socket, since the current implementation allows users to run `dune --watch` without restarting the lsp server.

Expanding the error message makes it clear that this behavior is intentional, and suggests the "fix".

See https://github.com/ocaml/ocaml-lsp/issues/802#issuecomment-1456883356

### DIFF
--- a/ocaml-lsp-server/src/dune.ml
+++ b/ocaml-lsp-server/src/dune.ml
@@ -643,7 +643,10 @@ let poll active last_error =
       | `Skip -> Fiber.return ()
       | `Print ->
         let message =
-          sprintf "failed to poll dune registry. %s" (Printexc.to_string exn)
+          sprintf
+            "failed to poll dune registry. %s \n\
+             Maybe you are not running dune in watch mode?"
+            (Printexc.to_string exn)
         in
         active.config.log ~type_:MessageType.Warning ~message
     in


ocaml-lsp-server/src/dune.ml
```diff
@@ -643,7 +643,10 @@ let poll active last_error =
       | `Skip -> Fiber.return ()
       | `Print ->
         let message =
-          sprintf "failed to poll dune registry. %s" (Printexc.to_string exn)
+          sprintf
+            "failed to poll dune registry. %s \n\
+             Maybe you are not running dune in watch mode?"
```
rgrinberg, line 648: Should the message come before the exception perhaps?

askvortsov1: That's how I had it at first, but "error description -> exception -> potential solution" seemed like a more natural order. Otherwise, the exception feels randomly stuck onto the end.

rgrinberg:
Merged. Thanks.

PR #1045, opened 2023-03-08
jsonrpc: add `Custom` constructor for error codes

### LINKED ISSUE #1044: jsonrpc: allow arbitrary error codes
Not an issue with ocaml-lsp specifically but just the general jsonrpc library published from here.

Currently JSON-RPC error codes are just a fixed enum with standard codes from the JSON-RPC and LSP specifications:
https://github.com/ocaml/ocaml-lsp/blob/0c5651cd4adb83623962eab1eaadf28d521b7ca5/jsonrpc/src/jsonrpc.ml#L102-L116

This makes it impossible to emit or parse JSON-RPC packets with any other error codes, although the JSON-RPC specification allows applications to do so using custom values. It would be great if the jsonrpc library allowed the same.

Changing the error code type to `int` means breaking changes though. Standard error codes could be provided via extra value definitions to simplify emitting errors, but it won't be as convenient for pattern matching on parsing.

sim642:
Closes #1044.


jsonrpc/src/jsonrpc.mli
```diff
@@ -93,6 +93,7 @@ module Response : sig
         | ServerCancelled
         | ContentModified
         | RequestCancelled
+        | Custom of int
```
rgrinberg, line 96: Could you rename this to `Other`? `Custom` is a fine name, but we use `Other` elsewhere in lsp

sim642: Sure, done!

Issue #911, opened 2022-11-05
Respect Client Capabilities


rgrinberg:
List of capabilities we don't currently respect (not exhaustive):

## Code Actions

- [ ] codeActionLiteralSupport
- [ ] isPreferredSupport

## Completions

- [ ] deprecatedSupport
- [ ] tagSupport
- [ ] resolveSupport

jfeser:
See also #1138

PR #1046, opened 2023-03-08
fix(ocamllsp): Respect codeActionLiteralSupport capability

3Rafal:
Mentioned in #911.

According to [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/):
```
/**
	 * The server provides code actions. The `CodeActionOptions` return type is
	 * only valid if the client signals code action literal support via the
	 * property `textDocument.codeAction.codeActionLiteralSupport`.
	 */
	codeActionProvider?: boolean | [CodeActionOptions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionOptions);
```

When client doesn't support code action literal, we return boolean instead.

I didn't modify returned code actions based on kinds supported by client, because client is supposed to handle actions that it doesn't support:
```
 /**
	 * The client supports code action literals as a valid
	 * response of the `textDocument/codeAction` request.
	 *
	 * @since 3.8.0
	 */
	codeActionLiteralSupport?: {
		/**
		 * The code action kind is supported with the following value
		 * set.
		 */
		codeActionKind: {

			/**
			 * The code action kind values the client supports. When this
			 * property exists the client also guarantees that it will
			 * handle values outside its set gracefully and falls back
			 * to a default value when unknown.
			 */
			valueSet: [CodeActionKind](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionKind)[];
		};
	};
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,10 @@
 
 ## Fixes
 
+- Respect codeActionLiteralSupport capability (#1046)
+
 - Fix a document syncing issue when utf-16 is the position encoding (#1004)
+
 - Disable "Type-annotate" action for code that is already annotated.
   ([#1037](https://github.com/ocaml/ocaml-lsp/pull/1037)), fixes
   [#1036](https://github.com/ocaml/ocaml-lsp/issues/1036)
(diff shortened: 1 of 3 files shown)


rgrinberg:
LGTM.

Could you add a CHANGES entry?

3Rafal:
@rgrinberg , done.

I wonder if we can add a changelog check similar to https://github.com/tarides/dune-release/blob/main/.github/workflows/changelog.yml , so that it's easier to remember.

Issue #801, opened 2022-07-25
Fix type annotation on function name in declaration


ulugbekna:
We shouldn't produce syntactically incorrect code, e.g., 

<img width="449" alt="image" src="https://user-images.githubusercontent.com/16353531/180753196-a4461ef9-1716-42b1-a5a7-beebae2985a5.png">

3Rafal:
I have tried to solve this with #1048, but found a problem when dealing with higher order functions.

Example:
We have two functions:
```
let my_fun x = fun y z -> 1

let my_fun' x y z = 1
```

We would probably want to annotate them like this:
```
let my_fun x : `a -> `b -> int = fun y z -> 1

let my_fun' x y z : int = 1
```
or this:
```
let my_fun x = (fun y z -> 1) : `a -> `b -> int

let my_fun' x y z = 1 : int
```

but they are represented almost identically in both parsetree and typedtree.
Parsetree dump diff:
```
-  structure_item (*buffer*[1,0+0]..[1,0+29])
+  structure_item (*buffer*[3,31+0]..[3,31+21])
     Pstr_value Nonrec
     [
       <def>
-        pattern (*buffer*[1,0+4]..[1,0+10])
-          Ppat_var \"my_fun\" (*buffer*[1,0+4]..[1,0+10])
-        expression (*buffer*[1,0+11]..[1,0+29]) ghost
+        pattern (*buffer*[3,31+4]..[3,31+11])
+          Ppat_var \"my_fun'\" (*buffer*[3,31+4]..[3,31+11])
+        expression (*buffer*[3,31+12]..[3,31+21]) ghost
           Pexp_fun
           Nolabel
           None
-          pattern (*buffer*[1,0+11]..[1,0+12])
-            Ppat_var \"x\" (*buffer*[1,0+11]..[1,0+12])
-          expression (*buffer*[1,0+15]..[1,0+29])
+          pattern (*buffer*[3,31+12]..[3,31+13])
+            Ppat_var \"x\" (*buffer*[3,31+12]..[3,31+13])
+          expression (*buffer*[3,31+14]..[3,31+21]) ghost
             Pexp_fun
             Nolabel
             None
-            pattern (*buffer*[1,0+20]..[1,0+21])
-              Ppat_var \"y\" (*buffer*[1,0+20]..[1,0+21])
-            expression (*buffer*[1,0+22]..[1,0+28]) ghost
+            pattern (*buffer*[3,31+14]..[3,31+15])
[...]

PR #1048, opened 2023-03-10
WIP: Fix type-annotate action for functions

3Rafal:
As described in #801

### DIFF
--- a/ocaml-lsp-server/src/code_actions/action_type_annotate.ml
+++ b/ocaml-lsp-server/src/code_actions/action_type_annotate.ml
@@ -82,7 +82,7 @@ let code_action uri doc str_fmt (env, typ, loc) =
       let open Merlin_analysis in
       let module Printtyp = Type_utils.Printtyp in
       Printtyp.wrap_printing_env env ~verbosity:(Lvl 0) (fun () ->
-          Printtyp.shared_type_scheme ppf ty)
+          Printtyp.type_expr ppf ty)
     in
     Format.fprintf ppf "%a%!" (pp_type env) typ;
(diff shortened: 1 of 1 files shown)


ocaml-lsp-server/src/code_actions/action_type_annotate.ml
```diff
@@ -59,6 +103,35 @@ let code_action_of_type_enclosing uri doc (loc, typ) =
     ~isPreferred:false
     ()
 
+let code_action_of_type_enclosing' uri doc (loc, arg_len, typ) =
+  let open Option.O in
+  let+ original_text = get_source_text doc loc in
+  let arrow = " -> " in
+  let typ' =
+    Str.split (Str.regexp arrow) typ
```
rgrinberg, line 111: `Str` isn't thread safe so we avoid it. Try using `Re.split` or something similar.

rgrinberg: It's also not ideal that we need to deal with strings here at all btw. Shouldn't it be possible to use the type tree representation to drop the args instead?

3Rafal: Hey Rudi, it's just a draft and work in progress. I have opened a PR so that it's visible that I'm working on this feature.
Thanks for your suggestions (will keep them in mind), but I'm not sure if it's worth it to do review now. 🙂

rgrinberg: Sure thing. Some people like the feedback early, others prefer it all at the end. I can go either way

Issue #1034, opened 2023-02-21
Wrong locations in semantic highlighting when using preprocessors


jboillot:
Hello!

I was using ocaml-lsp with cppo, a C-style preprocessor for OCaml while I noticed the semantic highlighting in files containing cppo code was broken:
![Screenshot from 2023-02-21 15-41-18](https://user-images.githubusercontent.com/52931370/220375592-194f00d0-247a-4921-b06e-493cfa14ae33.png)

After some investigation in the ocaml-lsp code I found this PR #932 that corrected the function `lident` https://github.com/ocaml/ocaml-lsp/blob/45b3198c718421d0f0514132646250b4eb4a6acf/ocaml-lsp-server/src/semantic_highlighting.ml#L354 that parses long identifiers.
In order to remember spaces that may appear in long identifiers, it ignores the long identifier object given as a parameter (that already has spaces trimmed) and instead reads again the source `M.source` using the function `source_excerpt` https://github.com/ocaml/ocaml-lsp/blob/45b3198c718421d0f0514132646250b4eb4a6acf/ocaml-lsp-server/src/semantic_highlighting.ml#L341 .

The problem is that, while using preprocessor directives, the location of the long identifier given as an argument is invalid (the location is specific to the code already preprocessed). Then the semantic highlighting returned is invalid.

I have tested to add to the `lident` function some logs after computing the source excerpt:
```
let lid = source_excerpt loc in
let lid2 = join (Longident.flatten lid2) in
debug (Printf.sprintf "I read \"%s\" while I could have read \"%s\"" lid lid2);
```

And here is an extract of the result:
```
I read "ec" while I could have read "()"
I read "e you" while I could have read "false"
I read "re" while I could have read "&&"
I read "warne" while I could have read "debug"
I read " i" while I could have read "()"
```

I am not sure what could be acceptable to do. Would it be possible to keep track of the source location? Or to keep the input text in the long identifier structure?

Thank you for reading!

PS: Maybe you have some clues @ulugbekna since you did the PR? :)

jboillot:
Fixed by #1049

PR #1049, opened 2023-03-10
fix(semantic highlighting): use preprocessed source to highlight long identifiers

jboillot:
Hello,

This PR follows up and solves issue #1034 where usage of preprocessors breaks semantic highlighting.

After a few minutes of investigation with @thierry-martinez we saw that, in order to parse long identifiers (e.g. `Foo.Bar.x`), ocaml-lsp reads again the source pre-preprocessor (`raw_source`) to semantic highlight the tokens in long identifiers (following PR #932).

The problem is that locations in the AST do not correspond to the ones in `raw_source`, but to the ones in `source`.

I'm not sure this modification could cause issues, but this source is only used to retrieve the source code of long identifiers. All tests still pass.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,10 @@
   ([#1037](https://github.com/ocaml/ocaml-lsp/pull/1037)), fixes
   [#1036](https://github.com/ocaml/ocaml-lsp/issues/1036)
 
+- Fix semantic highlighting of long identifiers when using preprocessors
+  ([#1049](https://github.com/ocaml/ocaml-lsp/pull/1049), fixes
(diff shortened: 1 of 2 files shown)


ulugbekna (approved):
This makes sense. 

Possibly, it also makes sense to ignore AST nodes, for which `ghost_loc` is true

jboillot:
Thank you for your feedback, in the latest commit I've added the modification you suggested that ignores semantic tokens at locations for which `loc_ghost` is true.

ulugbekna (approved):

ulugbekna:
Thanks heaps for investigation and the PR! :-)

rgrinberg:
Thanks. Could you add a CHANGES entry?

Issue #1056, opened 2023-03-28
DocumentSelector is of wrong type


qua3k:
Hi all, I’m using ocamllsp version 1.15.1-5.0 on macOS with Sublime Text, and I’m getting errors on the `client/registerCapability` response to `initialize`. I think it’s because documentSelector should be a `DocumentFilter list` instead of a `DocumentFilter list list`, but the ocamllsp code appears to be [correct](https://github.com/ocaml/ocaml-lsp/blob/66539c162dc49010d73535a40c439bc5a3a2b39f/ocaml-lsp-server/src/ocaml_lsp_server.ml#L290-L293). I’m not entirely sure where in the code is wrong, and would appreciate some help.

See example response here:
```python
[...]

PR #1058, opened 2023-03-30
Attempt to update metaModel

qua3k:
I'm not exactly sure how to generate the types. Running `dune runtest --auto-promote && dune fmt --auto-promote` just deleted a bunch of code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,10 @@
   ([#1037](https://github.com/ocaml/ocaml-lsp/pull/1037)), fixes
   [#1036](https://github.com/ocaml/ocaml-lsp/issues/1036)
 
+- Fix semantic highlighting of long identifiers when using preprocessors
+  ([#1049](https://github.com/ocaml/ocaml-lsp/pull/1049), fixes
+  [#1034](https://github.com/ocaml/ocaml-lsp/issues/1034))
+
 ## Features
 - Add "Remove type annotation" code action. (#1039)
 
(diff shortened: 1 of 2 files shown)


qua3k:
Fixes #1056

rgrinberg:
you can run `dune build @cinaps --auto-promote; dune build @fmt --auto-promote` to regenerate from the metamodel.

qua3k:
> you can run `dune build @cinaps --auto-promote; dune build @fmt --auto-promote` to regenerate from the metamodel.

I did that, but there is no difference.

qua3k:
Actually, I think that the changes I made are really broken. The type alias for Uri to string was removed, so there was some trouble where it was undefined, and I tried to have it directly be defined as a string, but that broke a bunch of other code that expected a Uri.t.

rgrinberg:
How about we start with porting only the change to metamodel that's required to fix the issue with DocumentSelector? Later, we can worry about updating everything else.

qua3k:
> How about we start with porting only the change to metamodel that's required to fix the issue with DocumentSelector? Later, we can worry about updating everything else.

Okay, so I reset my branch to HEAD@master but I get the same error after running `dune build @cinaps --auto-promote; dune build @fmt --auto-promote`

How did you generate this previously?

```
duckie@macbook ocaml-lsp λ make all                                  
dune build @all
File "lsp/src/types.ml", line 1345, characters 42-53:
1345 |     Json.To.literal_field "kind" "delete" yojson_of_t t
                                                 ^^^^^^^^^^^
Error: Unbound value yojson_of_t
Hint: Did you mean yojson_of_int?
Hint: If this is a recursive definition,
you should add the 'rec' keyword on line 1344
make: *** [all] Error 1
```

PR #1067, opened 2023-04-24
double opening the same document

rgrinberg:
It seems like neovim sometimes opens the same document twice under certain circumstances. We allow this behavior since it's mostly harmless.


ocaml-lsp-server/test/e2e-new/document_flow.ml
```diff
@@ -0,0 +1,55 @@
+open Test.Import
+
+let%expect_test "double open a document" =
```
3Rafal, line 3: I'd change the name of the text to something that's actually describing an expected behavior.
E.g. "[...] should allow double opening the same document"

Current description requires reading all test implementation to figure out expected outcome

ocaml-lsp-server/test/e2e-new/document_flow.ml
```diff
@@ -0,0 +1,55 @@
...
+  let diagnostics = Fiber.Mvar.create () in
```
3Rafal, line 4: an idea to consider in the future:
I've looked up an implementation of code_action.ml from e2e-new and the test environment setup looks pretty similar. I wonder if we can abstract this part so that we can write less boilerplate.

rgrinberg: if we improve our client implementation, i would imagine some of this code to be simpler.

PR #1068, opened 2023-04-25
fix: update DocumentSelector definition

rgrinberg:


lsp/src/types.ml
```diff
@@ -4878,85 +4878,19 @@ module DocumentFilter = struct
...
 
 module DocumentSelector = struct
-  type t = [ `String of string | `DocumentFilter of DocumentFilter.t ] list
-  [@@deriving_inline yojson]
+  type t = DocumentFilter.t list [@@deriving_inline yojson]
 
   let _ = fun (_ : t) -> ()
 
   let t_of_yojson =
     (let _tp_loc = "lsp/src/types.ml.DocumentSelector.t" in
```
3Rafal, line 4886: this let binding seems no longer needed

lsp/src/types.ml
```diff
@@ -4878,85 +4878,19 @@ module DocumentFilter = struct
...
-             | _ as yojson ->
-               Ppx_yojson_conv_lib.Yojson_conv_error.unexpected_stag
-                 _tp_loc
-                 yojson
-           with Ppx_yojson_conv_lib.Yojson_conv_error.No_variant_match ->
-             Ppx_yojson_conv_lib.Yojson_conv_error.no_matching_variant_found
-               _tp_loc
-               yojson)
-         t
+     fun t -> list_of_yojson DocumentFilter.t_of_yojson t
```
3Rafal, line 4887: nit:
I guess we can eta reduce to shorter version:
```ocaml
  let t_of_yojson =
    (list_of_yojson DocumentFilter.t_of_yojson
      : Ppx_yojson_conv_lib.Yojson.Safe.t -> t)
```

similarly with `yojson_of_t`

3Rafal (approved):

qua3k:
This PR works well but when I attempt to generate types.ml from the metamodel on my own I don’t get the same output

qua3k (approved):

lsp/src/types.ml
rgrinberg: It's not, but this part of the file is generated. So we would need to modify the code generator to make this cleaner.

lsp/src/types.ml
rgrinberg: Yes, but this would require tweaking the code generator which is a lot more work. Feel free to do it though.

rgrinberg:
> This PR works well but when I attempt to generate types.ml from the metamodel on my own I don’t get the same output

The code generation process is a bit janky at the moment. You need to run `@cinaps`, `@lint`, and `@fmt` in order to get the generated code. I need to improve dune first to make this simpler.

PR #1070, opened 2023-04-28
feature: Use Merlin as a libray

voodoos:
Thanks to @3Rafal's work we recently made multiple changes on Merlin's side to close the gap with the fork used by `ocaml-lsp`. Once everything is merged it should make Merlin-as-a-library usable by ocaml-lsp.

This PR intends to show the expected diff on ocaml-lsp side to gave us an opportunity to evaluate and discuss these changes and maybe do some final adjustments before moving on.

I made these changes to the `master` branch of `ocaml-lsp` that targets OCaml 4.14 since Merlin's patches target 4.14 too. However they should be similar for branches supporting 5.0 and 5.1. Furthermore, once we move to a clean dependency over `merlin-lib` we should be able to maintain a unique branch in `ocaml-lsp` with support for throth 4.14, 5.0 and 5.1 since merlin API's compatibility is not bound to compiler versions.

cc @tmattio @engil


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -863,8 +864,50 @@ let stream_of_channel : Lsp.Cli.Channel.t -> _ = function
...
+  let cmd = Format.sprintf "%s %s" prog args in
+
+  let prog = "/bin/sh" in
+  let argv = [ "sh"; "-c"; cmd ] in
+  let stdin = Unix.openfile "/dev/null" [ Unix.O_RDONLY ] 0x777 in
+  let stdout =
+    match stdout with
+    | Some file -> Unix.openfile file [ Unix.O_WRONLY ] 0x664
+    | None ->
+      (* Runned programs should never output on stdout since it is the channel
```
3Rafal, line 884: nit:
I think "output to stdout" is correct way to put it

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -863,8 +864,50 @@ let stream_of_channel : Lsp.Cli.Channel.t -> _ = function
     let sockaddr = Unix.ADDR_INET (Unix.inet_addr_loopback, port) in
     socket sockaddr
 
+(* Merlin uses [Sys.command] to run preprocessors and ppxes. We provide an
+   alternative version using spawn for unixes.*)
```
3Rafal, line 868: nit:
can we capitalize "spawn", or even change it to "Spawn library", so that it's more straightforward?

3Rafal (approved):

voodoos:
@rgrinberg what do you think about the current api ? I can cut  Merlin release if we think it is good.

voodoos:
@rgrinberg any thoughts ? I will probably make a release of Merlin next week or the one after. If we want to make (small) changes to the api it would be nice to do them before :-)

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -863,8 +864,50 @@ let stream_of_channel : Lsp.Cli.Channel.t -> _ = function
...
+    | WSIGNALED _ -> -1
+    | WSTOPPED _ -> -1
+  in
+  Unix.close stdin;
+  Unix.close stdout;
+  `Finished res
+
+let run_in_directory =
+  (* Merlin has specific stubs for Windows, we reuse them *)
+  let for_windows = !Merlin_utils.Std.System.run_in_directory in
```
rgrinberg, line 905: Can we stick to spawn on windows? it works quite well there.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -863,8 +864,50 @@ let stream_of_channel : Lsp.Cli.Channel.t -> _ = function
...
+    ?stderr:_ () =
+  (* Currently we assume that [prog] is always quoted and might contain
+     arguments such as [-as-ppx]. This is due to the way Merlin gets its
+     configuration. Thus we cannot rely on [Filename.quote_command]. *)
+  let args = String.concat ~sep:" " @@ List.map ~f:Filename.quote args in
+  let cmd = Format.sprintf "%s %s" prog args in
+
+  let prog = "/bin/sh" in
+  let argv = [ "sh"; "-c"; cmd ] in
+  let stdin = Unix.openfile "/dev/null" [ Unix.O_RDONLY ] 0x777 in
```
rgrinberg, line 879: Should be `nul` on windows.

rgrinberg:
To summarize the current situation with executing ppx:

1. We obtain the ppx command and argument from dune. This step is already lossy and we only get a shell command from dune rather than the raw command with arguments.
2. We pass this ppx command to merlin via our configuration mechanism
3. Whenever merlin wants to preprocess, it uses this ppx command as input to the process executing hook.

If we fix step 1 in dune, it will not be possible to make use of this fix without changing this API further. That's a sign that this API is inflexible, and I would also claim that it is cumbersome and needlessly indirect as well. Regardless, it's clear that we will have to change it further.

Given that this PR doesn't offer any user facing advantages to the user, there's no rush in merging it and immediately introducing constraints between lsp and merlin that we know are going to be broken. Until there's a user facing need, we can keep the work in a branch and maintain the flexibility to change things without disturbing anyone down stream.

ocaml-lsp-server/src/ocaml_lsp_server.ml
voodoos: I think that currently it is not the case that ocaml-lsp uses spawn on windows: 

https://github.com/rgrinberg/merlin/blob/91fd966fdc00e259c2488127961daf3b7986a736/src/ocaml/driver/pparse.ml#L44-L65

I simply replicated (hopefully) the current behaviour.

voodoos:
> If we fix step 1 in dune, it will not be possible to make use of this fix without changing this API further. That's a sign that this API is inflexible, and I would also claim that it is cumbersome and needlessly indirect as well. Regardless, it's clear that we will have to change it further.

Could you elaborate on that ?
If OCaml-LSP receives configuration from Dune in the form ("program", "argument array") then the proposed "api" should be reusable. It is almost identical to Spawn's signature.

> Until there's a user facing need, we can keep the work in a branch and maintain the flexibility to change things without disturbing anyone down stream.

I am not against keeping it in a separated branch, but there would be a user-facing advantage: Merlin is maintained for the last 3 versions of OCaml. I don't know what is the status of OCaml-LSP but you would get that support for free.

Another improvement for users and devs I can think of (and I think it is an important one) is that as soon as a new version of Merlin would be released, users of OCaml-LSP would benefit from it without having to cut a release. (unless, of course, there are API changes).

ddickstein:
To be compatible with the latest version there will also need to be some changes in `merlin_config.ml` to account for the new `USE_PPX_CACHE` directive.

ocaml-lsp-server.opam.template
```diff
@@ -9,3 +9,6 @@ build: [
     "--release"
   ]
 ]
+pin-depends: [
+  ["merlin-lib.dev" "git+https://github.com/voodoos/merlin#4db2f27de0e04abc170bb9344d6a719912657403"]
```
rgrinberg, line 13: Is the pin still necessary?

dune-project
```diff
@@ -66,7 +66,8 @@ possible and does not make any assumptions about IO.
   (csexp (>= 1.5))
   (ocamlformat-rpc-lib (>= 0.21.0))
   (odoc :with-doc)
-  (ocaml (and (>= 4.14) (< 4.15)))))
+  (ocaml (and (>= 4.14) (< 4.15)))
+  merlin-lib))
```
rgrinberg, line 70: There should be minimum version here I believe.

voodoos: Yes, also I removed the constraint on OCaml version, since it is now driven by the versions supported by merlin-lib

ddickstein:
Can this PR also add tests that exercise the new `run_in_directory` implementation by using Merlin on a file w/ ppx?

voodoos:
> Can this PR also add tests that exercise the new `run_in_directory` implementation by using Merlin on a file w/ ppx?

I guess we could but again, I did not add a lot of new logic here, just moved things around :-)

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -865,8 +866,50 @@ let stream_of_channel : Lsp.Cli.Channel.t -> _ = function
...
     socket sockaddr
 
+(* Merlin uses [Sys.command] to run preprocessors and ppxes. We provide an
+   alternative version using the Spawn library for unixes.*)
+let run_in_directory ~prog ~prog_is_quoted:_ ~args ~cwd ?stdin:_ ?stdout
+    ?stderr:_ () =
+  (* Currently we assume that [prog] is always quoted and might contain
+     arguments such as [-as-ppx]. This is due to the way Merlin gets its
+     configuration. Thus we cannot rely on [Filename.quote_command]. *)
+  let args = String.concat ~sep:" " @@ List.map ~f:Filename.quote args in
```
rgrinberg, line 876: Why does we assume that `prog` is always quoted? Shouldn't we inspect `prog_is_quoted` to determine this?

rgrinberg:
@voodoos there's no such tests at the moment. Adding one should be straightforward except that we use ppx_expect instead of cram tests.

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -865,8 +866,50 @@ let stream_of_channel : Lsp.Cli.Channel.t -> _ = function
...
+  in
+  let _, status = Unix.waitpid [] pid in
+  let res =
+    match (status : Unix.process_status) with
+    | WEXITED n -> n
+    | WSIGNALED _ -> -1
+    | WSTOPPED _ -> -1
+  in
+  Unix.close stdin;
+  Unix.close stdout;
```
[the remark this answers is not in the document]
voodoos: Good catch, fixed in https://github.com/ocaml/ocaml-lsp/pull/1070/commits/ba88ca5fcd72c1ecbad946fd47645f335f88d1bd

ocaml-lsp-server/src/ocaml_lsp_server.ml
voodoos: Yes, we should have logic to handle both at some point, but as of today these are always "quoted" and the argument is always `true`. I don't think it is worth it to implement the `false` case until Merlin does use it. Since it will require a configuration protocol change it will break the API anyway. Do you prefer that I do test it's value and make an `failwith "Not implemented"` branch ?

rgrinberg: I think it would have made more sense to just drop the argument then.

voodoos: I agree that it would be cleaner, but we already have 3 released versions of Merlin with it. I think it can wait for the next round of API modification (for example when we will be ready to change how the PPX configuration is obtained). It should not cause any issue while, on the other side, releasing ocaml-lsp now with merlin as a library would provide immediate support for OCaml 5.1 and latest Merlin patches.

rgrinberg: Okay, let's leave it as a TODO then.

voodoos:
> ```
> +- The following actions failed
> | - build base v0.16.0
> +- 
> ```
> 
> Oh no, we got same problem with Windows build as the one we dealt with here: [ocamllabs/vscode-ocaml-platform#1110](https://github.com/ocamllabs/vscode-ocaml-platform/pull/1110)

Should be fixed soon... https://github.com/janestreet/base/issues/150#issuecomment-1587095269

rgrinberg:
@voodoos could you rebase? I've moved over your tests to a separate PR. Could you rebase?

voodoos:
> @voodoos could you rebase? I've moved over your tests to a separate PR. Could you rebase?

I did that yesterday but somehow a dune submodule appeared. It is now fixed. 
Not sure why `make nix-tests` works on my machine and not in the CI however...

voodoos:
@rgrinberg : Thanks to @julow I now understand that the issue is caused by moving to `nixpkgs` instead of `opam-nix`, the former does not have the latest version of `merlin-lib`. Should we revert that change ?

voodoos:
Right, I think it would be much more efficient if you do it :-)

voodoos:
@julow is already preparing a PR to add 4.9 to `nixpkgs` 🎉

PR #1071, opened 2023-04-29
coverage: add merlin deps

Alizter:
Signed-off-by: Ali Caglayan <alizter@gmail.com>


rgrinberg:
It doesn't need to be installed. The dependency comes from a submodule.

PR #1073, opened 2023-04-30
chore(lsp): fix incorrect comment

rgrinberg:

cc @ddickstein

### DIFF
--- a/lsp/src/text_document.ml
+++ b/lsp/src/text_document.ml
@@ -16,13 +16,10 @@ type invalid_utf = String_zipper.invalid_utf =
 
 exception Invalid_utf = String_zipper.Invalid_utf
 
-(* Text is received as UTF-8. However, the protocol specifies offsets should be
-   computed based on UTF-16. Therefore we reencode every file into utf16 for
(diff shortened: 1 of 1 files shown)


ddickstein (approved):
LGTM. But so I'm understanding this correctly, in the current world:

1. If you receive position information independent of a text document edit, how are you supposed to match those positions to the corresponding text? If they are utf-8, it's 1:1 since the text is always utf-8, but if they are utf-16 what should clients do?
2. If you want to send `TextEdit`s to the client in utf-16, what's the right way to do that?

ddickstein:
Actually, for (1), is the answer that nothing special is needed? Since positions are expressed in terms of character offsets and not in terms of bytes, I think they should apply just as well for any given encoding.

rgrinberg:
For 2., I would like to add conversion functions in Text_document that will take edits (or any other ranges) and convert them from utf8 to utf16.

I would say that the ideal API would make utf16 an invisible implementation detail for the user and make it easy to "pretend" that utf16 doesn't exist at all. I don't think we're there yet though.

PR #1074, opened 2023-04-30
feature: support --clientProcessId argument

rgrinberg:
It does nothing for now, but it's in the standard so it's better to at
least explicitly ignore it.

### DIFF
--- a/lsp/src/cli.ml
+++ b/lsp/src/cli.ml
@@ -11,10 +11,18 @@ module Arg = struct
     ; mutable port : int option
     ; mutable stdio : bool
     ; mutable spec : (string * Arg.spec * string) list
+    ; mutable clientProcessId : int option
     }
 
   let create () =
-    let t = { pipe = None; port = None; stdio = false; spec = [] } in
+    let t =
+      { pipe = None
+      ; port = None
+      ; stdio = false
+      ; spec = []
+      ; clientProcessId = None
+      }
+    in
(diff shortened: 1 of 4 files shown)


lsp/src/cli.ml
```diff
@@ -11,25 +11,39 @@ module Arg = struct
...
     let spec =
       [ ("--pipe", Arg.String (fun p -> t.pipe <- Some p), "set pipe path")
       ; ("--socket", Arg.Int (fun p -> t.port <- Some p), "set port")
       ; ("--stdio", Arg.Unit (fun () -> t.stdio <- true), "set stdio")
       ; ( "--node-ipc"
         , Arg.Unit (fun () -> raise @@ Arg.Bad "node-ipc isn't supported")
         , "not supported" )
+      ; ( "--clientProcessId"
+        , Arg.Int (fun pid -> t.clientProcessId <- Some pid)
+        , "set the pid of the lsp client" )
```
3Rafal, line 35: I wonder if we should add an explicit remark about client pid being ignored. Currently, it's only mentioned in PR description. Maybe we should add it here?

3Rafal (approved):

lsp/src/cli.ml
rgrinberg: Note that this code is part of our general purpose lsp library. So we can't make assumptions whether it's being used or not.

PR #1076, opened 2023-05-02
feature: add --port as a synonym to --socket

rgrinberg:
cc @ddickstein. The spec is quite confusing about this:

> socket: uses a socket as the communication channel. The port is passed as next arg or with --port=.

So I believe that makes `--port` a synonym.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,8 @@
 
 - Accept the `--clientProcessId` command line argument. (#1074)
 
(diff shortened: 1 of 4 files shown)


ddickstein (approved):
TIL the CLI is in the spec. Interpreting these as synonyms seems fine to me, but I also think it would be fine to just provide one of them. It might be clearer if `--socket` is described as "synonym for --port" so it's explicit that the two flags mean the same thing.

PR #1079, opened 2023-05-03
uri_dune_overlap: normalize case under Windows

nojb:
This is necessary to communicate with Dune on Windows (The case of `uri` was full lowercase in some experiments).


nojb:
Note: conjunction with https://github.com/ocaml/dune/pull/7666, this is enough to get `ocaml-lsp`/dune rpc interaction working on Windows.

rgrinberg:
Merged in #1080. Thank you. I knew there was something fishy about this function...

PR #1088, opened 2023-05-08
Refactor comment parsing to use odoc-parser and cmarkit

tmattio:
Omd 2.X doesn't support printing markdown, so OCaml LSP is blocked to Omd 1.X. This creates conflicts with codebases that have migrated to Omd 2.X (e.g. in ocamlorg).

To avoid conflicts for users of Omd 2.X, we migrate to cmarkit, which has a smaller dependency profile and supports printing to markdown.

Not complete yet, the support for tags is incomplete and I'll add a few more tests to make sure we're not breaking anything.

cc @tatchi @ddickstein

### DIFF
--- a/ocaml-lsp-server/src/doc_to_md.ml
+++ b/ocaml-lsp-server/src/doc_to_md.ml
@@ -1,5 +1,15 @@
 open Import
-open Cmarkit
+
+open struct
+  open Cmarkit
+  module Inline = Inline
+  module Meta = Meta
+  module Block_line = Block_line
+  module Link_definition = Link_definition
+  module Block = Block
+  module Layout = Layout
+  module Doc = Doc
+end
 
 (** TODO:
 
@@ -30,32 +40,31 @@ let rec inline_element_to_inline
     (inline : Odoc_parser.Ast.inline_element Odoc_parser.Loc.with_location) :
     Inline.t =
   match inline with
-  | Odoc_parser.Loc.{ value = `Space _; location } ->
+  | { value = `Space _; location } ->
     let meta = loc_to_meta location in
     Inline.Text (" ", meta)
-  | Odoc_parser.Loc.{ value = `Word w; location } ->
+  | { value = `Word w; location } ->
     let meta = loc_to_meta location in
     Inline.Text (w, meta)
-  | Odoc_parser.Loc.{ value = `Code_span c; location } ->
+  | { value = `Code_span c; location } ->
     let meta = loc_to_meta location in
     Inline.Code_span (Inline.Code_span.of_string c, meta)
   | Odoc_parser.Loc.{ value = `Raw_markup (Some "html", text); location } ->
     let meta = loc_to_meta location in
     Inline.Raw_html (Block_line.tight_list_of_string text, meta)
-  | Odoc_parser.Loc.{ value = `Raw_markup (_, text); location } ->
+  | { value = `Raw_markup (_, text); location } ->
     (* Cmarkit doesn't have constructors for backend other than HTML for inline
        raw markups, only for blocks. *)
     let meta = loc_to_meta location in
     Inline.Text (text, meta)
-  | Odoc_parser.Loc.{ value = `Styled (style, inlines); location } ->
+  | { value = `Styled (style, inlines); location } ->
     let text = inline_element_list_to_inlines inlines in
     let meta = loc_to_meta location in
     style_inline ~meta style text
-  | Odoc_parser.Loc.
-      { value = `Reference (_kind, _ref, _inlines); location = _location } ->
+  | { value = `Reference (_kind, _ref, _inlines); location = _location } ->
     (* TODO: add support for references *)
     Inline.Break (Inline.Break.make `Hard, Meta.none)
-  | Odoc_parser.Loc.{ value = `Link (link, inlines); location } ->
+  | { value = `Link (link, inlines); location } ->
     let text = inline_element_list_to_inlines inlines in
     let ref =
       `Inline (Link_definition.make ~dest:(link, Meta.none) (), Meta.none)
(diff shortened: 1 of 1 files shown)


tmattio:
also cc @dbuenzli as you might be interested in this. Certainly not as solid as [ocamlmark](https://github.com/ocaml-doc/odoc-parser/pull/15) in the current state, but this essentially implements the reverse transformations you implemented in the PR on odoc-parser.

ocaml-lsp-server/test/e2e/__tests__/textDocument-signatureHelp.ts
```diff
@@ -288,31 +337,93 @@ describe_opt("textDocument/completion", () => {
...
+
+              *Unordered list item 1
+              *Unordered list item 2
+              
+              ### Code Blocks
+              
+              \`\`\`
+              let square x = x * x
+              let result = square 3
+              let result = square 3
```
tmattio, line 384: @dbuenzli I'll investigate more to confirm this isn't a problem in my code, but at first sight, it looks like cmarkit is duplicating the last line of code blocks. I'll open an issue once I manage to get a proper repro, but pinging you in case this rings a bell.

ocaml-lsp-server/src/doc_to_md.ml
```diff
@@ -76,6 +76,13 @@ and inline_element_list_to_inlines inlines =
   let inlines = List.map ~f:inline_element_to_inline inlines in
   Inline.Inlines (inlines, Meta.none)
 
+(** Hack: Because of bug described here:
+    https://github.com/dbuenzli/cmarkit/issues/7 , we use only first block line.
+    It contains whole string, which seems good enough for our use cases. To be
+    removed when: https://github.com/dbuenzli/cmarkit/pull/8 gets merged and
+    released *)
+let block_lines_of_string str = [ List.hd (Block_line.list_of_string str) ]
```
dbuenzli, line 84: But then that's simply `Fun.id` with the bug no ?

dune-project
```diff
@@ -56,7 +56,7 @@ possible and does not make any assumptions about IO.
   ordering
   dune-build-info
   spawn
-  cmarkit
+  (cmarkit (= 0.1.0))
```
tmattio, line 59: cmarkit 0.2.0 is now on opam-repository, so we can depend on >= 0.2.0 and get rid of the temporary fix

tmattio: https://github.com/ocaml/opam-repository/pull/23781

ocaml-lsp-server/test/e2e/__tests__/textDocument-signatureHelp.ts
```diff
@@ -338,72 +338,90 @@ describe_opt("textDocument/completion", () => {
           documentation: {
             kind: "markdown",
             value: outdent`This is an example of a docstring that demonstrates various ocamldoc syntax features.
-    
-              #### Sections and Labels
```
tmattio, line 342: Heading level are broken

tmattio (commented):
@3Rafal I've rebased the PR on master. I left a few comments in the tests for things that appear to be broken - we'll need to fix them before the PR can be merged.

ocaml-lsp-server/src/doc_to_md.ml
tmattio: We used cmarkit 0.2.0 in the end (thanks for the release!), so we removed that workaround.

ocaml-lsp-server/src/doc_to_md.ml
```diff
@@ -124,12 +124,12 @@ let rec nestable_block_element_to_block
...
     Block.List (l, meta)
   | Odoc_parser.Loc.
-      { value = `Code_block (metadata, { value = code; location = _code_loc })
+      { value = `Code_block (metadata, { value = code; location = code_loc })
       ; location
       } ->
     let info_string =
       match metadata with
-      | None -> None
+      | None -> Some ("ocaml", loc_to_meta code_loc)
```
tmattio, line 132: I don't think that's right, can we assume that every ocamldoc code block is in ocaml?

3Rafal: I assume that we default to ocaml when lang header is missing. See: https://ocaml.github.io/odoc/odoc_for_authors.html#ocaml_code_blocks

Also, I've added an explicit test case for this

tmattio:
@rgrinberg this should be ready to review/merge. The support is incomplete, but there shouldn't be any regression compared to the current implementation (and we're actually fixing a couple of issues as demonstrated in the tests).

ocaml-lsp-server/test/e2e/__tests__/textDocument-signatureHelp.ts
```diff
@@ -353,57 +363,75 @@ describe_opt("textDocument/completion", () => {
...
+              \`\`\`verb
                   This text will be displayed verbatim.
                   No formatting will be applied.
+              \`\`\`
               
               #### Module List
               
-              * * *
-              ***@param*** \`x\` dividend
+              *Array
```
tmattio, line 405: Module list are still broken

tmattio:
Spoke too soon, the lists are still broken. Should be ready to review though.

rgrinberg:
@Khady can you take a look? Since you're the original author of this feature.

3Rafal:
@tmattio , lists are already working. I've created an issue in cmarkit, so that it works by default:
https://github.com/dbuenzli/cmarkit/issues/9

ocaml-lsp-server/test/e2e/__tests__/textDocument-signatureHelp.ts
3Rafal: Wow, this is more involved. It didn't work previously so we can fix it later.

3Rafal: I think I misunderstood the feature. The list is fixed now :)

Khady:
actually I think that @jorisgio is the original author, but I'll try to have a look at it too

PR #1089, opened 2023-05-08
[WIP] Upgrade Omd to 2.0.0~alpha3

tmattio:
Alternative to #1088 which upgrades to Omd 2.0.0~alpha3. Unfortunately, Omd 2 doesn't support printing to Markdown yet, and the upgrade can't be completed until it does.

cc @tatchi @ddickstein


tmattio:
We'll go with #1088

PR #1097, opened 2023-05-16
fix(ocamllsp): workspace symbols fallback to _build folder

tatchi:
It checks that the `uri` exist and otherwise fallback to the `_build` folder. This might be the case for generated files (that only exist in the `_build` folder.

Before:

https://github.com/ocaml/ocaml-lsp/assets/5595092/10f65cf6-b048-4058-a0d1-e95ea4c0f136

After:

https://github.com/ocaml/ocaml-lsp/assets/5595092/4e23f045-b25b-400c-91cd-302de3e8e7d1

### DIFF
--- a/ocaml-lsp-server/src/workspace_symbol.ml
+++ b/ocaml-lsp-server/src/workspace_symbol.ml
@@ -205,8 +205,8 @@ end
 
 exception Cancelled
 
-let symbols_from_cm_file ~filter root_uri (cancel : Fiber.Cancel.t option)
-    cm_file =
+let symbols_from_cm_file ~build_dir ~filter root_uri
+    (cancel : Fiber.Cancel.t option) cm_file =
   let cmt =
     let filename = string_of_cm cm_file in
     let cancelled =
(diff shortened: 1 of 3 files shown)


ocaml-lsp-server/src/workspace_symbol.ml
```diff
@@ -230,7 +230,14 @@ let symbols_from_cm_file ~filter root_uri (cancel : Fiber.Cancel.t option)
         in
         let loc = Mbrowse.node_loc browse in
         let fname = loc.loc_start.pos_fname in
-        let uri = Uri.of_path (Filename.concat root_uri fname) in
+        let uri =
+          let sourcefile = Filename.concat root_uri fname in
+          let path =
+            if Fpath.exists (Fpath.of_string sourcefile) then sourcefile
+            else build_dir ^ "/" ^ fname
```
tatchi, line 237: We should probably also verify that the `uri` in `_build` folder exist? And if not remove the entry ?

3Rafal, line 237: I think it would be better if we use Fpath for path concatenation, instead of doing it by hand

3Rafal:
Pls add a changelog note

rgrinberg:
Superseded by #1742.

Issue #1082, opened 2023-05-03
MLI interaction features


EmileTrotignon:
I have two related features I believe would be a great addition.

The first and simplest one is to display in the .ml which values and type are exposed in the .mli.
A way to that would to add a `[pub]` before the type of the value : 
![image](https://user-images.githubusercontent.com/27727887/235867856-14b290d5-0c41-43bb-ba3b-53d8d3f0af3e.png)

The second one would be  to have a code action to expose or make a value private from the ml. It would be even better if it could detect a type mismatch from the .ml and the .mli (maybe in the case the error could be on the `[pub]` note), and offer to promote the type present in the .ml.

rgrinberg:
Both suggestions sound reasonable to me.

PR #1103, opened 2023-05-24
support settings through `didChangeConfiguration` notification

### LINKED ISSUE #1102: Add support to disable settings through `didChangeConfiguration`
Some settings can be disabled by the user. Currently the server has no means of disabling these features at the server level.

VSCode can disable codelens through the following setting:

```json
"[ocaml]": {
  "editor.codeLens": false
}
```

However other clients may not have means of disabling settings, such as Neovim.

### Suggested settings for the user to control

- Codelens
- Hover Extended

### Example config schema

```ocaml
"ocaml.codelens.enable": false,
"ocaml.hover.extended.enable": true
```

aspeddro:
This PR add support to enable/disable settings through `didChangeConfiguration` notification.

## Settings

- CodeLens
- Extended Hover

## Example using Neovim

```lua
require('lspconfig').ocamllsp.setup {
  cmd = {'ocamllsp'},
  on_attach = on_attach,
  settings = {
    extendedHover = { 
      enable = true 
    },
    codelens = {
      enable = false
    }
  },
  capabilities = capabilities,
}
```

Close #1102

### DIFF
--- a/ocaml-lsp-server/docs/ocamllsp/config.md
+++ b/ocaml-lsp-server/docs/ocamllsp/config.md
@@ -1,0 +1,21 @@
+# Configuration
+
+The ocamllsp support the folowing configuration. These configurations are sent through the [`didChangeConfiguration`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeConfiguration) notification.
+
+```ts
+interface config {
+  /**
+  * Enable Extended Hover
+  * @default false
+  * @since 1.16
+  */
+  extendedHover: { enable : boolean }
+
+  /**
+  * Disable CodeLens
+  * @default true
+  * @since 1.16
+  */
+  codelens: { enable : boolean }
+}
+```
--- a/ocaml-lsp-server/docs/ocamllsp/config.md
+++ b/ocaml-lsp-server/docs/ocamllsp/config.md
@@ -5,14 +5,14 @@ The ocamllsp support the folowing configuration. These configurations are sent t
 ```ts
 interface config {
   /**
-  * Enable Extended Hover
+  * Enable/Disabe Extended Hover
   * @default false
   * @since 1.16
   */
   extendedHover: { enable : boolean }
 
   /**
-  * Disable CodeLens
+  * Enable/Disable CodeLens
   * @default true
   * @since 1.16
   */
--- a/ocaml-lsp-server/src/config_data.ml
+++ b/ocaml-lsp-server/src/config_data.ml
@@ -2,7 +2,7 @@ open Import
 open Import.Json.Conv
 
 module Lens = struct
-  type t = { enable : bool Json.Nullable_option.t [@default Some true] }
+  type t = { enable : bool [@default true] }
   [@@deriving_inline yojson] [@@yojson.allow_extra_fields]
 
   let _ = fun (_ : t) -> ()
@@ -20,9 +20,7 @@ module Lens = struct
            | "enable" -> (
              match Ppx_yojson_conv_lib.( ! ) enable_field with
              | Ppx_yojson_conv_lib.Option.None ->
-               let fvalue =
-                 Json.Nullable_option.t_of_yojson bool_of_yojson _field_yojson
-               in
+               let fvalue = bool_of_yojson _field_yojson in
                enable_field := Ppx_yojson_conv_lib.Option.Some fvalue
              | Ppx_yojson_conv_lib.Option.Some _ ->
                duplicates := field_name :: Ppx_yojson_conv_lib.( ! ) duplicates)
@@ -48,7 +46,7 @@ module Lens = struct
            let enable_value = Ppx_yojson_conv_lib.( ! ) enable_field in
            { enable =
                (match enable_value with
-               | Ppx_yojson_conv_lib.Option.None -> Some true
+               | Ppx_yojson_conv_lib.Option.None -> true
                | Ppx_yojson_conv_lib.Option.Some v -> v)
            }))
      | _ as yojson ->
@@ -64,7 +62,7 @@ module Lens = struct
      | { enable = v_enable } ->
        let bnds : (string * Ppx_yojson_conv_lib.Yojson.Safe.t) list = [] in
        let bnds =
-         let arg = Json.Nullable_option.yojson_of_t yojson_of_bool v_enable in
+         let arg = yojson_of_bool v_enable in
          ("enable", arg) :: bnds
        in
        `Assoc bnds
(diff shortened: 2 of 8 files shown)


ocaml-lsp-server/src/config_data.ml
```diff
@@ -0,0 +1,261 @@
+open Import
+open Import.Json.Conv
+
+module Lens = struct
```
3Rafal, line 4: It would be nice to rename this module to `Codelens`, so that it's easier to figure out it's purpose and consistent with field name.

ocaml-lsp-server/src/config_data.ml
```diff
@@ -0,0 +1,261 @@
...
+  type t = { enable : bool Json.Nullable_option.t [@default Some true] }
```
3Rafal, line 5: Can we make `enable` to have type of bool, with default value of true?
That way we would have only two possibilities: true or false. Now, we have 3 (Some true, Some false, None), but two of them are equivalent.

ocaml-lsp-server/docs/ocamllsp/config.md
```diff
@@ -0,0 +1,21 @@
...
+The ocamllsp support the folowing configuration. These configurations are sent through the [`didChangeConfiguration`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeConfiguration) notification.
+
+```ts
+interface config {
+  /**
+  * Enable Extended Hover
+  * @default false
+  * @since 1.16
+  */
+  extendedHover: { enable : boolean }
```
3Rafal, line 12: would it be simpler to use boolean directly, instead of an object with `enable` key?
It would be similar to config that is used in VSCode:
```
"[ocaml]": {
  "editor.codeLens": false
}
```

ocaml-lsp-server/src/config_data.ml
```diff
@@ -0,0 +1,261 @@
...
+      : t -> Ppx_yojson_conv_lib.Yojson.Safe.t)
+
+  let _ = yojson_of_t
+
+  [@@@end]
+end
+
+type t =
+  { codelens : Lens.t Json.Nullable_option.t
+        [@default None] [@yojson_drop_default ( = )]
```
3Rafal, line 154: I think it would be simpler to have defaults of type bool (true/false), instead of option.

ocaml-lsp-server/src/configuration.ml
```diff
@@ -14,28 +17,38 @@ let default () =
     in
     Lev_fiber.Timer.Wheel.create ~delay
   in
-  { wheel }
+  let data =
+    Config_data.
+      { codelens = Some { enable = Some true }
```
3Rafal, line 22: Moving creation of defaults to `Config_data` module would make the structure better. That way, we would have them closer to type definition. Now we are inlining the defaults, which is harder to read. I think it's better to call here `Config_data.default` instead

3Rafal:
Please add a changelog note

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -598,12 +598,18 @@ let on_request :
...
+      | Some _ | None -> Hover_req.Default
+    in
+    later (fun (_ : State.t) () -> Hover_req.handle rpc req mode) ()
   | TextDocumentReferences req -> later references req
   | TextDocumentCodeLensResolve codeLens -> now codeLens
-  | TextDocumentCodeLens req -> later text_document_lens req
+  | TextDocumentCodeLens req -> (
+    match state.configuration.data.codelens with
+    | Some { enable = Some true } -> later text_document_lens req
+    | None | Some _ -> now [])
```
3Rafal, line 612: Does None here mean, that if we update extended_hover without codelens, they will get disabled?
This logic is a bit tricky, but it will be resolved if we use boolean defaults directly.

3Rafal:
We have tests for:
[hover](https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/test/e2e/__tests__/textDocument-hover.test.ts)
[hover extended](https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/test/e2e-new/hover_extended.ml)
[code lens](https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/test/e2e/__tests__/textDocument-codeLens.test.ts)

Can we add tests that change default configuration and check if it is applied?

3Rafal (commented):
Thanks @aspeddro for this PR. It looks good, but I think it would be nice to add suggested improvements. Feel free to ask me any questions in case something is not clear. :)

ocaml-lsp-server/docs/ocamllsp/config.md
aspeddro: If in the future more options related to codelens were added then it would be inside the `codelens` object.

An example: adding codelens to show public values and types https://github.com/ocaml/ocaml-lsp/issues/1082

```json
{
  "codelens": {
    "enabled": true,
    "displayPublicValuesTypes": true
  }
}
```

This is just a pattern I've seen on some servers.

ocaml-lsp-server/src/ocaml_lsp_server.ml
aspeddro: My mistake. `None` should be in the first case match case.

By default `codelens` is enabled.

aspeddro: Done 175d5e64a0df1d98ad4b31a3a7b5509973d8b554

ocaml-lsp-server/src/configuration.ml
aspeddro: Done 175d5e64a0df1d98ad4b31a3a7b5509973d8b554

ocaml-lsp-server/docs/ocamllsp/config.md
```diff
@@ -0,0 +1,21 @@
...
+
+  /**
+  * Disable CodeLens
```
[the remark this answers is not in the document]
aspeddro: Done c0dac92240a132e148e04b2264f043e8f9e4a9ea

ocaml-lsp-server/docs/ocamllsp/config.md
```diff
@@ -0,0 +1,21 @@
+# Configuration
```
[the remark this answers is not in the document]
aspeddro: d4111ab41a27dfcdac644b13b46647c4e20bf17e

aspeddro:
Related to this PR:

We can remove the `ocamllsp/hoverExtended` request introduced in #561. This is now supported by most clients and is within specification.

PR #1110, opened 2023-05-29
chore: Rewrite tests that fail because of Windows line endings

3Rafal:
Old tests failed on Windows, when their line endings had newlines. I've rewritten them to caml, to make CI green.
This is a minimum amount of work that passes. I will try to rewrite the rest of them in future (separate PRs).


3Rafal:
@rgrinberg , could you please add a label `no changelog`, so that it can be used to make changelog check pass?

PR #1113, opened 2023-05-31
Remove the `--fallback-read-dot-merlin` flag

anmonteiro:
Following the release of [Melange 1.0](https://github.com/ocaml/opam-repository/pull/23850), this flag is no longer needed in OCaml LSP.

This reverts commit c2751403ad55acef079a4ddfd2182371777c2749.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Unreleased
 
+## Breaking Change
+
+- Remove the `--fallback-read-dot-merlin` flag to `ocamllsp`, dropping support
+  for reading `.merlin`. (#1113)
+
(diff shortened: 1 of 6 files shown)


sidkshatriya:
Is `--fallback-read-dot-merlin` useful for other projects that have not moved to dune yet but still want to work with merlin?

ddickstein:
`.merlin` files are still used at Jane Street, so we still need this support.

CatherineGasnier:
Meta also need that option since we use buck instead of dune.

voodoos:
Closing this as the flag is still useful.

PR #1118, opened 2023-06-05
Vendoring Omd 1.3.2

cuihtlauac:
This removes the dependency on omd and adds the corresponding
sources without tests.

Requiring to have omd strictly before 2.0.0~alpha1 is a foot gun
for newbies. ocaml-lsp-server is among the first packages newbies
have to install. In that process, old omd 1.3.2 gets installed.
Later, when starting to code with Omd, they either don't understand
why their installed API doesn't match the publicly documented one
or start coding with an obsolete API.


rgrinberg:
Seems like we're moving away from omd here: https://github.com/ocaml/ocaml-lsp/pull/1088

Otherwise, I agree with your assessment.

cuihtlauac:
Sorry for the noise. I hadn't noticed @tmattio PR.

Issue #700, opened 2022-05-17
Code actions: handle "unused ..." warnings


ulugbekna:
We can add code actions to handle various compiler warnings for unused module/value/etc. 

- [ ] Warning 11: Redundant case in a pattern matching (unused match case).
- [ ] Warning 20: Unused function argument
- [ ] Warning 32: Unused value declaration
- [ ] Warning 33: Unused open 
  - [ ] remove unused open 
  - [ ] convert it to `open!`
- [ ] Warning 34: Unused type declaration
- [ ] Warning 35: Unused for-loop index
- [ ] Warning 36: Unused ancestor
- [ ] Warning 37: Unused constructor
- [ ] Warning 38: Unused extension constructor
- [ ] Warning 39: Unused rec flag
- [ ] Warning 60: Unused module
- [ ] Warning 66: Unused open! bang
- [ ] Warning 67: Unused functor parameter
- [ ] Warning 69: Unused record field 
- [ ] Warning 71: Unused @tail_mod_cons attribute

PR #1141, opened 2023-06-16
Extend mark/remove unused actions

jfeser:
Extends the mark/remove unused actions to handle unused diagnostics from dune.

Mark:

- [x] Open (convert to open!)
- [x] Type
- [x] For loop index

Remove:

- [x] Open
- [x] Open!
- [x] Type
- [x] Module
- [x] Case
- [x] Rec
- [x] Constructor
- [ ] Ext. constructor 

See #700


ocaml-lsp-server/src/code_actions/action_mark_remove_unused.ml
```diff
@@ -127,27 +172,226 @@ let code_action_remove_value doc pos (diagnostic : Diagnostic.t) =
...
+  match Position.compare_inclusion pos (Range.of_loc loc) with
+  | `Outside _ -> false
+  | `Inside -> true
+
+let action_mark_for_loop_index doc pos (d : Diagnostic.t) =
+  let open Option.O in
+  let module I = Ocaml_typing.Tast_iterator in
+  let exception Found of Warnings.loc in
+  let iterator =
+    let expr iter (e : Typedtree.expression) =
```
rgrinberg, line 216: I might be wrong here, but can't you iterate over the parsetree instead?

jfeser: Yes, and I'm not sure why I didn't. I'll change that.

PR #1145, opened 2023-06-18
Remove extended hover request `ocamllsp/hoverExtended`

aspeddro:
Since #1103 the server can provide more information on hover request `textDocument/hover`.

Only the `verbosity` field was not implemented. This PR add `verbosity` config to `extendedHover` and remove the custom lsp request endpoint `ocamllsp/hoverExtended`.

Note: `verbosity` is optional. Default is 0.


rgrinberg:
We can't really remove existing requests just like that. That would be a breaking change which I would argue is unnecessary.

Also, IIUC, this PR argues that to replicate the old behavior it's enough to just set the config appropriately. While that's possible, I think this API is brittle. In particular, it doesn't work well when there's requests being made automatically by the client and they all rely on different settings. It's also inferior when you consider users automating their editor using LSP requests. Now they would need to remember to set the config and then restore it.

PR #1150, opened 2023-06-19
Compatibility with OCaml 5.1

voodoos:
This PR aims to make `master` compatible with OCaml 4.14, 5.0 and 5.1.
Not sure if there will be differences in tests however.

See https://github.com/ocaml/opam-repository/pull/23950 for more context.


voodoos:
Is there no `ocaml-ci` for ocaml-lsp ? It is useful to test on all supported ocaml versions.

rgrinberg:
> edit: I added 5.0 and 5.1 to the matrix, but that makes for a lot of jobs, we might want to restrain some versions of ocaml to only one system / and build-only.

How about we only test 5.0 on Linux

voodoos:
So in the end:
- Tests are run only against the 4.14 version (output is different between versions due to changes in error messages)
- Build for 5.0 and 5.1 is performed only on ubuntu and works
- Windows is broken, but that seems independent to these changes...

smorimoto:
I encountered a crash loop when using lsp with 5.1, any ideas?

voodoos:
Do you have any log or reproduction ? We need more information to start investigate.

PR #1153, opened 2023-06-22
fix(ocamllsp): fix in pp handling the file permissions

jboillot:
Hello,
In the PR that made Merlin a dependency #1070 a bug was introduced that prevents from using preprocessor programs as https://github.com/ocaml-community/cppo.
This is because the file permissions are not provided in the right base and that the output files are currently not created if they do not exist yet.
I do not know if the addition of the flag `Unix.O_CREAT` could cause an issue with some pp or ppx.

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -878,12 +878,13 @@ let run_in_directory ~prog ~prog_is_quoted:_ ~args ~cwd ?stdin ?stdout ?stderr
   let argv = [ "sh"; "-c"; cmd ] in
   let stdin =
     match stdin with
-    | Some file -> Unix.openfile file [ Unix.O_WRONLY ] 0x664
-    | None -> Unix.openfile "/dev/null" [ Unix.O_RDONLY ] 0x777
+    | Some file -> Unix.openfile file [ Unix.O_WRONLY ] 0o664
+    | None -> Unix.openfile "/dev/null" [ Unix.O_RDONLY ] 0o777
   in
   let stdout, should_close_stdout =
     match stdout with
-    | Some file -> (Unix.openfile file [ Unix.O_WRONLY ] 0x664, true)
+    | Some file ->
+      (Unix.openfile file [ Unix.O_WRONLY; Unix.O_CREAT ] 0o664, true)
     | None ->
       (* Runned programs should never output to stdout since it is the channel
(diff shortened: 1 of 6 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -878,20 +878,21 @@ let run_in_directory ~prog ~prog_is_quoted:_ ~args ~cwd ?stdin ?stdout ?stderr
   let argv = [ "sh"; "-c"; cmd ] in
   let stdin =
     match stdin with
-    | Some file -> Unix.openfile file [ Unix.O_WRONLY ] 0x664
-    | None -> Unix.openfile "/dev/null" [ Unix.O_RDONLY ] 0x777
+    | Some file -> Unix.openfile file [ Unix.O_WRONLY ] 0o664
```
jboillot, line 881: Should not the input file be opened in read-only instead of write-only?

voodoos:
I think this needs to be fixed on Merlin's side: in the same way as for `-ppx` the `-pp` flag expect an arbitrary command, with it's arguments, so we should really handle this the same as ppxs (but we changed that behavior when refactoring).

I will do that asap.

voodoos:
However the proposed change looks reasonable: having input files be read-only and allowing creating new files.

voodoos (approved):
After having a look on Merlin's side I think things are all-right there. I added a test for preprocessing.

The issue doesn't show with `ppxes` because the driver is responsible for creating/writing to the file, it is not relying on redirecting `stdout`. 

The changes look good to me. 

@jboillot where you able to test the changes on your project ? 

It can be done using `opam pin https://github.com/jboillot/ocaml-lsp.git#bugfix/ocamllsp-pp-output-file-creation`.

@rgrinberg I guess this deserves a patch release. I can take care of it if you want (and if I have the rights).

rgrinberg:
Yes, give it a try.

Where's the test though?

voodoos:
The test I mentioned is on Merlin side. I am looking at how I could do the same in `ocaml-lsp` but it's not as simple without cram and the cli interface.

https://github.com/ocaml/merlin/pull/1630/files

jboillot:
@voodoos I was able to test the changes on a few files of a big OCaml project with cppo as the only pp (and a few other ppxes).

voodoos:
I added a test that was effectively failing without the changes in this PR.

voodoos:
Thanks a lot @jboillot !

PR #1156, opened 2023-06-27
Batch merlin work in code actions

jfeser:


ocaml-lsp-server/src/code_actions/action_destruct.ml
```diff
@@ -81,4 +81,4 @@ let code_action (state : State.t) doc (params : CodeActionParams.t) =
         } -> Fiber.return None
     | Error exn -> Exn_with_backtrace.reraise exn)
 
-let t state = { Code_action.kind; run = code_action state }
+let t state = { Code_action.kind; run = `Non_batchable (code_action state) }
```
rgrinberg, line 84: What makes destruct non-batchable? The fact that we try to format the result?

rgrinberg: Sounds fine, although better would be to just convert the actions into two part: one that uses fibers and the one that doesn't.

rgrinberg (approved):

PR #1166, opened 2023-07-18
Display text of references in doc strings

AurelienSaue:
This PR does not add support for cross-references, but Doc_to_md now outputs the text of the reference, instead of a line break as it does currently.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Features
+
+- Display text of references in doc strings (#1166)
+
 # 1.16.2
 
 ## Fixes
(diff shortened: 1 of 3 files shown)


tmattio:
Thanks @AurelienSaue! Could you add a test that showcases the improvement? https://github.com/ocaml/ocaml-lsp/pull/1088 should show how to do this.

tmattio (approved):
Looks good to me if CI is happy!

AurelienSaue:
I am confused by the error, I am not sure it comes from my addition.

tmattio:
It's not, we've had recurring problems with the Windows CI.

@rgrinberg or @voodoos I'll let you merge when you think this is ready, I was chiming in since I wrote a good part of the new markdown conversation, but I'll leave the merge to you.

PR #1170, opened 2023-07-28
retain super/subscripts in markdown conversion

### LINKED ISSUE #1168: Special characters in doc get eaten by LSP
Here is the documentation for `Random.int`, gotten by hovering in vscode:
image
And here is the doc on ocaml.org:
image
Notice the difference?
Turns out it's `2{^30}` in the docstring, which is a whole different ballpark from 230.
[Reference for the docstring](https://github.com/ocaml/ocaml/blob/5.0/stdlib/random.mli#L52-L54)
This can get real confusing if the subject is more complex than this.

jfeser:
Fixes  #1168

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Fixes
+
+- Fix missing super & subscripts in markdown documentation. (#1170)
+
 # 1.16.2
 
 ## Fixes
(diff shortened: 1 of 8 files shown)


rgrinberg (approved):

ElectreAAS:
Thank you!

PR #1173, opened 2023-08-19
Do not invoke dune at all if `--fallback-read-dot-merlin` flag is on

### LINKED ISSUE #859: .merlin based projects still require a blank dune-project file
Thanks for adding the `--fallback-read-dot-merlin` option in #705!

It seemed to do nothing though, until I've looked at the code and realized why:

- [config calls get_external_config](https://github.com/ocaml/ocaml-lsp/blob/2c31f3585e2579aa0dcd5bdb9e8934760bdd859d/ocaml-lsp-server/src/merlin_config.ml#L391-L397) *and if that finds nothing, uses an initial blank config, i.e. doesn't get to reading .merlin*
- [get_external_config calls find_project_context](https://github.com/ocaml/ocaml-lsp/blob/6cbde1bded8afedef1b645ba4d77eba333d9cdc6/ocaml-lsp-server/vendor/merlin/src/kernel/mconfig.ml#L206-L209)
- [find_project_context looks for a parent directory with `dune-project` or `dune-workspace`](https://github.com/ocaml/ocaml-lsp/blob/2c31f3585e2579aa0dcd5bdb9e8934760bdd859d/ocaml-lsp-server/src/merlin_config.ml#L339)!

So currently it is necessary to `touch dune-project` to make a legacy `.merlin` non-dune project work. It would be nice if that wasn't the case.

SamChou19815:
Fix #859.

For background, internally at Meta we use [buck2](https://buck2.build/) to build ocaml code, and we have a script to automatically generate a .merlin file based on the build artifacts. I'm not asking for a buck2 integration here. I'm simply fix the issue that blocks us from switching away from the long dead `ocaml-language-server`.

I find that the flag doesn't work if the environment does not have dune installed, because it will unconditionally run some dune command before even checking the `--fallback-read-dot-merlin` flag. This PR moves the check earlier, so that when the flag is on, it fully bypasses the dune workflow.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ## Fixes
 
 - Fix missing super & subscripts in markdown documentation. (#1170)
+- Do not invoke dune at all if `--fallback-read-dot-merlin` flag is on. (#1173)
 
 ## Features
 
--- a/ocaml-lsp-server/src/merlin_config.ml
+++ b/ocaml-lsp-server/src/merlin_config.ml
@@ -418,28 +418,30 @@ let config (t : t) : Mconfig.t Fiber.t =
     t.entry <- Some entry
   in
   let* () = Fiber.return () in
-  match find_project_context t.directory with
-  | None ->
-    let+ () = destroy t in
-    t.initial
-  | Some (ctx, config_path) ->
-    let* entry = get_process t.db ~dir:ctx.process_dir in
-    let* () =
-      match t.entry with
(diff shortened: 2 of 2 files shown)


gridbugs:
Thanks for the fix! I'm not an ocaml-lsp maintainer but I came across this fix because I was just bitten by https://github.com/ocaml/ocaml-lsp/issues/859 myself and this patch solves my problem.

PR #1177, opened 2023-09-12
Preview ocamlformat 0.26.1

Julow:
The aim of this preview is to gather feedback.

This is a bug-fix release and is compatible with the slightly changed let-binding syntax in OCaml 5.1.
Changelog can be found here: https://github.com/ocaml-ppx/ocamlformat/blob/main/CHANGES.md

The upgrade to OCamlformat 0.26.0 is made in a separate commit to easily check the regression. I can squash the commits if needed.


ocaml-lsp-server/src/semantic_highlighting.ml
jfeser, line 1: I wouldn't break the line after fun, but the change doesn't look any worse.

Julow:
OCamlformat has been released a week ago, this is ready to be merged. If possible not squashed for the `.git-blame-ignore-revs`.

PR #1181, opened 2023-09-17
Fix edge cases that prevent completion

faldor20:
I encountered a few cases where completion doesn't work correctly. Most obviously:
```ocaml
let a=[1]|>Li
```
This will not complete properly.
This was because the position_prefix code had a bunch of edge cases it didn't account for. I have added a whole collection of tests for these edge cases, and included two different proposed replacements for this prefix code, plus a benchmark of the difference between the old and new solutions.

I'm quite new to ocaml overall so i wanted to just run this by you all to get some feedback before i clean it up and settle on a solution.

After mucking around with this for many hours I think that the best solution would be to actually run the lexer over the tiny snippet and then use that token information to make an informed decision about what to feed back to merlin.

One key thing that occured to me is that currently none of these solution covers very odd edge cases like:
```ocaml
[1]|>List.
ma
[1]|> List. ma
```
but use lexer would solve this problem. though I'm honestly not sure if merlin would even deal with such a thing.


faldor20:
```
┌─────────────────┬────────────┬─────────┬────────────┐
│ Name            │   Time/Run │ mWd/Run │ Percentage │
├─────────────────┼────────────┼─────────┼────────────┤
│ regex           │   916.98ns │  90.00w │     44.62% │
│ regex_long      │ 1_168.30ns │  97.00w │     56.85% │
│ regex_only      │   392.91ns │  51.00w │     19.12% │
│ regex_only_long │ 2_055.10ns │  50.99w │    100.00% │
│ old             │   269.50ns │  22.00w │     13.11% │
│ old_long        │   260.33ns │  22.00w │     12.67% │
└─────────────────┴────────────┴─────────┴────────────┘
```
these are the results of running it on my laptop. 
This could get a bit if we are happy to reverse the list still.  but it does make the regex pretty weird, because it's backwards

rgrinberg:
I'm curious for why you added the benchmarks, anything in particular you expect to be slow here?

ocaml-lsp-server/src/compl.ml
```diff
@@ -95,8 +103,51 @@ let prefix_of_position ~short_path source position =
...
     else reconstructed_prefix
 
-(** [suffix_of_position source position] computes the suffix of the identifier
-    after [position]. *)
+module String = struct
+  include String
+
+  (**Filters a string keeping any chars for which f returns true and discarding
+     those for which it returns false*)
+  let filter str ~f =
```
rgrinberg, line 111: General function should be moved to import.ml

ocaml-lsp-server/src/compl.ml
```diff
@@ -260,8 +311,12 @@ module Complete_with_construct = struct
       List.mapi constructed_exprs ~f:completionItem_of_constructed_expr
 end
 
+let logCompletion log =
+  Log.log ~section:"resolveCompletion" (fun () -> Log.msg log [])
+
 let complete (state : State.t)
     ({ textDocument = { uri }; position = pos; _ } : CompletionParams.t) =
+  logCompletion "ho1";
```
rgrinberg, line 319: junk?

ocaml-lsp-server/src/compl.ml
```diff
@@ -365,10 +426,12 @@ let format_doc ~markdown doc =
 
 let resolve doc (compl : CompletionItem.t) (resolve : Resolve.t) query_doc
     ~markdown =
+  logCompletion "Starting  completion";
```
rgrinberg, line 429: Why the extra space?

ocaml-lsp-server/src/compl.ml
```diff
@@ -380,6 +443,7 @@ let resolve doc (compl : CompletionItem.t) (resolve : Resolve.t) query_doc
                 (Document.Merlin.source doc)
                 logical_position
             in
+            logCompletion @@ "completion prefix is:" ^ prefix;
```
rgrinberg, line 446: This log message is allocating its argument if no logging is enabled. Defeating the purpose of having `Log.log` taking a function argument.

ocaml-lsp-server/src/compl.ml
```diff
@@ -392,7 +456,9 @@ let resolve doc (compl : CompletionItem.t) (resolve : Resolve.t) query_doc
               character = position.character + String.length suffix
             }
           in
+
           let range = Range.create ~start ~end_ in
+
```
rgrinberg, line 461: pointless space

ocaml-lsp-server/src/compl.mli
```diff
@@ -27,7 +27,9 @@ val resolve :
   -> CompletionItem.t Fiber.t
 
 (** [prefix_of_position ~short_path source position] computes prefix before
-    given [position].
+    given [position]. 
+    A prefix is essentially a piece of code that refers to one thing eg a single infix operator "|>", a single reference to a function or variable: "List.map" a keyword "let" etc
```
rgrinberg, line 31: 80 chars wrapping

ocaml-lsp-server/bench/ocaml_lsp_bench.ml
```diff
@@ -0,0 +1,44 @@
+open Core
+open Core_bench
+open Ocaml_lsp_server.Testing
```
rgrinberg, line 3: Don't open this module and just qualify everything you need as `Testing.` please.

ocaml-lsp-server/src/prefix_parser.ml
```diff
@@ -0,0 +1,31 @@
+open Re
+
+(*Regex based parser*)
+
+let name_or_label_regex =
+  Re.compile @@ Re.Posix.re {|([~?`]|let%|and%)?([a-zA-Z0-9_']|[a-zA-Z0-9_'] *\. *)+$|}
+
+(** matches let%lwt and let* style expressions. See here:https://v2.ocaml.org/manual/bindingops.html *)
+let monadic_bind =
+  Re.compile @@ Re.Posix.re {|(let|and)([$&*+\-/=>@^|.]|(%[a-zA-Z0-9_']*))$|}
```
rgrinberg, line 10: Can you use the combinators from `Re` instead of `Re.Posix` please.

faldor20:
@rgrinberg Well, know not particularly. But for code that runs almost every keypress I think speed is key. I wanted to validate that I wasn't causing any meaningful slowdown

ocaml-lsp-server/src/compl.ml
faldor20: Yeah, good point

ocaml-lsp-server/src/compl.ml
faldor20: I thought ocamlfmt would have dealt with those, oh well done now :)

faldor20: I thought ocamlfmt would have dealt with those, oh well done now :)

ocaml-lsp-server/src/prefix_parser.ml
faldor20: How set are you on this? I far far prefer to use standard regex because 
a) it's something universal
b) it means it's easy to test and debug using tools like https://regexr.com/ 
But if you think there are good reasons to use the combinators I can convert it

ocaml-lsp-server/src/prefix_parser.ml
```diff
@@ -0,0 +1,31 @@
...
+
+let infixRegex = Re.compile @@ Re.Posix.re {|[~?:!$&*+\-\/=><@^|%<.#]+$|}
+
+open Import
+
+module Option = struct
+  include Option
+
+  let none_bind func option =
```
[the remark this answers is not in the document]
faldor20: do you mean make multiple copies of that some logic? or just to use the full match logic in every case?

ocaml-lsp-server/test/e2e/__tests__/textDocument-completion.test.ts
```diff
@@ -91,6 +91,395 @@ describe_opt("textDocument/completion", () => {
     expect(items).toMatchObject([{ label: "StringLabels" }]);
   });
 
+  it("can start completion after operator with space", async () => {
```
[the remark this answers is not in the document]
faldor20: I didn't really want to migrate the whole test suite in this PR and i didn't see any existing completion tests.

Would it be okay to just merge this with the new tests in the old suite and then bundle them in with a larger effort to convert all the completion tests to the new e2e suite?

faldor20:
Also I've switched to the reversed implementation because its performance is more predictable and writing some regexes backwards isn't that onerous

faldor20:
As for the benchmarks, I understand they may not make sense in this PR and you may want a different benchmarking lib etc. But I do think it makes sense to have something setup at some point. Even just for testing little things to ensure that new changes aren't causing perf regressions. I'm sure some people out there can always accurately guess the performance effect of a change just by looking at it, but I think microbenchmarks are a very good thing to have around.

An example of this is that i find this server mostly very performant(kudos, the startup time is super impressive) but every now and again it gets into a weird state or i open a particular file that causes major slowdown. I'd love to find out why

ocaml-lsp-server/src/prefix_parser.ml
rgrinberg, line 1: This module needs an mli

ocaml-lsp-server/src/prefix_parser.ml
rgrinberg: I think the main issue with the way you've written it is that you're using capturing groups that you don't actually need. That's a little confusing and also slows down the matching.

ocaml-lsp-server/src/prefix_parser.ml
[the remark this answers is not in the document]
rgrinberg: The explicit match would be better. I usually don't like introducing helper functions that nobody is familiar with. Put another way, my criteria is that it should be present in at least one popular stdlib before introducing it here.

You can also rewrite your code in the following way if I'm not mistaken:

```
List.find_map [ name_or_label_regex_rev ; monadic_bind_rev ; infixRegex_rev ] ~f:(fun re ->
  Re.exec_opt re text)
```

ocaml-lsp-server/test/e2e/__tests__/textDocument-completion.test.ts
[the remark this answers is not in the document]
rgrinberg: If you don't have time, I can do it myself. I'm afraid that the larger effort to convert everything to the new test suite is never going to come. So for now, we're just not going to let the current test suite grow.

ocaml-lsp-server/src/prefix_parser.ml
```diff
@@ -0,0 +1,36 @@
...
+let name_or_label_regex_rev =
+  Re.compile @@ Re.Posix.re {|^([a-zA-Z0-9_']| *\. *[a-zA-Z0-9_'])+([~?`]|%tel|%dna)?|}
+
+(** matches let%lwt and let* style expressions. See here:https://v2.ocaml.org/manual/bindingops.html *)
+let monadic_bind =
+  Re.compile @@ Re.Posix.re {|(let|and)([$&*+\-/=>@^|.]|(%[a-zA-Z0-9_']*))$|}
+let monadic_bind_rev =
+  Re.compile @@ Re.Posix.re {|^([$&*+\-/=>@^|.]|([a-zA-Z0-9_']*)%)(tel|dna)|}
+
+let infixRegex = Re.compile @@ Re.Posix.re {|[~?:!$&*+\-\/=><@^|%<.#]+$|}
```
rgrinberg, line 17: `infix_regex`

rgrinberg: Also, the `regex` suffix isn't particularly useful. We already know what it is from the type

ocaml-lsp-server/bench/dune
```diff
@@ -0,0 +1,16 @@
...
+  ocaml_lsp_server
+ core_unix.command_unix
+  merlin-lib.kernel
+
+  base
+  core
+  core_bench
+  )
+
+ (preprocess(pps ppx_jane ppx_bench))
```
rgrinberg, line 15: You should run `make fmt`

ocaml-lsp-server/src/testing.ml
```diff
@@ -0,0 +1,8 @@
+(**WARNING: This is for internal use in testing only *)
+
+module Compl=Compl
+module Document =Document
+module Import =Import
```
rgrinberg, line 5: I don't see this module being used in the tests.

rgrinberg:
> As for the benchmarks, I understand they may not make sense in this PR and you may want a different benchmarking lib etc. But I do think it makes sense to have something setup at some point. Even just for testing little things to ensure that new changes aren't causing perf regressions. I'm sure some people out there can always accurately guess the performance effect of a change just by looking at it, but I think microbenchmarks are a very good thing to have around.

I don't mind the benchmarking library you chose. It's just that there's so much overhead to getting completions that I doubt that running a regex is going to make any difference. Anyway, you can keep them.

> An example of this is that i find this server mostly very performant(kudos, the startup time is super impressive) but every now and again it gets into a weird state or i open a particular file that causes major slowdown. I'd love to find out why

Unfortunately most of the those performance problems are inherited from merlin itself.

ocaml-lsp-server/src/prefix_parser.ml
rgrinberg: Also, I think that if you used the combinators, constructing the "reverse" regex would be much simpler.

faldor20:
Yeah I would have expected that Merlin would be the bottleneck in most cases. But I thought that I may as well ensure I'm not making it worse :). 
I think the recent release of.soemthing like bun in the JS ecosystem really highlights that great performance often comes from "sweating the small stuff" ,or at least paying attention to it.

Like it'd be great if Merlin and a robust benchmark suite that could identify the areas that could be improved. 

If you're happy I'll leave it as is.

ocaml-lsp-server/src/compl.ml
```diff
@@ -26,8 +26,16 @@ let completion_kind kind : CompletionItemKind.t option =
...
+(* I should just rewrite all of this so that it uses a nice for loop. This
+   current soluction is a nice try but overall crap we need to be able to look
+   ahead and behind
+
+   Split it into name and infix name is obvious infix can be either a dot, a
+   label or an I could possibly do a regex based parser.
+
+   Name regex: ((\w)|\w.)*$ *)
+
+let prefix_of_position_old ~short_path source position =
```
jfeser, line 38: Let's remove this function unless it's used elsewhere. We shouldn't keep it just for benchmarks.

ocaml-lsp-server/src/compl.ml
```diff
@@ -95,8 +103,39 @@ let prefix_of_position ~short_path source position =
...
+  | text ->
+    let end_of_prefix =
+      let (`Offset index) = Msource.get_offset source position in
+      min (String.length text - 1) (index - 1)
+    in
+    let prefix_text =
+      let pos =
+        (*clamp the length of a line to process at 500 chars, this is just a
+          reasonable limit for regex performance*)
+        max 0 (end_of_prefix - 500)
```
jfeser, line 119: I think we should avoid limits like this unless required.

faldor20:
Okay, I have no idea why the formatter decided that it would be a good idea to format the result of that expect test, should be fixed now. Sorry this has been a bit of a mess, I'm actually traveling right now, so my work on this has been a bit erratic and distracted.

faldor20:
Okay I actually have no idea why the build failed, I didn't change the doc_to_md file, at all.
I'll try to check more tomorrow when I at my laptop.

faldor20:
Ahh I see, looks like I was hit with the issue fixed here https://github.com/ocaml/ocaml-lsp/pull/1204 could I get a rerun of CI @rgrinberg?

dune-project
```diff
@@ -59,6 +59,7 @@ possible and does not make any assumptions about IO.
   (odoc-parser (and (>= 2.0.0) (< 2.3.0)))
   (ppx_expect (and (>= v0.15.0) :with-test))
   (ocamlformat (and :with-test (= 0.24.1)))
+  (ppx_bench (and :with-test (>= 0.16.0)))
```
rgrinberg, line 62: Why do we need this for the tests?

rgrinberg:
Thanks for the thorough PR and apologies for such a late response. I've merged it in https://github.com/ocaml/ocaml-lsp/pull/1212 with a minor tweaks.

The benchmarks had to be disabled in CI because there's no good way to report their results and they just slow things down with the unnecessary dependencies otherwise.

faldor20:
All good, sorry I didn't quite have the time and expertise to finish that last little bit. Glad it got done :)

PR #1184, opened 2023-09-21
Compatibility with odoc-parser.2.3.0

panglesd:
This PR makes ocaml-lsp compatible with the soon release odoc-parser.2.3.0

odoc-parser adds two new constructs:
- "Outputs" of code blocks. This is meant to be generated by the code blocks and promoted by mdx.
- Tables

Tables are tricky, as odoc tables are allowed to have some block content inside ("nestable" block contents) while markdown is not.
This means that when translating odoc tables into cmarkit tables, we need to turn blocks into inlines. This issue is responsible for most of the code of the PR!

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -56,7 +56,8 @@ possible and does not make any assumptions about IO.
   ordering
   dune-build-info
   spawn
-  (odoc-parser (>= 2.3.0))
+  astring
+  camlp-streams
   (ppx_expect (and (>= v0.15.0) :with-test))
   (ocamlformat (and :with-test (= 0.24.1)))
   (ocamlc-loc (>= 3.7.0))
--- a/dune-project
+++ b/dune-project
@@ -56,7 +56,7 @@ possible and does not make any assumptions about IO.
   ordering
   dune-build-info
   spawn
-  (odoc-parser (and (>= 2.0.0) (< 2.3.0)))
+  (odoc-parser (>= 2.3.0))
   (ppx_expect (and (>= v0.15.0) :with-test))
   (ocamlformat (and :with-test (= 0.24.1)))
   (ocamlc-loc (>= 3.7.0))
--- a/flake.nix
+++ b/flake.nix
@@ -92,7 +92,8 @@
               ppx_yojson_conv_lib
               uutf
               lsp
-              odoc-parser
+              astring
+              camlp-streams
               merlin-lib
             ];
             doCheck = false;
@@ -110,7 +111,8 @@
               duneVersion = "3";
               buildInputs = with pkgs.ocamlPackages; [
                 ocamlc-loc
-                odoc-parser
+                astring
+                camlp-streams
                 dune-build-info
                 re
                 dune-rpc
--- a/ocaml-lsp-server/src/doc_to_md.ml
+++ b/ocaml-lsp-server/src/doc_to_md.ml
@@ -106,6 +106,41 @@ let rec nestable_block_element_to_block
     let paragraph = Block.Paragraph.make inline in
     let meta = loc_to_meta location in
     Block.Paragraph (paragraph, meta)
+  | { value = `Table ((grid, alignment), _); location } ->
+    let meta = loc_to_meta location in
+    let cell
+        ((c, _) : Odoc_parser.Ast.nestable_block_element Odoc_parser.Ast.cell) =
+      let c = nestable_block_element_list_to_inlines c in
+      (c, (" ", " ") (* Initial and trailing blanks *))
+    in
+    let header_row
+        (row : Odoc_parser.Ast.nestable_block_element Odoc_parser.Ast.row) =
+      let row = List.map ~f:cell row in
+      ((`Header row, Meta.none), "")
+    in
+    let data_row
+        (row : Odoc_parser.Ast.nestable_block_element Odoc_parser.Ast.row) =
+      let row = List.map ~f:cell row in
+      ((`Data row, Meta.none), "")
+    in
+    let alignment_row =
+      match alignment with
+      | None -> []
+      | Some alignment ->
+        let alignment =
+          List.map
+            ~f:(fun x -> ((x, 1 (* nb of separator *)), Meta.none))
+            alignment
+        in
+        [ ((`Sep alignment, Meta.none), "") ]
+    in
+    let rows =
+      match grid with
+      | [] -> assert false
+      | h :: t -> (header_row h :: alignment_row) @ List.map ~f:data_row t
+    in
+    let tbl = Block.Table.make rows in
+    Block.Ext_table (tbl, meta)
   | { value = `List (kind, style, xs); location } ->
     let type' =
       match kind with
@@ -140,19 +175,31 @@ let rec nestable_block_element_to_block
     let l = Block.List'.make ~tight type' list_items in
     let meta = loc_to_meta location in
     Block.List (l, meta)
-  | { value = `Code_block (metadata, { value = code; location = code_loc })
+  | { value =
+        `Code_block
+          { meta = metadata
+          ; delimiter = _
+          ; content = { value = code; location = code_loc }
+          ; output
+          }
     ; location
     } ->
     let info_string =
       match metadata with
       | None -> Some ("ocaml", loc_to_meta code_loc)
-      | Some ({ value = lang; location = lang_log }, _env) ->
+      | Some { language = { value = lang; location = lang_log }; tags = _ } ->
         Some (lang, loc_to_meta lang_log)
     in
     let block_line = Block_line.list_of_string code in
     let code_block = Block.Code_block.make ?info_string block_line in
     let meta = loc_to_meta location in
-    Block.Code_block (code_block, meta)
+    let main_block = Block.Code_block (code_block, meta) in
+    let output_block =
+      match output with
+      | None -> []
+      | Some output -> [ nestable_block_element_list_to_block output ]
+    in
+    Block.Blocks (main_block :: output_block, meta)
   | { value = `Verbatim code; location } ->
     let info_string = Some ("verb", Meta.none) in
     let block_line = Block_line.list_of_string code in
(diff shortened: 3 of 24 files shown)


panglesd:
I'm converting to draft as this should only be released when odoc-parser 2.3.0 is released.

jonludlam:
odoc-parser.2.3.0 is now released.

I'd like to suggest that the parser is vendored in order to increase compatibility. odoc-parser is also vendored by ocamlformat and mdx for similar reasons. Right now we can't co-install odoc and ocaml-lsp-server because of this.

ocaml-lsp-server/src/doc_to_md.ml
```diff
@@ -165,6 +212,68 @@ let rec nestable_block_element_to_block
     let meta = loc_to_meta location in
     Block.Ext_math_block (code_block, meta)
 
+and nestable_block_element_to_inlines
```
Julow, line 215: This function might be a lot of maintenance and hurts the content.
Would it be reasonable to use raw HTML tags for tables ?

Julow (commented):
This PR is not a draft and is ready to be reviewed.

ocaml-lsp-server/src/doc_to_md.ml
panglesd: I agree that this function is not ideal...

An advantage of not using proper table compared to raw html is that the output might be more readable even without translating it to html. Also, we get cmarkit type guaranty.

Raw html tags could be an idea, but their start-end semantics not that simple (https://spec.commonmark.org/0.30/#html-blocks).

In particular, it seems to break out of the containing block. For instance,

```
- first item
  + first subitem
  + second subitem
```
renders by github as:

- first item
  + first subitem
  + second subitem

while with html inside (the blank line is the end condition fot the uninterpreted html)

```
- first item
  + first subitem
<table><tr><td>

useless table
</td></tr></table>

  + second subitem
```
renders as

- first item
  + first subitem
<table><tr><td>

useless table
</td></tr></table>

  + second subitem

@voodoos would you like me to investigate the "raw html tag" solution?

panglesd:
The test is failing because the test depend on ocamlformat 0.24.1, which depends on odoc-parser `>= "2.0.0" & < "2.3.0"`. So the dependencies and test dependencies are not co-installable!

The solutions are:
- Switch to a higher ocamlformat version where odoc-parser has been vendored, or
- vendor odoc-parser.

In any case, I guess it would make more sense in another PR?

ocaml-lsp-server/src/doc_to_md.ml
voodoos: > In particular, it seems to break out of the containing block

So in the case of tables the content would break out of the table ?

voodoos:
> I'd like to suggest that the parser is vendored in order to increase compatibility
I am not fond of vendoring, but this looks like a legitimate use case. As we see with this PR, reducing co-instability with `ocamlformat` which is a commonly used tool along `ocaml-lsp` might impact a lot of users.

> In any case, I guess it would make more sense in another PR?
I think we should do the vendoring in that PR, rename it `Upgrade to odoc-parser 2.3.0` and allow the CI to run correctly.

ocaml-lsp-server/src/doc_to_md.ml
panglesd: > So in the case of tables the content would break out of the table ?

What I meant is that the _whole table_ would break out of its _container block_ (recall that all of this is because, in odoc, tables are allowed to be nested in blocks, while it is not allowed in markdown).

For instance, in the example above, I had in mind a table inside a list (which is possible in odoc).

Julow suggested (IIUC) to render in this case the table, not by using markdown tables, but by embedding raw `<table>`/`<tr>`/`<td>` before the table/line/cell, and the corresponding closing tag after.
The markdown example above was thus the translation of a (nested) list containing a table, using this method.

However, we can see, from how github renders it, that in fact the table is not interpreted as inside the list, but "toplevel", and thus also breaks the rest of the list.
(I included a nested list to make it more visible that the included html breaks the flow of the list interpretation, but it might have created noise!)

It seems that cmarkit type system [won't enforce constraint](https://github.com/dbuenzli/cmarkit/blob/main/src/cmarkit.mli#L802) on raw html embedding, which is frightening!

panglesd:
> I think we should do the vendoring in that PR

Then I'll do that!

Just to be sure, would a vendoring in the style of dune or mdx, with a `vendor` directory and an `update-odoc-parser.sh` script downloading the sources from a given hash, be good for this project? Or (as I'm seeing a `submodule` directory) would you rather have the vendoring made with git submodules?

ocaml-lsp-server/src/doc_to_md.ml
voodoos: So that's not suitable since it would also break more visibly things like numbered lists (see below).
So far I think the more convincing workaround is your current proposition. 

```
- first item
  1. first subitem
<table><tr><td>

useless table
</td></tr></table>

  2. second subitem
  ```

- first item
  1. first subitem
<table><tr><td>

useless table
</td></tr></table>

  2. second subitem

**Comment by rgrinberg** [comment]
Just copying the part that you need is good enough for me. Submodules are more useful when you're also developing the vendored component in conjunction with this repo. I don't have the intention to do that, and I assume you two don't either.

**Comment by rgrinberg** [comment]
By the way, isn't this a breaking change in odoc-parser? Shouldn't there be a major version bump.

I also don't see anything in the change log regarding the breaking change.

**Comment by panglesd** [comment]
The odoc parser was in a separate repo (https://github.com/ocaml-doc/odoc-parser, the README needs to be updated to reflect this) but it was merged back in the odoc repository in https://github.com/ocaml/odoc/pull/973.

Currently, the version number from odoc and odoc-parser are tied together (so, `2.3.0`), but I cannot find a reason for this in the past discussions. I can only find:

> We're in the happy position that the current releases of odoc-parser and odoc are such that we can coalesce the version numbers so we'll have 2.3.0 as the next release of both, which is one less thing to worry about.

but I agree that it's not exactly true, since `odoc-parser` needed a major bump. There is also https://github.com/ocaml/odoc/pull/973#discussion_r1236953709 which opens up to untie the two versions later, if I understand correctly.

File: ocaml-lsp-server/src/doc_to_md.ml (hunk 1 above)
Comment (panglesd) [reply ocaml-lsp-server/src/doc_to_md.ml:215]: Update: The example was wrong, it is in fact possible to have html inside list items in commonmark.

(With the help of cmarkit) I could generate the right markdown for the example above:

So:
```
1. first item
   1. first subitem
      <table><tr><td>
      
      useless table
      </td></tr></table>
      
   2. second subitem
```
renders as

1. first item
   1. first subitem
      <table><tr><td>
      
      useless **table**
      </td></tr></table>
      
   2. second subitem

So I think the suggestion to use html is still valid, and the discussion open again. Speaking of tables, let me use one to summarize the differences:

| | Current approach | Html embedding |
|-|-|-|
|Source| :orange_circle: Source is not very readable due to tables not properly formatted, and blocks turned into inlines | :orange_circle: Source is not very readable, due to html tags and blank lines added all over the place |
|Faithful rendering| :red_circle: The content is modified | :green_circle: The content is preserved |
|Compatibility| :orange_circle: The renderer needs to know about tables, which is not in the commonmark standard | :orange_circle: The parser needs to follow the commonmark standard, which is different from the original one if I understand correctly |
|Rendering target | :green_circle: No constraint | :red_circle: Has to be HTML/markdown, otherwise (eg in latex) embedded html will certainly be ignored |
|Scariness| :green_circle: Not scary, a problem won't leak outside of the table | :orange_circle: Html embedding is a litlle bit scary as making a mistake can break the whole output...|

If the "rendering target" can handle embedded html, maybe the html embedding approach is better? The "content is preserved" is quite a strong benefit, I guess...

File: vendor/odoc-parser/src/dune [hunk 2]
```diff
@@ -0,0 +1,5 @@
+(ocamllex lexer)
+
+(library
+ (name odoc_parser)
+ (libraries astring result camlp-streams))
```
panglesd, line 5: In the dune's vendoring, all the dune files I took inspiration from (such as [this one](https://github.com/ocaml/dune/blob/main/vendor/fmt/src/dune)) have `(wrapped false)`.

I'm not sure if it's needed, but it still compiles without it..

panglesd:
The vendoring is ready for review. The compatiblity patch needs a new conclusion to the thread started by https://github.com/ocaml/ocaml-lsp/pull/1184#discussion_r1352667947 !

ocaml-lsp-server/src/doc_to_md.ml
voodoos: According to [the specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContent), LSP clients that have the capability to parse markdown should follow the Github Flavored Markdown Specification.

That plays in favor of Html but they also warn that:
>  *Please Note* that clients might sanitize the return markdown. A client could
>   decide to remove HTML from the markdown to avoid script execution.

voodoos: And there is more: the client can advertise which tags will actually be allowed in markdown with the "[MarkdownClientCapabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markdownClientCapabilities).allowedTags" capability. 

It's clear that if we want to respect the protocol we need to have at least two fallbacks:
- A text-only version when markdown is not allowed. (Right now `ocaml-lsp` simply prints the raw comment which is okay)
- A markdown without html when required tags are not allowed

And we could do better by using raw html when the client accepts it.

In any case we need to keep the current version, and I'm not sure having the additional html one is really worth it ?

panglesd: > In any case we need to keep the current version, and I'm not sure having the additional html one is really worth it ?

I agree with this analysis, and I do not think it is worth it: in the current approach, the content is only hurt in the case of table/lists inside tables, which I think will happen very seldomly.

If it turns out a problem, we can add the html approach later.

vendor/odoc-parser/src/dune
rgrinberg: It's not needed. It's needed under the following circumstance: the code is being vendored in a library so we need to avoid name collisions with the original library if we link it to the same binary.

In this case, we're using this odoc parser only in ocamllsp, so we can drop it.

dune-project
```diff
@@ -56,7 +56,9 @@ possible and does not make any assumptions about IO.
   ordering
   dune-build-info
   spawn
-  (odoc-parser (and (>= 2.0.0) (< 2.3.0)))
+  astring
+  result
```
rgrinberg, line 60: we don't need this shim. we don't support versions of OCaml that don't have the result module.

panglesd:
I opened https://github.com/ocaml-opam/opam-repository-mingw/pull/12 for the windows CI failure.

voodoos (approved):

flake.nix
```diff
@@ -110,7 +112,9 @@
               duneVersion = "3";
               buildInputs = with pkgs.ocamlPackages; [
                 ocamlc-loc
-                odoc-parser
+                astring
+                result
```
rgrinberg, line 116: `result` can be removed.

flake.nix
```diff
@@ -92,7 +92,9 @@
               ppx_yojson_conv_lib
               uutf
               lsp
-              odoc-parser
+              astring
+              result
```
rgrinberg, line 96: here as well.

PR #1186, opened 2023-09-22
Fix semantic highlighting of infix operators

### LINKED ISSUE #1172: Floating-point operators are highlighted differently than other operators
Floating point operators are syntax-highlighted differently than other operators. They get colored like types/modules instead of like other operators. Attached screenshot is from VSCode.

operator-highlighting

jfeser:
Closes #1172


rgrinberg (approved):

rgrinberg:
How come this is marked as not needing a change log entry?

Issue #420, opened 2021-04-19
ocaml-lsp 1.5.0 does not suggest keywords


Tchou:
While working on a code base with OCaml 4.12.0 in VS Code, it occurred to me that the OCaml Platform extension was not giving keywords such as `in` as possible completions (which quickly gets annoying because, you don't use in_channel_length all that often ;)).

When I switched back to OCaml 4.11.2 (with opam), the same extension works as expected, and has since ocaml-lsp-server 1.2.0 (which fixed #173).

I also tried the current master with 4.12.0 and it does not work either. I wiped clean both switches (no pins) and created a simple directory with a simple `.ml` file to ensure that this was not caused by a local switch, forgotten pin or other.

Screenshot with OCaml 4.12.0/ocaml-lsp-server 1.5.0:

![image](https://user-images.githubusercontent.com/248583/115275998-ccc04c00-a142-11eb-97d0-1e4627a6c40b.png)

Screenshot with OCaml 4.11.2/ocaml-lsp-server 1.4.1:

![image](https://user-images.githubusercontent.com/248583/115276051-df3a8580-a142-11eb-8946-92686f4bdfa4.png)

rgrinberg:
Unfortunately we had to remove this feature. It wasn't going to be upstreamed by the merlin devs in a useful form and it is kind of a pain to maintain. In addition, the merlin devs are working on a smarter alternative to keyword completion. I don't have more information about an ETA but perhaps @voodoos does.

Tchou:
I see. That's unfortunate indeed. For reference and as a workaround in vscode,  `"editor.acceptSuggestionOnEnter": "off"` makes for a reasonable experience (I haven't found a way to set a length threshold so that completion only kicks in after 3 chars).
The keyword `in` has a really bad combination of features here, which is why it would maybe warrant a special treatment : 
- it's a prefix of types (`int`, `in_channel`, …) and identifiers (`incr`, `input_*`, …) from `Stdlib`.
- its use is pervasive in idiomatic OCaml code
- it's often at the end of a line, thus followed by an `[enter]` keypress (which accepts the first completion in vscode default config)

The only other keyword with these characteristics that I found is `do` with `do_at_exit`, and is used far less often. `for` is not  problematic as it is rarely followed by a new line.

[...]

PR #1217, opened 2023-12-10
Completion for `in` keyword

ulugbekna:
We could have a hacky completion for `in` to ease the pain for https://github.com/ocaml/ocaml-lsp/issues/420 ? 

I don't know how much a more complete solution with all keywords support is necessary, but even as is this helps I think


rgrinberg:
LGTM. Could you disable this hack in mli files though?

rgrinberg (approved):

Issue #1198, opened 2023-10-16
feature request: "have a goto type link on hover"


Alizter:
In rust-analyzer, you get the following UX:
![image](https://github.com/ocamllabs/vscode-ocaml-platform/assets/8614547/4ca8db9b-0af2-4fe8-9696-f065dbe04509)

Hovering over an identifier (`key`) shows the type, just like in OCaml. Additionally it also has a `Go to <type>` link that the user can click on and go to the type. It would be great to have something similar in OCaml.

- We already have a go to type function, which should work for simple types like `int`.
- Type applications like `int list` we could do something more fancy and display both `int` and `list`. This would require us to know the types present in a complicated type like this. Not sure if merlin can supply us with this information cc @voodoos. It is currently a shortfall of the already existing "go to type" functionality.

It seems that the hover text request is markdown so we can supply as many links as we would like. I'm thinking something like:

```ocaml
\```ocaml
(int, int list) result
\```

Go to:
- [result](loc of result)
  - [int](loc of int)
  - [list](loc of list)
    - [int](loc of int)  
```

Obviously this introduces some degree of redundancy, but the structure of the links might help users dissect the type more easily. Especially in complicated cases like `format`.

PR #1218, opened 2023-12-11
Demystifying OCaml Syntax with Merlin and OCaml-LSP

PizieDust:
# FEATURE: Add Syntax Documentation to Hover request

## Introduction
This PR introduces a new feature to OCaml-LSP for both beginners and also more experienced programmers. No more searching for obscure keywords or symbols on Google trying to find documentation. 
With this PR, you can hover over OCaml code in your editor, and instantly see a clear and summarized explanation of its syntax along with a link to the exact portion of OCaml's documentation for it.

### Benefits:
- Faster understanding: At a glance you can get a quick description of what the syntax is about.
- Improved learning: Beginners can easily understand how their code works while writing the code.
- Enhanced productivity: Very useful for understanding codebases and what a syntax is supposed to do.

### How it works:
- Hover your cursor over the OCaml code.
- The hover request will call the new `Syntax Documentation` command from Merlin which returns a `JSON` output containing a `name`, some `description` and a `url` pointing to the documentation.
- This information is then parsed and displayed on your editors hover UI.

### Scope
Currently the command covers only specific sections of OCaml's syntax such as `Language Extensions`. Subsequently, more syntax will be covered.

## Code Example
Example: Hovering over type definitions in VS code:
Code example:
```ocaml
(*  Variant type *)
type p = 
  | A (** comment for A*)
  | B (** comment for B*)

(* Record type *)
type point = {x: int; y: int}
```

## Current Behaviour
Before `Syntax document` command: UI will display only type definition 
> image

> image

Or display comments if they are available
> image

## New Feature
The UI now includes a short description about the syntax highlighted under the cursor.

> image

UI will also display comments if they are available 

> image

New `Syntax Document` command can differentiate different kinds of syntax and return appropriate documentation

> image

## WIP/Dependencies
This PR relies on a new Merlin command found in this PR (https://github.com/ocaml/merlin/pull/1706) which is currently unmerged. To test this feature out, add a pin dependency of Merlin to https://github.com/PizieDust/merlin/tree/syntax_documentation.

This PR is related to https://github.com/ocaml/merlin/issues/1692 

> @pitag-ha suggested using `JSON` instead of one blob string for the `Syntax Document` command which works great for this purpose.

cc @voodoos @pitag-ha

### DIFF
--- a/ocaml-lsp-server/docs/ocamllsp/config.md
+++ b/ocaml-lsp-server/docs/ocamllsp/config.md
@@ -28,5 +28,12 @@ interface config {
   * @since 1.18
   */
   duneDiagnostics: { enable : boolean }
+
+  /**
+  * Enable/Disable Syntax Documentation
+  * @default false
+  * @since 1.18
+  */
+  syntaxDocumentation: { enable : boolean }
 }
 ```
--- a/ocaml-lsp-server/src/document.ml
+++ b/ocaml-lsp-server/src/document.ml
@@ -281,13 +281,23 @@ module Merlin = struct
     | `Found s | `Builtin s -> Some s
     | _ -> None
 
+  let syntax_doc pipeline pos =
+    let res =
+      let command = Query_protocol.Syntax_document pos in
+      Query_commands.dispatch pipeline command
+    in
+    match res with
+    | `Found s -> Some s
+    | `No_documentation -> None
+
   type type_enclosing =
     { loc : Loc.t
     ; typ : string
     ; doc : string option
+    ; syntax_doc : Query_protocol.syntax_doc_result option
     }
 
-  let type_enclosing doc pos verbosity =
+  let type_enclosing doc pos verbosity ~with_syntax_doc =
     with_pipeline_exn doc (fun pipeline ->
         let command = Query_protocol.Type_enclosing (None, pos, Some 0) in
         let pipeline =
@@ -308,7 +318,12 @@ module Merlin = struct
         | [] | (_, `Index _, _) :: _ -> None
         | (loc, `String typ, _) :: _ ->
           let doc = doc_comment pipeline pos in
-          Some { loc; typ; doc })
+          let syntax_doc =
+            match with_syntax_doc with
+            | true -> syntax_doc pipeline pos
+            | false -> None
+          in
+          Some { loc; typ; doc; syntax_doc })
 
   let doc_comment doc pos =
     with_pipeline_exn doc (fun pipeline -> doc_comment pipeline pos)
--- a/ocaml-lsp-server/src/hover_req.ml
+++ b/ocaml-lsp-server/src/hover_req.ml
@@ -201,7 +201,7 @@ let format_as_code_block ~highlighter strings =
 let format_type_enclosing ~syntax ~markdown ~typ ~doc
     ~(syntax_doc : Query_protocol.syntax_doc_result option) =
   (* TODO for vscode, we should just use the language id. But that will not work
-      for all editors *)
+     for all editors *)
   let syntax_doc =
     Option.map syntax_doc ~f:(fun syntax_doc ->
         sprintf
@@ -212,30 +212,32 @@ let format_type_enclosing ~syntax ~markdown ~typ ~doc
   in
   `MarkupContent
     (if markdown then
-        let value =
-          let markdown_name = Document.Syntax.markdown_name syntax in
-          let type_info = Some (format_as_code_block ~highlighter:markdown_name [ typ ]) in
-          let doc =
-            Option.map doc ~f:(fun doc ->
-                match Doc_to_md.translate doc with
-                | Raw d -> d
-                | Markdown d -> d)
-          in
-          print_dividers (List.filter_opt [ type_info; syntax_doc; doc ])
-        in
-        { MarkupContent.value; kind = MarkupKind.Markdown }
-      else
-        let value =
-          print_dividers (List.filter_opt [ Some typ; syntax_doc; doc ])
-        in
-        { MarkupContent.value; kind = MarkupKind.PlainText })
+       let value =
+         let markdown_name = Document.Syntax.markdown_name syntax in
+         let type_info =
+           Some (format_as_code_block ~highlighter:markdown_name [ typ ])
+         in
+         let doc =
+           Option.map doc ~f:(fun doc ->
+               match Doc_to_md.translate doc with
+               | Raw d -> d
+               | Markdown d -> d)
+         in
+         print_dividers (List.filter_opt [ type_info; syntax_doc; doc ])
+       in
+       { MarkupContent.value; kind = MarkupKind.Markdown }
+     else
+       let value =
+         print_dividers (List.filter_opt [ Some typ; syntax_doc; doc ])
+       in
+       { MarkupContent.value; kind = MarkupKind.PlainText })
 
 let format_ppx_expansion ~ppx ~expansion =
   let value = sprintf "(* ppx %s expansion *)\n%s" ppx expansion in
   `MarkedString { Lsp.Types.MarkedString.value; language = Some "ocaml" }
 
-let type_enclosing_hover ~(server : State.t Server.t) ~(doc : Document.t) ~with_syntax_doc
-    ~merlin ~mode ~uri ~position =
+let type_enclosing_hover ~(server : State.t Server.t) ~(doc : Document.t)
+    ~with_syntax_doc ~merlin ~mode ~uri ~position =
   let state = Server.state server in
   let verbosity =
     let mode =
@@ -263,7 +265,11 @@ let type_enclosing_hover ~(server : State.t Server.t) ~(doc : Document.t) ~with_
       v
   in
   let* type_enclosing =
-    Document.Merlin.type_enclosing merlin (Position.logical position) verbosity ~with_syntax_doc
+    Document.Merlin.type_enclosing
+      merlin
+      (Position.logical position)
+      verbosity
+      ~with_syntax_doc
   in
   match type_enclosing with
   | None -> Fiber.return None
@@ -299,7 +305,12 @@ let type_enclosing_hover ~(server : State.t Server.t) ~(doc : Document.t) ~with_
           client_capabilities
           ~field:(fun td -> Option.map td.hover ~f:(fun h -> h.contentFormat))
       in
-      format_type_enclosing ~syntax ~markdown ~typ ~doc:documentation ~syntax_doc
+      format_type_enclosing
+        ~syntax
+        ~markdown
+        ~typ
+        ~doc:documentation
+        ~syntax_doc
     in
     let range = Range.of_loc loc in
     let hover = Hover.create ~contents ~range () in
@@ -427,7 +438,14 @@ let handle server { HoverParams.textDocument = { uri }; position; _ } mode =
             | Some { enable = true } -> true
             | Some _ | None -> false
           in
-          type_enclosing_hover ~server ~doc ~merlin ~mode ~uri ~position ~with_syntax_doc
+          type_enclosing_hover
+            ~server
+            ~doc
+            ~merlin
+            ~mode
+            ~uri
+            ~position
+            ~with_syntax_doc
         | Some ((`Ppx_expr _ | `Ppx_typedef_attr _) as ppx_kind) -> (
           let+ ppx_parsetree =
             Document.Merlin.with_pipeline_exn
--- a/ocaml-lsp-server/src/hover_req.ml
+++ b/ocaml-lsp-server/src/hover_req.ml
@@ -193,9 +193,15 @@ let hover_at_cursor parsetree (`Logical (cursor_line, cursor_col)) =
   in
   !result
 
-let format_type_enclosing ~syntax ~markdown ~typ ~doc =
+let print_dividers sections = String.concat ~sep:"\n---\n" sections
+
+let format_as_code_block ~highlighter strings =
+  sprintf "```%s\n%s\n```" highlighter (String.concat ~sep:" " strings)
+
+let format_type_enclosing ~syntax ~markdown ~typ ~doc
+    ~(syntax_doc : Query_protocol.syntax_doc_result option) =
   (* TODO for vscode, we should just use the language id. But that will not work
-     for all editors *)
+      for all editors *)
   let syntax_doc =
     Option.map syntax_doc ~f:(fun syntax_doc ->
         sprintf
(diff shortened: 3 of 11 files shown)


jfeser:
I suspect that users will want this to be configurable. I could see how novice users might find this information helpful, but it's noisy. 

https://github.com/ocaml/ocaml-lsp/issues/1198 suggests go-to links for types. We could consider a go-to link for documentation, instead of inlining the documentation sentence into the hover.

PizieDust:
> #1198 suggests go-to links for types. We could consider a go-to link for documentation, instead of inlining the documentation sentence into the hover.

This is certainly a good idea. I like the [Go to Documentation]() format. It definitely trims is down a bunch.

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -12,31 +12,78 @@ let environment_mode =
...
+           | Markdown d -> d
+         in
+         sprintf
+           "```%s\n%s\n```\n---\n%s %s\n\n---\n\n%s\n\n---\n\n%s"
+           markdown_name
+           typ
+           (if String.length syntax_doc_name == 0 then ""
+           else syntax_doc_name ^ ":")
+           syntax_doc_des
+           (if String.length syntax_doc_url == 0 then ""
```
voodoos, line 62: You should not have to rely on such tests. Merlin commits to always give a manual page url so either there is no documentation at all, or the url should not be empty.

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -115,7 +167,17 @@ let handle server { HoverParams.textDocument = { uri }; position; _ } mode =
...
+            let syntax_doc_des = syntax_documentation.description in
+            let syntax_doc_url = syntax_documentation.documentation in
+            format_contents
+              ~syntax
+              ~markdown
+              ~typ
+              ~doc:documentation
+              ~syntax_doc_name
+              ~syntax_doc_des
+              ~syntax_doc_url
```
voodoos, line 180: There is no need to pass each field separately here. You can pass the whole record and destruct it later.

Also, there is sometimes no syntax documentation so you should pass a `Query.protocol.info option` or simply keep the `` `No_documentation / `Found _ `` values returned by merlin (as said in another comment).

voodoos (commented):
Made a few comments, you are doing good progress !

To make the feature opt-in (or opt-out ?) can take inspiration from other configurable features of ocaml-lsp like "extended hover".

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -12,33 +12,63 @@ let environment_mode =
...
+    String.concat ~sep:"\n\n --- \n\n" sections
+
+let format_as_code_block strings =
+  match strings with
+  | [] -> ""
+  | _ -> sprintf "```\n %s \n```" (String.concat ~sep:" " strings)
+
+
+let format_contents ~syntax ~markdown ~typ ~doc ~(syntax_documentation:[`Found of Query_protocol.syntax_doc_result
+| `No_documentation ]) ~syntax_doc_setting =
```
voodoos, line 28: This line will be reflect multiple other changes I suggested:
- We can call the argument `syntax_doc` here and everywhere else, it's shorter and mimic `doc`.
- There is no need to annotate the argument with it's type, especially when this type will simply be a `syntax_doc_result option`
- I think there it's too late to check if the feature is enabled: we don't want Merlin to perform extra work if it's not going to be displayed in the end.

ocaml-lsp-server/src/document.ml
```diff
@@ -308,7 +317,8 @@ module Merlin = struct
         | [] | (_, `Index _, _) :: _ -> None
         | (loc, `String typ, _) :: _ ->
           let doc = doc_comment pipeline pos in
-          Some { loc; typ; doc })
+          let syntax_doc_result = syntax_doc pipeline pos in
```
voodoos, line 320: This means that Merlin will have to perform extra work that most of the time gets thrown away. We should not call `syntax_doc` if the option is not enabled and always return `false` in that case.

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -12,33 +12,63 @@ let environment_mode =
...
-       in
-       { MarkupContent.value; kind = MarkupKind.PlainText })
+  let syn_doc = match 
+    (Env_vars._ENABLE_SYNTAX_DOCUMENTATION (), syntax_doc_setting) with 
+    | _, true | Some true, _  -> 
+      begin match 
+        syntax_documentation with 
+        | `Found syntax_documentation -> 
+            print_dividers [syntax_documentation.name^": "^syntax_documentation.description; 
+                            "Go to [Documentation]("^syntax_documentation.documentation^")"]
```
voodoos, line 38: The shorter, the better here: I think we can simply put the `See [Manual]` link at the end of the line without adding an extra separator.

We might also want to start the line with a constant prefix like `_syntax_ ` to make it clearer what the section is about ?

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -12,33 +12,63 @@ let environment_mode =
...
-     else
-       let value =
-         match doc with
-         | None -> sprintf "%s" typ
-         | Some d -> sprintf "%s\n%s" typ d
-       in
-       { MarkupContent.value; kind = MarkupKind.PlainText })
+  let syn_doc = match 
+    (Env_vars._ENABLE_SYNTAX_DOCUMENTATION (), syntax_doc_setting) with 
+    | _, true | Some true, _  ->
```
voodoos, line 33: As stated in other comments, checking that the feature is enabled should be done before we query Merlin. Then in this function we simply have to check if there is `Some syntax_doc` or `None`.

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -12,33 +12,63 @@ let environment_mode =
   | Some true -> Extended_variable
   | Some false | None -> Default
 
-let format_contents ~syntax ~markdown ~typ ~doc =
+let print_dividers sections =
+  match sections with
+  | [] -> ""
```
voodoos, line 17: This case is useless, `String.concat ~sep:"some sep" []` should already return `""`

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -12,33 +12,63 @@ let environment_mode =
...
+       match doc with
+       | None -> sprintf "%s" (print_dividers [type_info; syn_doc])
+       | Some s ->
+         let doc =
+           match Doc_to_md.translate s with
+           | Raw d -> sprintf "(** %s *)" d
+           | Markdown d -> d
+         in
+         
+         sprintf
```
voodoos, line 57: You don't need to call `sprintf` since `print_dividers` already returns a string (`sprintf "%s"` is always a useless call).

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -12,33 +12,63 @@ let environment_mode =
...
+    (Env_vars._ENABLE_SYNTAX_DOCUMENTATION (), syntax_doc_setting) with 
+    | _, true | Some true, _  -> 
+      begin match 
+        syntax_documentation with 
+        | `Found syntax_documentation -> 
+            print_dividers [syntax_documentation.name^": "^syntax_documentation.description; 
+                            "Go to [Documentation]("^syntax_documentation.documentation^")"]
+        | `No_documentation -> ""
+      end
+    | Some false, false | None, false -> ""
```
voodoos, line 41: This should be options, not empty strings for a more idiomatic approach.
Since `syntax_doc` will also be an option you will simply have to to an `Option.map`: 

```
Option.map syntax_doc ~f:(fun result -> sprintf "_syntax_ %s: %s. See [Manual](%s)" result....
```

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -12,33 +12,63 @@ let environment_mode =
...
+  in
+     `MarkupContent 
+    (if markdown then 
+     let value =
+       let markdown_name = Document.Syntax.markdown_name syntax in
+       let type_info = format_as_code_block [markdown_name; typ] in
+       match doc with
+       | None -> sprintf "%s" (print_dividers [type_info; syn_doc])
```
voodoos, line 49: Same here, you can keep `doc` as an option and use `let doc = Option.map doc ...`.

Then at the end, to print all non-None info with the required dividers you can simply filter the list:

```
print_dividers (List.filter_opt [type_info; syntax_doc; doc])
```

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -12,33 +12,63 @@ let environment_mode =
...
+let print_dividers sections =
+  match sections with
+  | [] -> ""
+  | _ ->
+    String.concat ~sep:"\n\n --- \n\n" sections
+
+let format_as_code_block strings =
+  match strings with
+  | [] -> ""
+  | _ -> sprintf "```\n %s \n```" (String.concat ~sep:" " strings)
```
voodoos, line 24: You're not formatting the code blocks the same as before (not sure if the extra spaces have an impact however).
But more importantly this function should have two arguments: the name and the type, which are both strings. No need for a list here given the usage of the function.

ocaml-lsp-server/docs/ocamllsp/config.md
```diff
@@ -21,5 +21,12 @@ interface config {
   * @since 1.16
   */
   codelens: { enable : boolean }
+
+  /**
+  * Enable/Disable Syntax Documentation
+  * @default false
+  * @since
```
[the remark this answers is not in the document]
PizieDust: Okay great. I didn't if to use the current version or the version if a new release is made so I just left it blank. I'll update in the next commit.

ocaml-lsp-server/src/document.ml
PizieDust: Thanks for pointing this out

smorimoto:
Changes for this PR on the VSCode extension side have just been merged.

ocaml-lsp-server/src/env_vars.ml
```diff
@@ -17,3 +17,9 @@ let _IS_HOVER_EXTENDED () : bool option =
   match v with
   | "true" | "1" -> Some true
   | _ -> Some false
+
+let _ENABLE_SYNTAX_DOCUMENTATION () : bool option =
```
rgrinberg, line 21: Is this config option experimental? If it's meant for end users, it shouldn't be an environment variable but a proper config option that can be passed when initializing the server.

PizieDust: So, currently it works via the environment variable OR via a config option, whichever way the user chooses to activate it. 
I'd environment variables are not very convenient, i could easily refactor to just use only config options.

rgrinberg: Yeah, I'd drop the env var. I mostly use them to ease debugging or for experimental stuff. Users already have the LSP config system to work with. That should be enough.

PizieDust: Great. The latest commit fixes this.

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -12,29 +12,40 @@ let environment_mode =
   | Some true -> Extended_variable
   | Some false | None -> Default
 
-let format_contents ~syntax ~markdown ~typ ~doc =
+let print_dividers sections = String.concat ~sep:"\n\n --- \n\n" sections
+
+let format_as_code_block ?(highlighter="ocaml") strings =
```
rgrinberg, line 17: Since this API is private, there's no need to maintain backwards compatibility. Thus, optional arguments aren't particularly useful and just allow the callers not to think about the correct value for this parameter. Making it required will be better

ocaml-lsp-server/src/document.ml
```diff
@@ -281,13 +281,23 @@ module Merlin = struct
     | `Found s | `Builtin s -> Some s
     | _ -> None
 
+  let syntax_doc pipeline pos =
+    let res =
+      let command = Query_protocol.Syntax_document pos in
```
rgrinberg, line 286: This query is available in only recent version of merlin, right? If so, an appropriate constraint should be added to the ocaml-lsp-server package (in dune-project).

rgrinberg (changes requested):

rgrinberg (approved):

rgrinberg:
I'll let @voodoos merge

voodoos:
> I'll let @voodoos merge

ack, I will try to release Merlin so that we can properly run the testsuite before merging.

voodoos (commented):
Sorry for the late review. Nice job Pizie, it looks like it's almost ready !

I think you changed the formatting of the on_hover responses slightly which breaks other tests. (The 5.0/5.1 CI only pass because it doesn't run the tests.) Did you run them on your machine ? Did they pass on your machine ?

For example, the following diff is clearly linked to your new printing of the dividers:
```
    - ---
    +
    +  --- 
    +
 ```
 
 There is no reason to change the results of other tests in this PR, unless the new feature is activated. You can run the `make test` target to run the test on your machine and check that the result is correct.

**Comment by PizieDust** [comment]
> Sorry for the late review. Nice job Pizie, it looks like it's almost ready !
> 
> I think you changed the formatting of the on_hover responses slightly which breaks other tests. (The 5.0/5.1 CI only pass because it doesn't run the tests.) Did you run them on your machine ? Did they pass on your machine ?
> 
> For example, the following diff is clearly linked to your new printing of the dividers:
> 
> ```
>     - ---
>     +
>     +  --- 
>     +
> ```
> 
> There is no reason to change the results of other tests in this PR, unless the new feature is activated. You can run the `make test` target to run the test on your machine and check that the result is correct.

Ahh yeah, I was using `dune build @runtest` and I think that's why I didn't notice that the tests were failing. It's all sorted now.

File: ocaml-lsp-server/src/hover_req.ml [hunk 16]
```diff
@@ -12,29 +12,40 @@ let environment_mode =
...
   (* TODO for vscode, we should just use the language id. But that will not work
      for all editors *)
+  let syntax_doc =
+    Option.map syntax_doc ~f:(fun syntax_doc ->
+        sprintf
+          "`syntax` %s: %s. See [Manual](%s)"
+          syntax_doc.name
+          syntax_doc.description
+          syntax_doc.documentation)
+  in
```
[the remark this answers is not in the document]
PizieDust: I think so too. There should be no reason using markdown syntax for clients which require only plain text.

voodoos:
Thanks @PizieDust, I will merge once the checks finish.

PR #1226, opened 2024-01-13
flake: remove flake-utils dependency

toastal:
By inlining the for loop from the utility library, downstream consumers do not need to pull in any additional dependencies & bloat their flake.lock files. Some additional reorganizing due to inlining as well—with the net effect of more lines deleted than added.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Remove `flake-utils` Nix Flake dependency
+
 ## Features
 
 - Introduce a configuration option to control dune diagnostics. The option is
--- a/flake.lock
+++ b/flake.lock
@@ -1,23 +1,5 @@
 {
   "nodes": {
-    "flake-utils": {
-      "inputs": {
-        "systems": "systems"
-      },
-      "locked": {
-        "lastModified": 1694529238,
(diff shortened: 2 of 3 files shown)


anmonteiro (commented):
> By inlining the for loop from the utility library, downstream consumers do not need to pull in any additional dependencies & bloat their flake.lock files.

- `flake-utils` adds a comparable number of lines to `flake.lock` than any other flake. I don't understand why this is considered "bloating"
    - `flake.lock` is a generated file that most people don't inspect, apart from when the git diff changes
- the utility library doesn't provide just "a basic for loop", and this reductionism feels a bit disingenuous; rather, it attempts to settle on a common set of systems that flake-utils users can support, apart from abstracting away the nitty-gritty of specifying the correct structure for systems in flake.nix.

Personally, and subjectively, this PR also affects readability. What was once a single function call to `eachDefaultSystem` now becomes 3 calls to `forAllSystems`.

toastal:
My stance on this has changed a bit…

You should be exporting overlays in Nix projects that can be imported by everyone, flakes or not. *flake-utils* can help smooth over that gap for flake users.

However, this project has its entire Nix code locked behind the experimental flake paradigm which is a much bigger issue that needs refactoring before deciding whether or not to keep *flake-utils*. In fixing the Nix fundamentally, it becomes clearer if this dependency is useful (I’d still argue it isn’t, but I would argue that flakes’ design is a more fundamental issue & *flake-utils* dependency is just one symptom).

PR #1233, opened 2024-02-23
wip: compatibility with merlin-lib 5.1-502

voodoos:
Preliminary support for the 414-5.2~preview version of merlin-lib for ocaml 5.2.

@jfeser the changes in the representation of functions breaks the beta-reduction feature of the action-inline code action. It would be great if you could find some time to restore it before the final release (sometime in April I think).


jfeser:
Dumb question, but how do I get the right version of merlin for building this PR? I've been using nix and I'm not very familiar with the opam pinning workflow.

jfeser:
I've pushed some commits here: https://github.com/jfeser/ocaml-lsp/tree/merlin-update. The new version of beta reduction is more correct but less aggressive. It only reduces fully applied functions and does not attempt to reduce functions with labeled parameters.

voodoos:
Thanks for taking the time @jfeser !
I will cherry-pick these commits to the update's branch.

.ocamlformat
```diff
@@ -1,4 +1,4 @@
-version=0.26.1
+version=0.26.2
```
rgrinberg, line 1: Can we hold off on this upgrade until 26.2 is available in nix? We could make an override for ocamlformat, but it doesn't seem like it's worth it.

rgrinberg:
> Dumb question, but how do I get the right version of merlin for building this PR? I've been using nix and I'm not very familiar with the opam pinning workflow.

I've added a flake input for merlin here https://github.com/ocaml/ocaml-lsp/pull/1264. So you can just set the branch/tag/commit in the flake.nix and then run `nix develop`.

.ocamlformat
voodoos: I dropped it in my last rebase.

ocaml-lsp-server/test/e2e-new/action_inline.ml
```diff
@@ -270,18 +272,15 @@ let _ =
...
   inline_test {|
 let _ =
   let $f : int -> int = fun x -> x in
   f 0
 |};
-  [%expect {|
-    let _ =
-      let f : int -> int = fun x -> x in
-      (0) |}]
+  [%expect {| |}]
```
rgrinberg, line 283: What happened in this test?

rgrinberg:
My plan was to wait for @awilliambauer to finish submitting his series of patches and make a release.  So that we can achieve the following two goals:

1. Release all the new features to users of 4.14 and 5.1
2. Sync the fork of ocamllsp and upstream

Once that is done, we'd no longer need to care about preserving 5.1 (and below) compatibility in the master branch

awilliambauer:
Just a heads up there will be some delay getting the rest of the patches submitted. It will be a little while before I have more time to look into the issue with https://github.com/ocaml/ocaml-lsp/pull/1290. We also have a patch to add more inlay hint configuration that involves changes to typedtree stuff that is different upstream, and it will take some work to port and test that.

rgrinberg:
Okay, in that case I’ll just disable this feature while running the test suite. The implementation is obviously correct, so I would still like end users to benefit from this.

ocaml-lsp-server/src/rename.ml
```diff
@@ -10,7 +10,9 @@ let rename (state : State.t)
     let command =
       Query_protocol.Occurrences (`Ident_at (Position.logical position), `Buffer)
     in
-    let+ locs = Document.Merlin.dispatch_exn ~name:"rename" merlin command in
+    let+ locs, _desync =
+      Document.Merlin.dispatch_exn ~name:"rename" merlin command
```
rgrinberg, line 14: Maybe we should warn about syncing when renaming as well?

dune-project
```diff
@@ -60,14 +60,13 @@ possible and does not make any assumptions about IO.
   astring
   camlp-streams
   (ppx_expect (and (>= v0.15.0) :with-test))
-  (ocamlformat (and :with-test (= 0.24.1)))
+  (ocamlformat (and :with-test (= 0.26.1)))
```
rgrinberg, line 63: I moved this one to https://github.com/ocaml/ocaml-lsp/pull/1303

ocaml-lsp-server/src/rename.ml
voodoos: In this PR renaming is still only local to the buffer.

OCaml-LSP tries to do smart things with syntactic sugar like punning, that require more work to do properly in the case of project-wide renaming (we need to load and parse every involved source file). I don't have a clean patch for that yet.

dune-project
xvw: I think it should be `0.26.2` (since `0.26.1` is not compatible with `5.2.0` (https://ocaml.org/p/ocamlformat/0.26.1)?

voodoos: Yes, but #1303 targeted master which is still OCaml < 5.1.
I will update this PR to actually perform the bump from 0.26.1 to 0.26.2.

ocaml-lsp-server/test/e2e-new/action_inline.ml
voodoos: cc @jfeser 

It looks like this does not fit the description you made of the "less agressive" beta reduction:
> The new version of beta reduction is more correct but less aggressive. It only reduces fully applied functions and does not attempt to reduce functions with labeled parameters.

voodoos:
@rgrinberg there is no `ppx_expect.common` library since `0.17.0` I updated the dune files in [2aee922](https://github.com/ocaml/ocaml-lsp/pull/1233/commits/2aee922f28e7a53ba79d543d0865310a91888fae).

rgrinberg:
> @rgrinberg there is no `ppx_expect.common` library since `0.17.0` I updated the dune files in [2aee922](https://github.com/ocaml/ocaml-lsp/pull/1233/commits/2aee922f28e7a53ba79d543d0865310a91888fae).

Can this be done in a separate PR?

ocaml-lsp-server/test/e2e-new/action_inline.ml
rgrinberg: Give me a chance to fix the nix compatibility before you merge though.

voodoos:
> Can this be done in a separate PR?

Will do.


About the release plan, what I have in mind right now is:

- Wait for `merlin-lib.4.16` to be released (ocaml/opam-repository#26052)
- Merge ocaml/ocaml-lsp#1265
- Release `ocaml-lsp` for 4.14 /  5.1
- Release `merlin-lib.5.1-502` and `ocaml-index.1.0` with support for project-wide occurrences
- Cleanup this upgrade PR, add a `post` dependency on `ocaml-index` and merge it
- Release `ocaml-lsp` for 5.2 with project-wide occurrences search

crackcomm:
I switched to OCaml 5.2 so I'm using `voodoos/ocaml-lsp#0573158` and lsp references work great.

There is one issue that I never had before: `inconsistent assumptions over interface`.

I used to be able to use `runtest -w` but now it doesn't work if the library (or executable) has no tests (or `inline_tests` stanza).
I have to use `esy dune build -w @ocaml-index` or `esy dune build -w @check` in the background.

I'm not sure if it's a known issue and if it's related to this PR (I jumped from 4.14 to 5.2) or it's just something I'll have to get used to :shrug:.

Let me know if I can provide any more information or help.

voodoos:
> I switched to OCaml 5.2 so I'm using `voodoos/ocaml-lsp#0573158` and lsp references work great.
> 
> There is one issue that I never had before: `inconsistent assumptions over interface`.
> 
> I used to be able to use `runtest -w` but now it doesn't work if the library (or executable) has no tests (or `inline_tests` stanza). I have to use `esy dune build -w @ocaml-index` or `esy dune build -w @check` in the background.
> 
> I'm not sure if it's a known issue and if it's related to this PR (I jumped from 4.14 to 5.2) or it's just something I'll have to get used to 🤷.
> 
> Let me know if I can provide any more information or help.

I doubt this is due to this PR, usually these issues are due to a mismatch between the compiler version and ocaml-lsp version. Does it still appear after a clean ? Maybe you don't use the same switch to build your project and to start ocaml-lsp ?

crackcomm:
I'm using esy and I tried rebuilding everything two times after removing all build and cache directories. I'm starting ocaml-lsp through esy and I don't even have ocaml & opam installed on my system other than through esy.

The lsp log shows:

```
[WARN][2024-07-11 11:09:02] ...lsp/handlers.lua:537	'failed to poll dune registry.\nUnix.Unix_error(Unix.ENOENT, "stat", "/run/user/1000/dune/rpc")\nMaybe you are not running dune in watch mode?\nHint: $ dune build -w'
```

dune is version 3.16.0

anmonteiro:
> This exception occurs when I try to hover on a module with _inconsistent assumptions over interface_.

you may want to run `esy` / `ocaml-lsp` inside the build sandbox rather than the execution sandbox (i.e. `esy b dune build`)

PR #1237, opened 2024-04-01
refactor: get doc kind info from merlin

andreypopp:
Currently doc kind (impl vs intf) is hardcoded as a set of file extensions.

Instead this PR tried to get kind info from merlin first, and if it fails, fallback to the old way.

Ref: https://github.com/ocaml/dune/pull/8567

### DIFF
--- a/ocaml-lsp-server/src/document.mli
+++ b/ocaml-lsp-server/src/document.mli
@@ -97,7 +97,7 @@ val close : t -> unit Fiber.t
     counterparts for the URI [uri].
 
     For instance, the counterparts of the file [/file.ml] are [/file.mli]. *)
-val get_impl_intf_counterparts : Uri.t -> Uri.t list
+val get_impl_intf_counterparts : Merlin.t option -> Uri.t -> Uri.t list
 
 (** [edits t edits] creates a [WorkspaceEdit.t] that applies edits [edits] to
     the document [t]. *)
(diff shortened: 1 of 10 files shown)


ocaml-lsp-server/src/document.mli
```diff
@@ -97,7 +97,7 @@ val close : t -> unit Fiber.t
     counterparts for the URI [uri].
 
     For instance, the counterparts of the file [/file.ml] are [/file.mli]. *)
-val get_impl_intf_counterparts : Uri.t -> Uri.t list
+val get_impl_intf_counterparts : Merlin.t option -> Uri.t -> Uri.t list
```
andreypopp, line 100: I kept this to accept `Merlin.t option` as I understand it still could work for docs outside of the current LSP workspace (e.g. do jump to definition to some opam installed lib and then still can use this switch impl/intf command to switch between .ml/.mli).

rgrinberg (approved):
LGTM. Let's see what @voodoos says about using the merlin API this way

voodoos:
Looking at the configuration suffixes before fallbacking to standard extensions seems like the correct thing to do, yes.

rgrinberg:
Could you add a CAHNGES entry? Maybe include a test as well. I have no idea how hard it is to write one for this feature though.

andreypopp:
Updated `CHANGES` under `Fixes`. But not clear to me, indeed, how to test this easily...

PR #1238, opened 2024-04-01
Support `--context` flag of `dune ocaml-merlin`

jchavarri:
This PR integrates two new Dune commands into the LSP server. See related issue  [vscode-ocaml-platform#1432](https://github.com/ocamllabs/vscode-ocaml-platform/issues/1432).

## Dependencies and related PRs

This PR relies on features being added in [dune#10324](https://github.com/ocaml/dune/pull/10324).

It also unblocks the feature implemented in [vscode-ocaml-platform#1449](https://github.com/ocamllabs/vscode-ocaml-platform/pull/1449). A screen capture demo is shown in that issue.

## Description

- Add a new request `ocamllsp/getDuneContexts` that will return the list of available Dune contexts in the workspace (through the newly added `dune describe contexts` subcommand)
- Introduce a new command line flag `--context` in the server. This flag, when present, will be passed through to the `dune ocaml-merlin` instance being created.


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -931,10 +936,11 @@ let run_in_directory =
   let for_windows = !Merlin_utils.Std.System.run_in_directory in
   fun () -> if Sys.win32 then for_windows else run_in_directory
 
-let run channel ~read_dot_merlin () =
+let run channel ~dune_context ~read_dot_merlin () =
   Merlin_utils.Lib_config.set_program_name "ocamllsp";
   Merlin_utils.Lib_config.System.set_run_in_directory (run_in_directory ());
   Merlin_config.should_read_dot_merlin := read_dot_merlin;
+  Merlin_config.dune_context := Option.map dune_context ~f:Fun.id;
```
andreypopp, line 943:
```suggestion
  Merlin_config.dune_context := dune_context;
```

ocaml-lsp-server/src/custom_requests/req_dune_contexts.ml
```diff
@@ -0,0 +1,49 @@
...
+  match Bin.which "dune" with
+  | None ->
+    Jsonrpc.Response.Error.raise
+      (Jsonrpc.Response.Error.make
+         ~code:InternalError
+         ~message:"dune binary not found"
+         ())
+  | Some dune -> (
+    let describe =
+      "DUNE_CONFIG__GLOBAL_LOCK=disabled " ^ Fpath.to_string dune
```
andreypopp, line 17: `DUNE_CONFIG__GLOBAL_LOCK=disabled`
^ :(

but the question is — is `dune describe contexts` command's output stable?

jchavarri: I see the need for `DUNE_CONFIG__GLOBAL_LOCK` as a bug / limitation of Dune, rather than a hack in this PR. The main issue is that Dune doesn't differentiate between commands that write in `_build` vs commands that only read. So it will always lock the folder regardless. I opened an issue to discuss making the lock mechanism smarter: https://github.com/ocaml/dune/issues/10430

> but the question is — is `dune describe contexts` command's output stable?

This subcommand is being added in the companion PR https://github.com/ocaml/dune/pull/10324. It only outputs the names of the available contexts, so it shouldn't need to change over time.

jchavarri:
I talked to @andreypopp and he suggested to remove the `duneContexts` protocol command, and instead call `dune describe contexts` directly from the vscode extension. I applied these changes in https://github.com/ocaml/ocaml-lsp/pull/1238/commits/9ada8752ddf9fefd220699fcb530491df7fff06d and its companion commit https://github.com/ocamllabs/vscode-ocaml-platform/pull/1449/commits/9d348689874890487734bb20b02d1daeb282d7c2.

ocaml-lsp-server/src/merlin_config.ml
```diff
@@ -192,7 +194,17 @@ module Process = struct
...
+          match !dune_context with
+          | None -> [ prog; "ocaml-merlin"; "--no-print-directory" ]
+          | Some dune_context ->
+            [ prog
+            ; "ocaml-merlin"
+            ; "--no-print-directory"
+            ; "--context"
+            ; dune_context
+            ]
+        in
```
andreypopp, line 207:
```suggestion
        let argv =
          match !dune_context with
          | None -> argv
          | Some dune_context -> argv @ [ "--context"; dune_context ]
        in
```

ocaml-lsp-server/src/merlin_config.ml
```diff
@@ -244,7 +256,7 @@ module Dot_protocol_io =
 let should_read_dot_merlin = ref false
 
 type db =
-  { running : (string, entry) Table.t
+  { running : (String.t, entry) Table.t
```
andreypopp, line 259: nit/curious: why this change?

andreypopp (approved):
Looks good!

ocaml-lsp-server/src/merlin_config.ml
jchavarri: i can't remember. just removed it 👍

jchavarri:
Thanks for the reviews @andreypopp. From my side, there's nothing else to add to this PR. Also, the parent PR in dune https://github.com/ocaml/dune/pull/10324 has been merged, and it should be included in next release 3.16.

voodoos:
@jchavarri is it the responsibility of the editor plugin to not use that flag if the installed dune version does not understand it (`< 3.16`)

CHANGES.md
```diff
@@ -8,6 +8,8 @@
 
 - Support folding of `ifthenelse` expressions (#1031)
 
+- Add `--context` flag (#1238)
```
voodoos, line 11: I think we could add a bit more context here at least define what "context" means here.

voodoos:
@jchavarri is this PR ready ? 

What will happen if the installed dune does not understand that flag ?

rgrinberg:
I don't think this is needed anymore. Turns out that contexts alone aren't good enough for melange support.

jchavarri:
@rgrinberg Should https://github.com/ocaml/dune/pull/10324 be reverted? That PR was added exclusively to support this use case. If context selection is not valid for any other use cases besides melange, then it probably doesn't make much sense to keep the code added in that PR.

rgrinberg:
Yes, there's probably no point to it. Though I think we should focus on implementing the new feature and what would that look like. We can clean things up in the end.

PR #1239, opened 2024-04-17
stop server on exit

jchavarri:
I noticed that the vscode extension was stopping [only the client but not the server](https://github.com/ocamllabs/vscode-ocaml-platform/blob/bb60961679f248963204a71853e28e2454931d62/src/extension_instance.ml#L84-L90). This can potentially leave server processes hanging.

The protocol [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#exit) is clear:

> A notification to ask the server to exit its process. The server should exit with `success` code 0 if the shutdown request has been received before; otherwise with `error` code 1.

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -534,7 +534,9 @@ let on_request :
     | None -> now None
     | Some doc -> now (Some (Msource.text (Document.source doc))))
   | DebugEcho params -> now params
-  | Shutdown -> Fiber.return (Reply.now (), state)
+  | Shutdown ->
+    let state = { state with shutdown_status = Shutdown_received } in
+    Fiber.return (Reply.now (), state)
   | WorkspaceSymbol req ->
     later (fun state () -> Workspace_symbol.run server state req) ()
   | CodeActionResolve ca -> now ca
@@ -743,8 +745,15 @@ let on_notification server (notification : Client_notification.t) :
   | WillSaveTextDocument _
   | Initialized
   | WorkDoneProgressCancel _
-  | WorkDoneProgress _
-  | Exit -> Fiber.return state
+  | WorkDoneProgress _ -> Fiber.return state
+  | Exit ->
+    let exit_code =
+      (* https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#exit *)
+      match state.shutdown_status with
+      | Shutdown_received -> 0
+      | Shutdown_not_received -> 1
+    in
+    exit exit_code
   | SetTrace { value } -> Fiber.return { state with trace = value }
   | UnknownNotification req ->
     let+ () =
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,8 @@
   This enables more lsp features for non-.ml/.mli files. Though it still
   depends on merlin's support. (#1237)
 
+- Stop server when exit notification is received (#1239)
+
 # 1.17.0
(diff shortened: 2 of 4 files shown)


voodoos:
These changes look reasonable to me. @rgrinberg was there a specific reason why the server was not respecting the `exit` notification ?

rgrinberg:
No idea why we ignored it. Probably just didn't get around to it.

Could you add a test for this?

ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -743,8 +745,15 @@ let on_notification server (notification : Client_notification.t) :
...
-  | Exit -> Fiber.return state
+  | WorkDoneProgress _ -> Fiber.return state
+  | Exit ->
+    let exit_code =
+      (* https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#exit *)
+      match state.shutdown_status with
+      | Shutdown_received -> 0
+      | Shutdown_not_received -> 1
+    in
+    exit exit_code
```
rgrinberg, line 756: This looks wrong. With this, any sub-processes launched by lsp would not be cleaned up

rgrinberg (changes requested):

ocaml-lsp-server/src/ocaml_lsp_server.ml
jchavarri: > This looks wrong. With this, any sub-processes launched by lsp would not be cleaned up

How does `Client.stop` handle this? I can see some finalization steps being taken here

https://github.com/ocaml/ocaml-lsp/blob/cb338c754bcc99c329fb45e0c4554c387cd185f2/ocaml-lsp-server/src/ocaml_lsp_server.ml#L828-L837

but I have no idea if they apply for the sub-processes being referred to.

jchavarri:
> Could you add a test for this?

I started looking into how to write a test for this. My initial step was to verify that the lsp server was actually being killed forcefully in the tests by some timeout, but that doesn't seem to be the case, see related https://github.com/ocaml/ocaml-lsp/pull/1240.

If the server is stopping itself when `Client.stop` is called, then I wonder how can I test the behavior added in this PR?

jchavarri:
My assumptions were wrong, it seems the server exist properly on `Exit` notification. I added a test to check the behavior in `master` in https://github.com/ocaml/ocaml-lsp/pull/1241. Please review that PR first and then we can finalize this one, if it makes sense. I looked further in the `master` branch, but I still don't understand how the server gracefully stops on `Exit` 😕

jchavarri:
I was looking in the wrong handler:
https://github.com/ocaml/ocaml-lsp/blob/2fa2e84f8a7e507e702e113e037072280c6ae14c/ocaml-lsp-server/src/ocaml_lsp_server.ml#L747

The code that stops the server is on a handler in the `lsp-fiber` `Rpc` module (??):

https://github.com/ocaml/ocaml-lsp/blob/2fa2e84f8a7e507e702e113e037072280c6ae14c/lsp-fiber/src/rpc.ml#L415-L419

I guess that `Rpc`'s `State.t` type could be adapted to include information about `Shutdown` having been received or not, but idk how this information can be passed over to the executable exit code. I haven't found how a channel closing relates to this exit code.

rgrinberg:
Yes, that makes sense. We want to handle exit generically in the generic lsp server.

You could set the state to say if you've received the exit response and then set the exit code based on it `Ocaml_lsp_server.start` is done

PR #1240, opened 2024-04-17
test: remove unused timeout

jchavarri:
It seems that somehow when calling `Client.stop`, the server will also halt itself. I think it might this call to `Session.stop`:

https://github.com/ocaml/ocaml-lsp/blob/cb338c754bcc99c329fb45e0c4554c387cd185f2/lsp-fiber/src/rpc.ml#L346

But I'm not sure exactly why, there is a lot of code between lsp-fiber, jsonrpc-fiber and regular fiber...

### DIFF
--- a/ocaml-lsp-server/test/e2e-new/test.ml
+++ b/ocaml-lsp-server/test/e2e-new/test.ml
@@ -123,32 +123,13 @@ end = struct
       let client = Client.make handler io () in
       f client
     in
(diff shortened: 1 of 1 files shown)


rgrinberg:
Indeed the timeout doesn't do anything when your tests are passing. It proves to be very useful however if your tests are buggy and the ocamllsp instance stays alive.

jchavarri:
Right, I realized that just now when working on #1241 😅

PR #1241, opened 2024-04-17
test: add tests for exit and shutdown notifications

jchavarri:
As requested in https://github.com/ocaml/ocaml-lsp/pull/1239#issuecomment-2061097336.

It seems that the `ocamllsp` server behaves according to the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#exit), except for the case where `Exit` is sent without a previous `Shutdown`, in which case it exits with code 0, but it should be 1.

### DIFF
--- a/ocaml-lsp-server/test/e2e-new/dune
+++ b/ocaml-lsp-server/test/e2e-new/dune
@@ -50,6 +50,7 @@
     completion
     doc_to_md
     document_flow
+    exit_notification
     for_ppx
     hover_extended
     inlay_hints
--- a/ocaml-lsp-server/test/e2e-new/exit_notification.ml
+++ b/ocaml-lsp-server/test/e2e-new/exit_notification.ml
@@ -2,91 +2,18 @@ open Test.Import
 
 let client_capabilities = ClientCapabilities.create ()
 
-(* Mostly a copy of Test.run, but without optional args and with some extra
-   printing to track how `ocamllsp` executable exited *)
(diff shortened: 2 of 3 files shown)


jchavarri:
The tests are failing because the changelog is missing, but I don't think this PR deserves an entry.

rgrinberg (changes requested):
Why did you decide to copy paste `Test.T` into your test case? I might be missing something

rgrinberg (changes requested):
Why did you decide to copy paste `Test.T` into your test case? I might be missing something

jchavarri:
> Why did you decide to copy paste `Test.T` into your test case? I might be missing something

To be able to print stuff to the console on specific places that were not covered in `Test.T`, for example look for the strings `"ocamllsp finished with code = %d@.%!"` or `"ocamllsp process timed out\n"`.

rgrinberg:
It's still way too much copy pasting to justify adding 3 print statements. You could instead modify the run function to return the status of the process when it's finished

jchavarri:
Tried something in https://github.com/ocaml/ocaml-lsp/pull/1241/commits/7e73fea1126d8da93277f758123d3b8375d21c1d. I wasn't too sure how to pass the status through the `Fiber.all_concurrently` function (which expects all fibers to return a uniform type), so tried to handle it with a new variant type + unwrap with exceptions right after it.

PR #1245, opened 2024-05-07
Improving hover behavior

awilliambauer:
This PR makes several changes to hover behavior to reduce overly noisy hovers, and adds hovering on some PPXes to see the PPX expansion.

### Restricting unhelpful hovers
The LSP currently provides hovers in many places that are unhelpful and noisy, including on comments and many keywords. 
This patch, instead of providing a hover on every position where `Document.Merlin.type_enclosing` returns a result, uses the `hover_at_cursor` function to iterate over the parsetree and select the subset of nodes where hover information is useful.
The primary impact of this is the LSP no longer provides hovers on whitespace, comments, or most keywords and punctuation.

### Suppressing multiline highlights
The `TextDocumentHighlight` request can highlight very large ranges in a way that's distracting and not helpful.
This happens frequently with some PPXes.
This PR adds logic to filter out highlights with multiline ranges.

### Adding hovers on PPXes
This PR adds hovers on some PPXes that show the expanded code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,15 @@
 
 - Support folding of `ifthenelse` expressions (#1031)
 
+- Improve hover behavior (#1245)
+
+  Hovers are no longer displaye on useless parsetree nodes such as keywords,
+  comments, etc.
+
+  Multiline hovers are now filtered away.
+
+  Display expanded ppx's in the hover window.
+
 ## Fixes
 
 - Detect document kind by looking at merlin's `suffixes` config.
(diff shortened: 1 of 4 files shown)


rgrinberg:
Awesome work. I fixed up the tests, formatting, and added a CHANGES entry.

PizieDust:
Currently, it doesn't handle:
- hover over exceptions

It is my understanding that PPX derivers work on 8 different nodes:
- type declaration in structure
- type declaration in signature
- type extension in structure
- type extension in signature
- exception in structure
- exception in signature
- module type declaration in structure
- module type declaration in signature

I think this PR handles mostly only `type_declarations` and won't display the expanded ppx for other types of nodes.

PR #1246, opened 2024-05-07
Disable trigger-character completions and signature help in comments

awilliambauer:
When the user requests signature help or completions via a trigger character, do not provide a repsonse if the request was generated from within a comment. This reduces unhelpful noise when writing comments.

### DIFF
--- a/ocaml-lsp-server/src/check_for_comments.ml
+++ b/ocaml-lsp-server/src/check_for_comments.ml
@@ -1,0 +1,16 @@
+open Import
+
+let position_in_comment ~position ~merlin =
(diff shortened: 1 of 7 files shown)


ocaml-lsp-server/src/check_for_comments.ml
```diff
@@ -0,0 +1,16 @@
+open Import
+
+let position_in_comment ~position ~merlin =
+  let loc_contains_position (_, (loc : Merlin_parsing.Location.t)) =
```
rgrinberg, line 4: Not sure `Merlin_parsing` is coming from. I fixed this to be `Loc.t`

PR #1265, opened 2024-05-13
Draft: Add a custom query for raw invocation of Merlin

xvw:
This PR is based on #1233 and depends on https://github.com/ocaml/merlin/pull/1758 

It introduces the custom query: `ocamllsp/merlinCallCompatible`, to call Merlin directly via LSP.


ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
```diff
@@ -0,0 +1,156 @@
...
+let partition_args =
+  (* The function is relatively optimistic and attempts to treat literal data as
+     strings of characters. *)
+  function
+  | `String s -> Either.Left s
+  | `Bool b -> Either.Left (string_of_bool b)
+  | `Float f -> Either.Left (string_of_float f)
+  | `Int i -> Either.Left (string_of_int i)
+  | `Intlit i -> Either.Left i
+  | x -> Either.Right x
```
voodoos, line 55: The field `args` is already an array of strings: `"args": string[]`
Why do you need to parse it as if it was an array of heterogeneous values ?

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
```diff
@@ -0,0 +1,156 @@
...
+    let merlin = Document.merlin_exn doc in
+    Document.Merlin.with_pipeline_exn merlin f
+  | None ->
+    let message =
+      "Document " ^ Uri.to_string uri ^ " wasn't found in the document store"
+    in
+    raise_invalid_params ~message ()
+
+let parse_arguments state uri specs raw_args cmd_args =
+  let conf = Merlin_config.DB.get state.State.merlin_config uri in
```
voodoos, line 109: @rgrinberg I am not familiar with how ocaml-lsp handles configuration, but I would expect it to be retrived from the pipeline here with `Mpipeline.final_config`. What is the correct thing to do ?

voodoos (commented):
Just made a few comments but overall the changes looks correct.

You chose to write the json parser manually, but maybe it would have been worth it to derive ?

xvw:
> You chose to write the json parser manually, but maybe it would have been worth it to derive ?

Derivation doesn't seem to be particularly used in the code base, so I didn't want to introduce it in this PR. In an ideal world, I'd probably go for applicative validation, with the combinators that go with it. But I think that could be the subject of another PR. WDYT?

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
xvw: It is not, it is an array of `Yojson-ish` (ie: a list of `[> \`String of string]`).

voodoos: Right, but we could also consider invalid any other values than ``` `String ```.

I don't think it's wrong to accept heterogeneous values, but then it should be said in the spec:
````
```json
{
  "uri": DocumentUri,
  "command": string,
  "args": <string|bool|...>[],
  "resultAsSexp": boolean
}
```
````

I am just not sure that this is worth the added complexity...

Comment (xvw) [reply ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml:55]: Ok, I'll probably change the specs

File: ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml [hunk 3]
```diff
@@ -0,0 +1,156 @@
...
+  match List.assoc_opt "args" params with
+  | Some (`List args) -> (
+    (* Only lists containing stringish data are accepted. *)
+    match List.partition_map ~f:partition_args args with
+    | args, [] -> Some args (* No invalid arg. *)
+    | _, _ :: _ -> None (* Some invalid args. *))
+  | _ ->
+    (* If args is not a list or is absent, we assume that an empty list is
+       passed. *)
+    Some []
```
rgrinberg, line 67: Does that mean we just ignore non list arguments? Doesn't seem intuitive.

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
rgrinberg: The correct thing to do would be to fetch the document and then call `Document.Merlin.mconfig`

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
```diff
@@ -0,0 +1,156 @@
...
+  in
+  `Assoc [ ("class", `String class_); ("value", output) ]
+
+let on_request ~params state =
+  Fiber.of_thunk (fun () ->
+      let { result_as_sexp; command; args; uri } =
+        from_structured_json_exn params
+      in
+      match
+        Merlin_commands.New_commands.(find_command command all_commands)
```
rgrinberg, line 140: Given that this API is new, can we make it not raise an exception?

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
rgrinberg: If merlin accepts named arguments, then 3. Otherwise, 1.

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
```diff
@@ -0,0 +1,156 @@
+open Import
+
+let capability = ("handleMerlinCallCompatible", `Bool true)
+
+let meth = "ocamllsp/merlinCallCompatible"
+
+type params =
```
rgrinberg, line 7: This type should be in its own submodule and should have the following two functions:

```ocaml
val yojson_of_t : t -> Json.t
val t_of_yojson : Json.t -> t
```

Even if the two functions above are manually written, they should conform to our API.

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
```diff
@@ -0,0 +1,156 @@
...
+    | None ->
+      let data =
+        `Assoc
+          [ ("expectedParams", expected_params)
+          ; ("receivedParams", (params :> Json.t))
+          ]
+      in
+      raise_invalid_params ~data ~message:"Unexpected params format" ())
+
+let with_pipeline state uri f =
```
rgrinberg, line 97: You don't really need this function. `Document_store.get` already does the error handling for bad documents for you.

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
```diff
@@ -0,0 +1,156 @@
...
+    config
+    cmd_args
+  |> snd
+
+let perform_query action params pipeline =
+  let action () = action pipeline params in
+  let class_, output =
+    match action () with
+    | result -> ("return", result)
+    | exception Failure message -> ("failure", `String message)
```
rgrinberg, line 127: When is this exception possible? if this is a user facing error, it should have its own exception type.

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
```diff
@@ -0,0 +1,156 @@
...
+    | exception exn ->
```
rgrinberg, line 128: There's already a catch-all for exceptions so I think this is not necessary.

xvw: Why is this? Catching the exception returns the result in a dedicated response class and mimics the behaviour of the merlin binary.

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
xvw: Regarding https://github.com/ocaml/ocaml-lsp/pull/1265#pullrequestreview-2070665576, I'll probably go for derivation. But may I ask why (it is about curiosity)? Since in other custom query, This is sometimes the case (in a `Request_param` module, which is never exposed) sometimes not (ie: when a list of `uri` is requested), so does this have a different implication (code-style, additional metrics?)

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
xvw: (Specs changed)

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
xvw: Ok, fixed on https://github.com/ocaml/ocaml-lsp/pull/1265/commits/97471ecdd0a16f214a720331b97a32c2cefa56d0. Thanks for the pointer!

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
xvw: Fixed

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
```diff
@@ -0,0 +1,181 @@
...
+    let specs = List.map ~f:snd specs in
+    let args =
+      Mconfig.parse_arguments
+        ~wd:(Sys.getcwd ())
+        ~warning:ignore
+        specs
+        raw_args
+        config
+        cmd_args
+      |> snd
```
voodoos, line 137: I don't think we should ignore the new configuration here: command line arguments might change the configuration for a specific query. 

I pushe a test illustrating this in 7799ef878836a476257c85b8d02cb2e5f1b2eaaf

xvw: Thanks for the spot!

xvw: Fixed! Thanks a lot

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
```diff
@@ -0,0 +1,180 @@
...
+          [ ("expectedParams", Request_params.expected)
+          ; ("receivedParams", (params :> Json.t))
+          ]
+      in
+      raise_invalid_params ~data ~message:"Unexpected params format" ())
+
+let with_pipeline state uri specs raw_args cmd_args f =
+  match Document_store.get_opt state.State.store uri with
+  | Some doc ->
+    let merlin = Document.merlin_exn doc in
```
voodoos, line 125: I think you should not rely on `merlin_exn` here but match on the possible cases as is done here for example:

https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/src/ocaml_lsp_server.ml#L172-L174

voodoos (approved):
Thanks @xvw, these changes looks reasonable to me.

ocaml-lsp-server/src/document.mli
```diff
@@ -55,6 +55,9 @@ module Merlin : sig
 
   val with_pipeline_exn : ?name:string -> t -> (Mpipeline.t -> 'a) -> 'a Fiber.t
 
+  val with_configurable_pipeline_exn :
+    ?name:string -> config:Mconfig.t -> t -> (Mpipeline.t -> 'a) -> 'a Fiber.t
```
voodoos, line 59: I think we should document how configuration is handled in both `with_pipeline_exn` and `with_configurable_pipeline_exn`.

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
rgrinberg: Returning a dedicated response is already the default behavior. It's a matter of choosing the correct error code I suppose. Does merlin guarantee that every `Failure` that it raises is caused by the caller making an error? Or are some `Failure` related to the server misbehaving in general?

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
rgrinberg: Looks good to me. The reason why we always use the interface above is to make it trivial to switch between manually written json converters and auto generated ones.

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
rgrinberg: No change is worth a re-release of any existing version in my eyes. However, a new version that contains the improved API is maybe a possibility?

As a side note, It feels wrong to introduce a suboptimal API and refuse to improve it before it even has any users. Are we waiting so it to gain some users before we make a change so that we can break the API for them? Seems counter-intuitive.

This particular issue isn't very important, but I do think it points out a more general flaw of how development is handled. There should be a tighter feedback loop between how the merlin library is developed and its only client (for the forseable future) so that we realize when an API is suboptimal ASAP and not after it's too late to change it.

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
xvw: Done in https://github.com/ocaml/ocaml-lsp/pull/1265/commits/738dffdd0ea893b8fdb309bc037c851d7e4b20dd but it returns `Null` if the document is not a merlin one.

ocaml-lsp-server/src/document.mli
xvw: Done in https://github.com/ocaml/ocaml-lsp/pull/1265/commits/055f23db04f988e34e136a34a402093d2250e16a but feel free to rephrase it (or suggest rephrase) if it does not fit your expectation.

ocaml-lsp-server/test/e2e/__tests__/ocamllsp-merlinCallCompatible.ts
```diff
@@ -0,0 +1,138 @@
+import outdent from "outdent";
+import * as LanguageServer from "../src/LanguageServer";
+import * as Protocol from "vscode-languageserver-protocol";
+import * as Types from "vscode-languageserver-types";
+
+describe("ocamllsp/merlinCallCompatible", () => {
```
rgrinberg, line 6: I should have mentioned this earlier, but we try not to extend the typescript test suite ever since we've introduced the OCaml LSP client. Do you think you can port these tests to e2e-new?

xvw: Of course yes!

xvw: Done in https://github.com/ocaml/ocaml-lsp/pull/1265/commits/2e506c6a8e37b7b0a356c4f07eb469601e371fb0, thanks for the feedback!

ocaml-lsp-server/src/custom_requests/req_merlin_call_compatible.ml
voodoos: It's true that we still have a somewhat slow release cycle for Merlin, but we are in no way opposed to make quick evolution to the API when needed. We are hopefully leaning toward the tighter feedback loop you're mentioning now that we are getting more hands-on with ocaml-lsp dev. 

This particular function is so ad-hoc in its usage by ocaml-lsp, and exposing a deep merlin-specific behavior, that it did not seems worthy to rework its API when exposing it. It is very unlikely that this part of the API will ever change, unless we stop supporting the legacy Merlin protocol altogether. 

Still, since we have releases planned soon anyway, I don't mind adding an option or result-based version of that function to the API, which we will also backport to older versions of merlin-lib so that this PR can be merged.

ocaml-lsp-server/test/e2e/__tests__/ocamllsp-merlinCallCompatible.ts
xvw: in https://github.com/xvw/ocaml-lsp/blob/tunneling-merlin-lsp-p2-use-merlin-lib/ocaml-lsp-server/test/e2e-new/merlin_call_compatible.ml#L116, the last test of the test suite, I randomly face to the following error: `"Lev_fiber.Io: already closed"`. Did I miss something?

```
@@ -124,4 +124,31 @@ let%expect_test "errors: warning is disabled" =
   Helpers.test source request;
   [%expect
     {|
-    { "resultAsSexp": false, "result": "{\"class\":\"return\",\"value\":[]}" } |}]
+    (* CR expect_test_collector: This test expectation appears to contain a backtrace.
+       This is strongly discouraged as backtraces are fragile.
+       Please change this test to not include a backtrace. *)
+
+    { "resultAsSexp": false, "result": "{\"class\":\"return\",\"value\":[]}" }
+    detached: /-----------------------------------------------------------------------
+    | Internal error: Uncaught exception.
+    | ("Lev_fiber.Io: already closed",
+    | { source =
+    |     "Raised by primitive operation at Lev_fiber.Io.callstack.(fun) in file \"submodules/lev/lev-fiber/src/lev_fiber.ml\", line 590, characters 31-58\n\
+    |      Called from Lev_fiber.Io.create in file \"submodules/lev/lev-fiber/src/lev_fiber.ml\", line 743, characters 17-29\n\
+    |      Called from Lev_fiber.Io.Lazy_fiber.force.(fun) in file \"submodules/lev/lev-fiber/src/lev_fiber.ml\", line 1047, characters 23-27\n\
+    |      Called from Fiber__Scheduler.exec in file \"fiber/src/scheduler.ml\", line 73, characters 8-11\n\
+    |      Called from Fiber__Scheduler.start in file \"fiber/src/scheduler.ml\", line 216, characters 2-42\n\
+    |      Called from Fiber.run.(fun) in file \"fiber/src/fiber.ml\" (inlined), line 17, characters 28-47\n\
+    |      Called from Lev_fiber.run in file \"submodules/lev/lev-fiber/src/lev_fiber.ml\", line 1264, characters 10-59\n\
+    |      Called from Ocaml_lsp_server.run in file \"ocaml-lsp-server/src/ocaml_lsp_server.ml\", lines 987-989, characters 2-51\n\
+    |      Called from Stdune__Exn_with_backtrace.try_with in file \"otherlibs/stdune/src/exn_with_backtrace.ml\", line 9, characters 8-12\n\
+    |      Called from Dune__exe__Main in file \"ocaml-lsp-server/bin/main.ml\", lines 41-42, characters 6-72\n\
+    |      "
+    | })
+    | Raised at Stdune__Code_error.raise in file "otherlibs/stdune/src/code_error.ml", line 10, characters 30-62
+    | Called from Lev_fiber.Io.with_ in file "submodules/lev/lev-fiber/src/lev_fiber.ml", line 983, characters 10-62
+    | Called from Fiber__Scheduler.exec in file "fiber/src/scheduler.ml", line 73, characters 8-11
+    | Re-raised at Stdune__Exn.raise_with_backtrace in file "otherlibs/stdune/src/exn.ml" (inlined), line 38, characters 27-56
+    | Called from Stdune__Exn_with_backtrace.reraise in file "otherlibs/stdune/src/exn_with_backtrace.ml", line 20, characters 33-71
+    | Called from Fiber__Scheduler.exec in file "fiber/src/scheduler.ml", line 73, characters 8-11
```

rgrinberg: Usually this happens when you're being sent diagnostics (likely) or code actions (less likely) and you are closing the client before they arrive.

Would be good to see what is being sent in this line:

> Called from Ocaml_lsp_server.run in file \"ocaml-lsp-server/src/ocaml_lsp_server.ml\", lines 987-989, characters 2-51\n\

But regardless, the fix is just to wait for what is being sent to arrive before terminating the test.

I wonder if we should fix ocamllsp to just silently shut itself down on such errors.

xvw:
```ocaml
```

xvw: @rgrinberg 

> Would be good to see what is being sent in this line:

```ocaml
Lev_fiber.run ~sigpipe:`Ignore (fun () ->
      let* input, output = stream_of_channel channel in
      start (Lsp_fiber.Fiber_io.make input output))
  |> Lev_fiber.Error.ok_exn
```

> But regardless, the fix is just to wait for what is being sent to arrive before terminating the test.

I was inspired by other e2e-test suite, did I miss something about properly wait at the end of the test? Thanks in advance!

xvw:
I took the opportunity to rebase on `master` and clean up the history a bit.

ocaml-lsp-server/test/e2e/__tests__/ocamllsp-merlinCallCompatible.ts
xvw: Currently https://github.com/ocaml/ocaml-lsp/actions/runs/9548374342/job/26316523866?pr=1265 is failing because of that error.

PR #1277, opened 2024-05-16
respect completion prefix when completing optional arguments

awilliambauer:
For completions on labels that the LSP gets from merlin, take into account whether the prefix being completed starts with `~` or `?`.

Merlin appears to always returns the full set of labeled/optional arguments when asked for completions. The LSP forwards these on, leaving it to the client to filter them. The  problem is that merlin supplies the optional argument completions with a leading `?` regardless of the prefix the user typed. This causes the client to either filter them out (VSCode) or offer completions that might flip an optional argument being passed with `~` to one passed with `?` (Emacs). 

This PR changes the label completions that start with `?` to start with `~` when the prefix being completed starts with `~`. It also adds a test case and does a bit of refactoring of test helper functions.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,6 +43,11 @@
   toggled on, allows display of sytax documentation on hover tooltips. Can be
   controlled via environment variables and by GUI for VS code. (#1218)
 
+- For completions on labels that the LSP gets from merlin, take into account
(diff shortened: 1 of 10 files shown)


rgrinberg:
There were some async leftovers I had to tweak. LGTM now.

Issue #1207, opened 2023-10-29
Text_document has bugs when performing updates and getting positions


faldor20:
The Text_document module has some weird bugs:
These expect tests show the issue:
```ocaml
let%expect_test "replace second line first line is \\n" =
  let range = tuple_range (1, 2) (1, 2) in
  let doc = make_document (Uri.of_path "foo.ml") ~text:"\nfoo\nbar\nbaz\n" in
  let edit = TextEdit.create ~newText:"change" ~range in
  let newDoc = Text_document.apply_text_document_edits doc [ edit ] in
  newDoc |> Text_document.text |> String.escaped |> print_endline;
  [%expect {|
    result: \nfochangeo\nbar\nbaz\n |}]

let%expect_test "get position after change" =
  let range = tuple_range (1, 2) (1, 2) in
  let doc = make_document (Uri.of_path "foo.ml") ~text:"\nfoo\nbar\nbaz\n" in
  let edit = TextDocumentContentChangeEvent.create ~text:"change" ~range () in
  let newDoc = Text_document.apply_content_changes doc [ edit ] in
  let pos = Text_document.absolute_position newDoc range.start in
  newDoc |> Text_document.text |> String.escaped |> print_endline;
  printf "pos: %d\n" pos;
  [%expect {|
    pos: 3 |}]
```
Test one demonstrates how making the change causes text to become duplicated 
Test two demonstrates how after making a change the absolute position comes out messed up.
I noticed that in my testing the end_ position when running `absolute_range` seemed unaffected, maybe that helps.

I discovered these issues while trying to use the lsp as a library in another project, they were very very confusing to debug :sweat_smile: 

I've noticed that sometimes it seems ocmallsp kind of gets confused and has to be restarted, I suspect this is a contributor to that issue as it could cause the editor and language-service to go out of sync.

I wonder if we could set up a property based test for the text_Document that run's the same edit's on a string using a simpler method than the zipper, then just feed in a huge sample of random positions and change text and see what happens. Just a thought.

faldor20:
Another possibility is to have the lsp check the on disk file, when it gets a file-saved notification to see if the current editor file has diverged from the file on disk, if so it could freak out and also load the on-disk file

rgrinberg:
I don't understand the issue with the first test. As far as I see, you're inserting the string `change` at the 3rd position on the second line and the output demonstrates that's exactly what happens.

faldor20:
[...]

PR #1278, opened 2024-05-17
Fix abs_pos computing in `String_zipper.drop_until`

emilienlemaire:
This PR should fix a problem that occurs when editing a text document with `apply_text_document_edits`, where some text could be duplicated, as shown in #1207.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,6 +48,8 @@
   completions that start with `?` to start with `~` when the prefix being
   completed starts with `~`. (#1277)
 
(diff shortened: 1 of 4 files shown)


rgrinberg:
Thanks for the fix. In fact, this PR is two independent fixes and I would like tests for each individual fix separately. The abs_pos fix is in fact not covered by the tests, as you can see here: https://github.com/ocaml/ocaml-lsp/pull/1280

rgrinberg:
Thanks!

PR #1283, opened 2024-05-17
Add new destruct-line code action

awilliambauer:
Merlin-destruct is made more useful by adding a new LSP code-action called `destruct-line` which does some pre- and post-processing surrounding a call to Merlin-destruct.

This new code-action can be invoked on a line where the user has typed `match...` or `match...with`, or within an incomplete match statement on a line with an existing case. When invoked, the code-action will fill in the remaining cases of the match. This means that in VSCode, if you have one of the following lines, you can hit `C-.` from anywhere on the line to achieve the corresponding expansion.

Supposing `x` is a `bool`, then the line
```ocaml
match x
```
expands to
```ocaml
match x with
| false -> _
| true -> _
```

Supposing `get_card` returns a `card_value` variant, then the line
```ocaml
match get_card (int_of_string s) with
```
expands to
```ocaml
match get_card (int_of_string s) with
| Ace -> _
| King -> _
| Queen -> _
| Jack -> _
| Number _ -> _
```

Or supposing you've already filled in
```ocaml
match get_card (int_of_string s) with
| King -> "Face Card"
| Queen -> "Face Card"
```
then `destruct-line` on either the `King` or `Queen` line will produce
```ocaml
match get_card (int_of_string s) with
| King -> "Face Card"
| Queen -> "Face Card"
| Ace -> _
| Jack -> _
| Number _ -> _
```

In addition, the code action detects a sub-case of the `| x -> y` form, where the cursor is on an underscore within `x`.

The following expansions result from repeated applications of `destruct-line`:

```ocaml
let zip (type a b) (xs : a list) (ys : b list) : (a * b) list =
  match xs, ys
```
After code action is invoked anywhere on the match line:
```ocaml
let zip (type a b) (xs : a list) (ys : b list) : (a * b) list =
  match (xs, ys) with
  | (_, _) -> _
```
After code action is invoked on the first underscore:
```ocaml
let zip (type a b) (xs : a list) (ys : b list) : (a * b) list =
  match (xs, ys) with
  | ([], _) -> _
  | (_::_, _) -> _
```
After code action is invoked on the first underscore:
```ocaml
let zip (type a b) (xs : a list) (ys : b list) : (a * b) list =
  match (xs, ys) with
  | ([], []) -> _
  | ([], _::_) -> _
  | (_::_, _) -> _
```
After code action is invoked on the second-to-last underscore:
```ocaml
let zip (type a b) (xs : a list) (ys : b list) : (a * b) list =
  match (xs, ys) with
  | ([], []) -> _
  | ([], _::_) -> _
  | (_::_, []) -> _
  | (_::_, _::_) -> _
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,9 @@
 
   This now includes PPXs like `let%expect_test` or `let%bench` in the outline.
 
+- Introduce a `destruct-line` code action. This is an improved version of the
+  old `destruct` code action. (#1283)
+
 ## Fixes
 
 - Detect document kind by looking at merlin's `suffixes` config.
--- a/dune-project
+++ b/dune-project
@@ -45,6 +45,7 @@ possible and does not make any assumptions about IO.
  (description "An LSP server for OCaml.")
  (depends
   yojson
+  base
   (re (>= 1.5.0))
   (ppx_yojson_conv_lib (>= "v0.14"))
   (dune-rpc (>= 3.4.0))
--- a/flake.nix
+++ b/flake.nix
@@ -96,6 +96,9 @@
                 yojson
                 ppx_yojson_conv_lib
                 merlin-lib
+                re2
+                base
+                core
               ];
               propagatedBuildInputs = [ ];
               doCheck = false;
--- a/flake.nix
+++ b/flake.nix
@@ -96,9 +96,7 @@
                 yojson
                 ppx_yojson_conv_lib
                 merlin-lib
-                re2
                 base
-                core
               ];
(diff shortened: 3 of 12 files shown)


rgrinberg:
LGTM, but I'd like to avoid Re2. Are you in favor of replacing it with ocaml-re?

rgrinberg:
Alright, I managed to get rid of `re2` in favor of `re`. `core` was unused so I just got rid of it. `base` can stay and we should feel free to use it to replace `stdune`.

PR #1289, opened 2024-05-20
Improve the infer-interface code action and add an update-signature code action

awilliambauer:
This PR improves the `insert-inferred-interface` code-action by restricting the set of signatures inserted to only the names not already present in the interface file. Previously, this code action was only really useful on a completely empty .mli file, since all signaures from the corresponding .ml file were inserted, including duplicates of signatures already present in the .mli.

Repeated applications of this code-action will now more appropriately be no-ops.

This PR also adds an `update-signatures` code action that updates the signatures for selected elements of the interface based on types inferred from the corresponding implementation.

The code action works by checking for an implementation file corresponding to the current interface and querying Merlin for type analysis on both. It then extracts items from the interface that overlap with the range the user has selected and attempts to identify matching items from the inferred interface of the implementation. For any matching items, it asks Merlin to print signatures for the old and new types, and if those strings differ, it produces a text edit for the interface to make it match the implementation.

# Testing
I added a new expect-test that confirms the inserted signatures exclude ones already present in the interface. The existing expect-test that assumed an empty .mli is unaffected.

New expect-tests verify the intended behavior of `update-signatures` in the following situations:
- new arguments added to a function
- arguments dropped from a function
- type of a function argument changed
- multiple functions selected; some changed some not
- within a module exposed in the mli: a function added and cases of a variant changed

This can be slow (~2s) on very large .mli files (~5k lines), but I've confirmed that the pre-existing version of `insert-inferred-interface` took a similarly long time.


rgrinberg:
LGTM after a couple of changes:

1. removed `ppx_let` in favor of the syntax that comes with OCaml. It's not as good as `ppx_let`, but for the simple use cases here it's enough and it's nice to not require users to install ppx for lsp

2. remove all the base references to `Import`. Eventually, we should just write `include Base` in `Import` and replace `Stdune` completely, but that's a bit too much for this PR.

ocaml-lsp-server/src/inference.ml
```diff
@@ -66,25 +107,193 @@ let open_document_from_file (state : State.t) uri =
...
+  | Typedtree.Tsig_type _
+  | Typedtree.Tsig_typesubst _
+  | Typedtree.Tsig_typext _
+  | Typedtree.Tsig_exception _
+  | Typedtree.Tsig_recmodule _
+  | Typedtree.Tsig_open _
+  | Typedtree.Tsig_include _
+  | Typedtree.Tsig_class _
+  | Typedtree.Tsig_class_type _
+  | Typedtree.Tsig_attribute _ -> None
```
voodoos, line 152: If we want to relax `ocaml-lsp`'s dependency on the `Typedtree` on the long term we should really put such utilities in `merlin-lib` instead. 

Also, I am curious about the reason why classes' ids are not needed for the update-signature action ?

PR #1290, opened 2024-05-21
Disable inlay hints under ghost location

awilliambauer:
Inlay hints are both noisy and useless when applied to generated code. Examples include lots of unnecessary hints on the fields of a record definition that derives things like `sexp` or `fields`, or confusing annotations on format strings using `!`. 

All the generated code exists under typedtree nodes with ghost locations. This PR changes the inlay hint logic to stop iterating over the typedtree when it sees an expression or a structure item with a ghost location.

### DIFF
--- a/ocaml-lsp-server/src/inlay_hints.ml
+++ b/ocaml-lsp-server/src/inlay_hints.ml
@@ -56,6 +56,8 @@ let hint_binding_iter ?(hint_let_bindings = false)
       | Texp_match (expr, cases, _) ->
         iter.expr iter expr;
         List.iter cases ~f:(case hint_pattern_variables iter)
+      (* Stop iterating when we see a ghost location to avoid annotating generated code *)
+      | _ when e.exp_loc.loc_ghost -> ()
       | _ -> I.default_iterator.expr iter e
   in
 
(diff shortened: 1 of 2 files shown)


xvw (approved):

rgrinberg:
Somehow this drops all the inlay hints in all of our tests. Those shouldn't be ghost locs.

ocaml-lsp-server/src/inlay_hints.ml
```diff
@@ -56,6 +56,8 @@ let hint_binding_iter ?(hint_let_bindings = false)
       | Texp_match (expr, cases, _) ->
         iter.expr iter expr;
         List.iter cases ~f:(case hint_pattern_variables iter)
+      (* Stop iterating when we see a ghost location to avoid annotating generated code *)
+      | _ when e.exp_loc.loc_ghost -> ()
```
xvw, line 60: Shouldn't the case be above the Tstr_match and continue iteration with the default iterator?

awilliambauer: What makes you say that? Continuing with the default iterator will create inlay hints on the generated code.

awilliambauer: That seems to be what's happening. When I build within Jane Street, the tests work fine (presumably because the document is created with non-ghost locations), but building with external libraries causes this issue where the test document has ghost locations. I'm trying to track down the source of the difference.

rgrinberg:
Okay, I disabled this fix in the tests. LGTM

PR #1296, opened 2024-06-04
Fix parens around function types in signature help

awilliambauer:
Fix the algorithm that decides when to wrap a signature hint with parens.

Testing
-------
Ran the LSP locally and confirmed this correctly surrounds function argument types with parens.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -62,6 +62,8 @@
 
 - Stop generating inlay hints on generated code (#1290)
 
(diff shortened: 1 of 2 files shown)


voodoos (approved):
Looks good to me.

(link to the upstreaming PR: https://github.com/ocaml/merlin/pull/1720)

voodoos:
Could you add a changelog entry ?

PR #1302, opened 2024-06-09
Add `CONTRIBUTING.md`

xvw:
The aim is to make it easier for outsiders to contribute. Here is a first version of a contribution document. At first I'd added examples (ie: how to add a custom query, what Fiber is and how to use it etc.) but I've decided to remove them to focus the review on ‘this base’. As I'm not a native English speaker, I'm open to any rewording. (And of course to any modifications, in general). 

Thanks in advance.


CONTRIBUTING.md
```diff
@@ -0,0 +1,121 @@
...
+
+Apart from that, the project tries to apply implicit conventions, at the
+decision of the maintainers. At the same time, it tries to follow certain naming
+conventions:
+
+- use of `t` manifest types in modules, when it makes sense;
+- conversion functions respecting the naming scheme: `t_of_xxx` or `xxx_of_t`.
+  For example, when converting to or from data described by
+  [Yojson](https://ocaml.org/p/yojson/latest), use `t_of_yojson` and
+  `yojson_of_t`.
```
rgrinberg, line 56: Note that this convention is only followed so that it plays well with `ppx_yojson_conv`. The more customary naming scheme would be `to_json` and `of_json` - but that wouldn't be usable by `[@@deriving yojson]` unfortunately.

CONTRIBUTING.md
```diff
@@ -0,0 +1,121 @@
...
+In most cases, it is likely that the contributions will focus solely on this
+project.
+
+##### Warning
+
+For historical reasons, but also for development convenience, `ocaml-lsp-server`
+should not build logic based on `Parsetree` (which changes from version to
+version and migration logic is not provided by `lsp`, nor `ocaml-lsp-server` but
+by [Merlin](https://github.com/ocaml/merlin)). If a command, or a constant,
+relies on the `Parsetree`, it is **strongly** recommended to build a command in
```
rgrinberg, line 94: I think you meant the typed tree rather than the parse tree. The parse tree is already essentially a public interface as it's used by hundreds of ppx's. Ocamllsp isn't any special here.

Also, I would weaken the language a little. It's certainly fine to experiment with things, and the downstream projects that have asked to maintain this discipline don't always follow this rule themselves. I'd say it's just recommended to be done with care and marked as unstable. In particular, it should be fairly easy to get rid of the feature if it proves intrusive or hard to maintain.

CONTRIBUTING.md
```diff
@@ -0,0 +1,121 @@
...
+protocol. The code lives mainly in the following directories: [lsp/](lsp/) and
+[lsp-fiber/](lsp-fiber/).
+
+##### Warning
+
+The set of types forming the LSP protocol API is generated automatically by a
+[preprocessor](lsp/bin) based on the [protocol
+specification](https://microsoft.github.io/language-server-protocol/overviews/lsp/overview/).
+The pair of [types.ml](lsp/src/types.ml) and [types.mli](lsp/src/types.mli)
+files must therefore never be modified manually.
```
rgrinberg, line 113: They can actually be modified manually. Only the parts inside the cinaps comment marks cannot be modified.

CONTRIBUTING.md
```diff
@@ -0,0 +1,121 @@
...
+### Styleguides
+
+- **OCaml**: a large part of the code base is written in OCaml and the project
+  is configured to work with
+  [ocamlformat](https://ocaml.org/p/ocamlformat/latest) (version defined in the
+  [`.ocamlformat` file](.ocamlformat)).
+
+- **TypeScript**: TypeScript is used to describe certain end-to-end tests
+  (abbreviated as `e2e`) and the project uses the
+  [prettier](https://prettier.io/) formatter.
```
rgrinberg, line 46: Would be good to mention that these tests are deprecated and we do not allow extending them anymore. Gradually we'll rewrite them all to OCaml.

rgrinberg (approved):
LGTM. I added some minor clarifications. Adding more information about the various other components sounds good to me.

CONTRIBUTING.md
xvw: You are right ! Thinking about "dealing with parse tree, avoiding typed tree", i mixed the two.

xvw:
Thanks for the review @rgrinberg, I try to take it account the last commit! WDYT?

PR #1304, opened 2024-06-10
Type enclosing query

xvw:
This PR is based on [1265](https://github.com/ocaml/ocaml-lsp/pull/1265) and add a `type enclosing` custom request.

### DIFF
--- a/ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
+++ b/ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
@@ -25,11 +25,10 @@ There is no client capability relative to this request.
 
   ```json
   {
-    "textDocument": TextDocumentIdentifier,
-    "position": Position,
+    "uri": TextDocumentIdentifier,
+    "at": (Position | Range),
     "index": uinteger,
     "verbosity?": uinteger,
-    "rangeEnd?": Position
   }
   ```
 
@@ -37,9 +36,10 @@ There is no client capability relative to this request.
     the types lazily: normally, Merlin would return the signature of all enclosing
     modules, which can be very expensive.
   - `verbosity` determines the number of expansions of aliases in answers.
-  - `rangeEnd` an optional end position. If provided, only enclosings that contain the
-    range `[super.position; end[` will be included in the answer.
-
+  - `at` :
+    - if a `Position` is given, it will returns all enclosing around the position
+    - if a `Range` is given, only enclosings that contain the range
+    `[range.start; range.end[` will be included in the answer
 
 
 ## Response
--- a/ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
+++ b/ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
@@ -39,7 +39,7 @@ There is no client capability relative to this request.
   - `at` :
     - if a `Position` is given, it will returns all enclosing around the position
     - if a `Range` is given, only enclosings that contain the range
-    `[range.start; range.end[` will be included in the answer
+    `[range.start; range.end]` will be included in the answer
 
 
 ## Response
--- a/ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
+++ b/ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
@@ -1,0 +1,59 @@
+# Type Enclosing Request
+
+## Description
+
+Merlin has a concept of `type enclosing` that gets the type of ident under the
+cursor. It will highlight the ident and display its type. You can climb the
+typed-tree and display the type of bigger expressions surrounding the cursor. In
+order to keep the request stateless, the manipulation related to growing or
+shrinking enclosings is delegated to the client. This request allows to request
+type enclosing under the cursor and then its surrounding enclosings.
+
+## Client capability
+
+There is no client capability relative to this request.
+
+## Server capability
+
+- property name: `handleTypeEnclosing`
+- property type: `boolean`
+
+## Request
+
+- method: `ocamllsp/typeEnclosing`
+- params:
+
+  ```json
+  {
+    "textDocument": TextDocumentIdentifier,
+    "position": Position,
+    "index": uinteger,
+    "verbosity?": uinteger,
+    "rangeEnd?": Position
+  }
+  ```
+
+  - `index` can be used to print only one type information. This is useful to query
+    the types lazily: normally, Merlin would return the signature of all enclosing
+    modules, which can be very expensive.
+  - `verbosity` determines the number of expansions of aliases in answers.
+  - `rangeEnd` an optional end position. If provided, only enclosings that contain the
+    range `[super.position; end[` will be included in the answer.
+
+
+
+## Response
+
+```json
+{
+  "enclosings": Range[],
+  "index": uinteger,
+  "type": string
+}
+```
+
+- `enclosings`: The surrounding enclosings
+- `index` The index of the provided type result: the index corresponds to a
+  zero-indexed enclosing in the `enclosings`' array. It is the same value as the
+  one provided in this request's `TypeEnclosingParams`
+- `type`: The type of the enclosing `enclosings[index]` as a raw string
--- a/ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
+++ b/ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
@@ -7,28 +7,27 @@ let meth = "ocamllsp/typeEnclosing"
 
 module Request_params = struct
   type t =
-    { text_document_position : TextDocumentPositionParams.t
+    { text_document : TextDocumentIdentifier.t
+    ; at : [ `Range of Range.t | `Position of Position.t ]
     ; index : int
-    ; range_end : Position.t option
     ; verbosity : int
     }
 
-  let yojson_of_t { text_document_position; index; range_end; verbosity } =
-    match TextDocumentPositionParams.yojson_of_t text_document_position with
+  let yojson_of_at = function
+    | `Range r -> Range.yojson_of_t r
+    | `Position p -> Position.yojson_of_t p
+
+  let yojson_of_t { text_document; index; at; verbosity } =
+    match TextDocumentIdentifier.yojson_of_t text_document with
     | `Assoc assoc ->
       let index = ("index", `Int index) in
-      let range_end =
-        ( "rangeEnd"
-        , match range_end with
-          | Some x -> Position.yojson_of_t x
-          | None -> `Null )
-      in
+      let range_end = ("at", yojson_of_at at) in
       let verbosity = ("verbosity", `Int verbosity) in
       `Assoc (index :: range_end :: verbosity :: assoc)
     | _ -> (* unreachable  *) assert false
 
-  let create ?range_end ?(verbosity = 0) ~text_document_position ~index () =
-    { text_document_position; index; range_end; verbosity }
+  let create ?(verbosity = 0) ~text_document ~at ~index () =
+    { text_document; index; at; verbosity }
 
   let json_error json =
     Json.error "invalid Req_type_enclosing.Request_params" json
@@ -49,23 +48,23 @@ module Request_params = struct
          the we ask for a verbosity level set to 0. *)
       0
 
-  let range_end_of_yojson params =
-    match List.assoc_opt "rangeEnd" params with
-    | Some range_end -> Some (Position.t_of_yojson range_end)
+  let at_of_yojson json params =
+    match List.assoc_opt "at" params with
+    | Some at -> (
+      try `Position (Position.t_of_yojson at)
+      with _ -> `Range (Range.t_of_yojson at))
     | _ ->
-      (* If the parameter is incorrectly formatted or missing, it is assumed that
-         the we do not provide rangeEnd parameter. *)
-      None
+      (* If the parameter is incorrectly formatted or missing, we refuse to build
+         the parameter, [at] is mandatory. *)
+      json_error json
 
   let t_of_yojson = function
     | `Assoc params as json ->
       let verbosity = verbosity_of_yojson params in
-      let range_end = range_end_of_yojson params in
+      let at = at_of_yojson json params in
       let index = index_of_yojson json params in
-      let text_document_position =
-        TextDocumentPositionParams.t_of_yojson json
-      in
-      { index; range_end; verbosity; text_document_position }
+      let text_document = TextDocumentIdentifier.t_of_yojson json in
+      { index; at; verbosity; text_document }
     | json -> json_error json
 end
 
@@ -104,19 +103,9 @@ let make_enclosing_command position index =
 let get_first_enclosing_index range_end enclosings =
   List.find_mapi enclosings ~f:(fun i (loc, _, _) ->
       let range = Range.of_loc loc in
-      match
-        ( Position.compare range_end range.start
-        , Position.compare range_end range.end_ )
-      with
-      | Ordering.(Gt, Gt)
-      | Ordering.(Eq, Lt)
-      | Ordering.(Gt, Eq)
-      | Ordering.(Eq, Eq)
-      | Ordering.(Gt, Lt) -> Some i
-      | Ordering.Lt, Ordering.Lt
-      | Ordering.Lt, Ordering.Eq
-      | Ordering.Lt, Ordering.Gt
-      | Ordering.Eq, Ordering.Gt -> None)
+      match Position.compare range_end range.end_ with
+      | Ordering.Lt | Ordering.Eq -> Some i
+      | Ordering.Gt -> None)
 
 let dispatch_command pipeline command first_index index =
   let rec aux i acc = function
@@ -161,10 +150,8 @@ let dispatch_without_range_end pipeline position index =
   let command = make_enclosing_command position index in
   dispatch_command pipeline command 0 index
 
-let dispatch_type_enclosing
-    (text_document_position : TextDocumentPositionParams.t) index range_end
-    pipeline =
-  let position = Position.logical text_document_position.position in
+let dispatch_type_enclosing position index range_end pipeline =
+  let position = Position.logical position in
   let result =
     match range_end with
     | None -> dispatch_without_range_end pipeline position index
@@ -181,11 +168,15 @@ let dispatch_type_enclosing
 let on_request ~params state =
   Fiber.of_thunk (fun () ->
       let params = (Option.value ~default:(`Assoc []) params :> Json.t) in
-      let Request_params.
-            { index; verbosity; text_document_position; range_end; _ } =
+      let Request_params.{ index; verbosity; text_document; at } =
         Request_params.t_of_yojson params
       in
-      let uri = text_document_position.textDocument.uri in
+      let position, range_end =
+        match at with
+        | `Position p -> (p, None)
+        | `Range r -> (r.start, Some r.end_)
+      in
+      let uri = text_document.uri in
       let verbosity = Mconfig.Verbosity.Lvl verbosity in
       with_pipeline state uri verbosity
-      @@ dispatch_type_enclosing text_document_position index range_end)
+      @@ dispatch_type_enclosing position index range_end)
--- a/ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
+++ b/ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
@@ -103,8 +103,7 @@ let make_enclosing_command position index =
   Query_protocol.Type_enclosing (None, position, Some index)
 
 let get_first_enclosing_index range_end enclosings =
-  List.find_mapi
-    ~f:(fun i (loc, _, _) ->
+  List.find_mapi enclosings ~f:(fun i (loc, _, _) ->
       let range = Range.of_loc loc in
       match
         ( Position.compare range_end range.start
@@ -119,7 +118,6 @@ let get_first_enclosing_index range_end enclosings =
       | Ordering.Lt, Ordering.Eq
       | Ordering.Lt, Ordering.Gt
       | Ordering.Eq, Ordering.Gt -> None)
-    enclosings
 
 let dispatch_command pipeline command first_index index =
   let rec aux i acc = function
--- a/ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
+++ b/ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
@@ -1,0 +1,191 @@
+open Import
+
+let capability = ("handleTypeEnclosing", `Bool true)
+
+let meth = "ocamllsp/typeEnclosing"
+
+type params =
+  { text_document_position : Lsp.Types.TextDocumentPositionParams.t
+  ; index : int
+  ; range_end : Position.t option
+  ; verbosity : int
+  }
+
+let expected_params =
+  `Assoc
+    [ ("index", `String "uinteger")
+    ; ("rangeEnd?", `String "<Position>")
+    ; ("verbosity?", `String "uinteger")
+    ; ("position", `String "<Position>")
+    ; ("textDocument", `String "<TextDocumentIdentifier>")
+    ]
+
+let index_of_yojson params =
+  match List.assoc_opt "index" params with
+  | Some (`Int index) -> Some index
+  | _ ->
+    (* If the parameter is incorrectly formatted or missing, we refuse to build
+       the parameter, [index] is mandatory. *)
+    None
+
+let verbosity_of_yojson params =
+  match List.assoc_opt "verbosity" params with
+  | Some (`Int verbosity) -> verbosity
+  | _ ->
+    (* If the parameter is incorrectly formatted or missing, it is assumed that
+       the we ask for a verbosity level set to 0. *)
+    0
+
+let range_end_of_yojson params =
+  match List.assoc_opt "rangeEnd" params with
+  | Some range_end -> (
+    try Some (Position.t_of_yojson range_end) with _ -> None)
+  | _ ->
+    (* If the parameter is incorrectly formatted or missing, it is assumed that
+       the we do not provide rangeEnd parameter. *)
+    None
+
+let raise_invalid_params ?data ~message () =
+  let open Jsonrpc.Response.Error in
+  raise @@ make ?data ~code:Code.InvalidParams ~message ()
+
+let of_yojson = function
+  | `Assoc params as full_params ->
+    let verbosity = verbosity_of_yojson params in
+    let range_end = range_end_of_yojson params in
+    let open Option.O in
+    let* index = index_of_yojson params in
+    let text_document_position =
+      Lsp.Types.TextDocumentPositionParams.t_of_yojson full_params
+    in
+    Some { index; range_end; verbosity; text_document_position }
+  | _ -> None
+
+let of_yojson_exn = function
+  | None -> raise_invalid_params ~message:"Expected params but received none" ()
+  | Some params -> (
+    match of_yojson params with
+    | Some params -> params
+    | None ->
+      let data =
+        `Assoc
+          [ ("expectedParams", expected_params)
+          ; ("receivedParams", (params :> Json.t))
+          ]
+      in
+      raise_invalid_params ~data ~message:"Unexpected params format" ())
+
+let render_result index result =
+  let typ, enclosings =
+    match result with
+    | None -> ("<no information>", [])
+    | Some (typ, enclosings) ->
+      (typ, List.map ~f:Lsp.Types.Range.yojson_of_t enclosings)
+  in
+  `Assoc
+    [ ("index", `Int index)
+    ; ("enclosings", `List enclosings)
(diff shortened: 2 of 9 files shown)


voodoos:
I don't remember: is there any reason why this is based on the 5.2 branch instead of master ?

xvw:
> I don't remember: is there any reason why this is based on the 5.2 branch instead of master ?

I am currently rebased on master!

ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
```diff
@@ -0,0 +1,51 @@
+# Type Enclosing Request
+
+## Description
+
+Merlin has a concept of `type enclosing` that gets the type of ident under the
+cursor. It will highlight the ident and display its type. You can climb the
+typed-tree and display the type of bigger expressions surrounding the cursor.
+Since _LSP_ is stateless, the manipulation related to growing or shrinking
```
voodoos, line 8: Even if the interface we propose for `type-enclosing` is somehow stateless, LSP itself is not a stateless protocol.

ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
```diff
@@ -0,0 +1,51 @@
...
+enclosings is delayed to the client. This request allows to request type
```
voodoos, line 9:
```suggestion
enclosings is delegated to the client. This request allows to request type
```

ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
```diff
@@ -0,0 +1,51 @@
...
+enclosing under the cursor with a support of parent enclosings.
```
voodoos, line 10:
```suggestion
enclosing under the cursor and then its surrounding enclosings.
```

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,224 @@
+open Import
+
+let capability = ("handleTypeEnclosing", `Bool true)
+
+let meth = "ocamllsp/typeEnclosing"
+
+type params =
+  { text_document_position : Lsp.Types.TextDocumentPositionParams.t
+  ; work_done_progress : Lsp.Types.WorkDoneProgressParams.t
```
voodoos, line 9: We do nothing with this param. Most standard LSP queries accept it to report progress, but other custom queries does not appear to handle it. If we don't actually make use of it maybe we should simply remove it ?

ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
```diff
@@ -0,0 +1,51 @@
...
+
+  ```json
+  {
+    "textDocument": TextDocumentIdentifier,
+    "position": Position,
+    "workDoneToken?": ProgressToken,
+    "index": uinteger,
+    "verbosity": uinteger,
+    "rangeEnd?": Position
+  }
```
voodoos, line 34: I think I forgot the `uri` of the document in my specification, we should respect this interface:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams

Not sure about the utility of the `workDoneToken` however.

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,224 @@
...
+    match index |> List.nth result with
+    | None -> "<no information>"
+    | Some (_, typ) -> typ
+  in
+  let enclosings =
+    List.map
+      ~f:(fun (loc, typ) ->
+        `Assoc
+          [ ("start", Position.yojson_of_t loc.Range.start)
+          ; ("end_", Position.yojson_of_t loc.Range.end_)
```
voodoos, line 194: I don't think we need the `_` in the field name.

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,224 @@
...
+  let current_typ =
+    match index |> List.nth result with
+    | None -> "<no information>"
+    | Some (_, typ) -> typ
+  in
+  let enclosings =
+    List.map
+      ~f:(fun (loc, typ) ->
+        `Assoc
+          [ ("start", Position.yojson_of_t loc.Range.start)
```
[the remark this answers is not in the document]
xvw: There is an additional field (`type`) in _our_ range.

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
xvw: It is the name in `Types`.

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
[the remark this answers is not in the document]
xvw: We can for something like that: 

```json
enclosing = {
  "range": <Range.t>,
  "type" : string
}
```

ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
xvw: I think it is already handle by `textDocumentPosition` (that has a `texDocument` field, which is a `textDocumentIdentifier`, that has a `uri` field :) )

voodoos: Oh sorry ! I didn't see that :-)

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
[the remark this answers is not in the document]
voodoos: Ah yes, I didn't see that. But then we should remove that type field, but we don't want to return the type of every enclosing, only the one at the given index.

In the design document the return type is:
```
/**
 * The result of a typeEnclosing request.
 */
export interface TypeEnclosing {
    /**
     * The surrounding enclosings.
     */
    enclosings: [Range](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range)[];

    /**
     * The index of the provided type result: the index corresponds to a zero-indexed 
     * enclosing in the `enclosings`' array. It is the same value as the one provided 
     * in this request's TypeEnclosingParams.
     */
    index:  [uinteger](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#uinteger);

    /**
     * The type of the enclosing `enclosings[index]` as a raw string.
     */
    type: string;
}
```

ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
```diff
@@ -0,0 +1,50 @@
...
+
+- method: `ocamllsp/typeEnclosing`
+- params:
+
+  ```json
+  {
+    "textDocument": TextDocumentIdentifier,
+    "position": Position,
+    "index": uinteger,
+    "verbosity": uinteger,
```
voodoos, line 31:
```suggestion
    "verbosity?": uinteger,
```

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,162 @@
...
+    | (loc, _, _) :: xs ->
+      let range = Range.of_loc loc in
+      if overlap_with_range_end range range_end then Some i else aux (succ i) xs
+    | _ -> None
+  in
+  aux 0 enclosings
+
+let dispatch_type_enclosing text_document_position index range_end pipeline =
+  let position = get_logical_position text_document_position in
+  let dummy_command = make_enclosing_command position (-1) in
```
voodoos, line 133: We should comment somewhere about this hack.

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,162 @@
...
+          ]
+      in
+      raise_invalid_params ~data ~message:"Unexpected params format" ())
+
+let overlap_with_range_end range = function
+  | None -> true
+  | Some position ->
+    let lend = range.Range.end_ in
+    if lend.line = position.Position.line then
+      lend.character > position.character
```
voodoos, line 83:
```suggestion
      lend.character >= position.character
```
I think ?

ocaml-lsp-server/test/e2e-new/type_enclosing.ml
```diff
@@ -0,0 +1,241 @@
...
+          "end": { "character": 20, "line": 23 },
+          "start": { "character": 13, "line": 23 }
+        },
+        {
+          "end": { "character": 20, "line": 23 },
+          "start": { "character": 8, "line": 23 }
+        }
+      ],
+      "type": "sig type t = A of int val x : int end"
+    } |}]
```
voodoos, line 241: We should also put a focus on the `range_end` cases that rely on tricky position comparisons and index shifting :-)

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,162 @@
...
+let dispatch_type_enclosing text_document_position index range_end pipeline =
+  let position = get_logical_position text_document_position in
+  let dummy_command = make_enclosing_command position (-1) in
+  let enclosings = Query_commands.dispatch pipeline dummy_command in
+  Option.bind
+    (get_first_enclosing_index range_end enclosings)
+    ~f:(fun first_index ->
+      let command = make_enclosing_command position (first_index + index) in
+      match Query_commands.dispatch pipeline command with
+      | (_, `String typ, _) :: enclosings ->
```
voodoos, line 140: What if we were not asking for the type of the first enclosing but one of the following ?
You could fold to both reconstruct the enclosing list (by simple concatenation and rev) and return the type of the required enclosing.

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
voodoos: And problably only rely on it when renage_end <> None

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
xvw: Not really necessary because it is already guarded by the first condition: 

```ocaml
let overlap_with_range_end range = function
  | None -> true
  | Some position ->
    let lend = range.Range.end_ in
    if lend.line = position.Position.line (* this one *) then
      lend.character > position.character
    else lend.line > position.line
```

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
xvw: Not sure to understand. When you say "not the type of the first enclosing", are you saying "giving an index <> 0"?

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
xvw: Fixed on the last commit

ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
```diff
@@ -0,0 +1,50 @@
...
+- method: `ocamllsp/typeEnclosing`
+- params:
+
+  ```json
+  {
+    "textDocument": TextDocumentIdentifier,
+    "position": Position,
+    "index": uinteger,
+    "verbosity?": uinteger,
+    "rangeEnd?": Position
```
voodoos, line 32: I think we should describe the roles of the parameters here, especially rangeend (we could simply reuse the comments in the design doc if they are up-to-date)

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,185 @@
...
+      ~config:(config_with_given_verbosity config verbosity)
+      merlin
+      with_pipeline
+
+let make_enclosing_command position index =
+  Query_protocol.Type_enclosing (None, position, Some index)
+
+let get_logical_position tdp =
+  let p = tdp.Lsp.Types.TextDocumentPositionParams.position in
+  Position.logical p
```
voodoos, line 117: Nitpick: not sure about the need for this function. It is far from it's unique callsite and does nothing more than calling a function that already bear a quite similar name.

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
voodoos: Are you sure about that ? I don't see it... they could be on the same line and the same character ?

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,185 @@
...
+          [ ("expectedParams", expected_params)
+          ; ("receivedParams", (params :> Json.t))
+          ]
+      in
+      raise_invalid_params ~data ~message:"Unexpected params format" ())
+
+let overlap_with_range_end range position =
+  let lend = range.Range.end_ in
+  if lend.line = position.Position.line then lend.character > position.character
+  else lend.line > position.line
```
voodoos, line 81: Not sure about that, but maybe we could simply use `Position.compare_inclusion` to check that the range_end pos is inside the enclosing range. This has the main advantage of not duplicating this kind of error-prone reasoning...

ocaml-lsp-server/test/e2e-new/type_enclosing.ml
```diff
@@ -0,0 +1,431 @@
...
+  Client.request client req
+
+let print_type_enclosing result =
+  result |> Yojson.Safe.pretty_to_string ~std:false |> print_endline
+
+let%expect_test "type enclosing on simple example" =
+  let source = {|let x = string_of_int 2002|} in
+  let request client =
+    let open Fiber.O in
+    let position = Position.create ~line:0 ~character:22 in
```
voodoos, line 32: I am trying to read the tests: these are positions 0-based right ? For the line it looks like it is the case, and for the column (`character`) number too ?

ocaml-lsp-server/test/e2e-new/type_enclosing.ml
```diff
@@ -0,0 +1,431 @@
...
+    {|
+    {
+      "index": 0,
+      "enclosings": [
+        {
+          "end": { "character": 26, "line": 0 },
+          "start": { "character": 8, "line": 0 }
+        }
+      ],
+      "type": "int"
```
voodoos, line 48: This is wrong isn't it ? 

Looks like the enclosing of `string_of_int 2002` which should have type string...
We expect something like this: 

```    {
      "index": 0,
      "enclosings": [
        {
          "end": { "character": 26, "line": 0 },
          "start": { "character": 22, "line": 0 }
        },
        {
          "end": { "character": 26, "line": 0 },
          "start": { "character": 8, "line": 0 }
        }
      ],
      "type": "int"
}
```

And if the required index was 1:
```    {
      "index": 1,
      "enclosings": [
        {
          "end": { "character": 26, "line": 0 },
          "start": { "character": 22, "line": 0 }
        },
        {
          "end": { "character": 26, "line": 0 },
          "start": { "character": 8, "line": 0 }
        }
      ],
      "type": "string"
}
```

voodoos (commented):
It's starting to look good ! 
There is still some weirdness in the tests however...

(Also, but that's personal taste: I think utility functions are sometimes a bit too far from there call point. Especially when they are only used once, sometime you could just inline them or move them in a deeper scope. But that's really nitpicking, don't feel obliged to change anything.)

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
xvw: oh you're right. I was think that your where speaking about `line`!

rgrinberg (changes requested):

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,191 @@
...
+  raise @@ make ?data ~code:Code.InvalidParams ~message ()
+
+let of_yojson = function
+  | `Assoc params as full_params ->
+    let verbosity = verbosity_of_yojson params in
+    let range_end = range_end_of_yojson params in
+    let open Option.O in
+    let* index = index_of_yojson params in
+    let text_document_position =
+      Lsp.Types.TextDocumentPositionParams.t_of_yojson full_params
```
rgrinberg, line 59: `Lsp.Types` is redundant here

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,191 @@
...
+    match of_yojson params with
+    | Some params -> params
+    | None ->
+      let data =
+        `Assoc
+          [ ("expectedParams", expected_params)
+          ; ("receivedParams", (params :> Json.t))
+          ]
+      in
+      raise_invalid_params ~data ~message:"Unexpected params format" ())
```
rgrinberg, line 76: All of this error handling is done at the wrong layer. If we get a json encoding error, there should be a single place that converts into the proper error code. So instead of collecting the option type and worrying about a pretty error message, doing the easy thing and just raising the same exception that ppx_yojson_conv raises is the correct thing to do. If we want to improve error messages for json, we should just do it in one place for all requests and rely on the exceptions being informative.

rgrinberg (changes requested):

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,191 @@
...
+      match
+        ( Position.compare range_end range.start
+        , Position.compare range_end range.end_ )
+      with
+      | Ordering.(Gt, Gt)
+      | Ordering.(Eq, Lt)
+      | Ordering.(Gt, Eq)
+      | Ordering.(Eq, Eq)
+      | Ordering.(Gt, Lt) -> Some i
+      | _ -> aux (succ i) xs)
```
rgrinberg, line 123: Can you expand this pattern?

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,191 @@
...
+    Document.Merlin.with_configurable_pipeline_exn
+      ~config:(config_with_given_verbosity config verbosity)
+      merlin
+      with_pipeline
+
+let make_enclosing_command position index =
+  Query_protocol.Type_enclosing (None, position, Some index)
+
+let get_first_enclosing_index range_end enclosings =
+  let rec aux i = function
```
rgrinberg, line 111: I think this is just `List.findi`?

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,191 @@
...
+      Some
+        ( typ
+        , List.map
+            ~f:(fun (loc, _, _) -> Range.of_loc loc)
+            (List.rev_append acc enclosings) )
+    | curr :: enclosings when i >= first_index ->
+      aux (succ i) (curr :: acc) enclosings
+    | _ :: enclosings -> aux (succ i) acc enclosings
+    | [] -> None
+  in
```
rgrinberg, line 140: I think you just need `List.drop first_index` here?

ocaml-lsp-server/test/e2e-new/type_enclosing.ml
```diff
@@ -0,0 +1,491 @@
...
+    let req =
+      Lsp.Client_request.UnknownRequest
+        { meth = "ocamllsp/typeEnclosing"; params }
+    in
+    Client.request client req
+
+  let print_type_enclosing result =
+    result |> Yojson.Safe.pretty_to_string ~std:false |> print_endline
+
+  let test ?(verbosity = 0) ?(index = 0) ?range_end ~line ~character source =
```
rgrinberg, line 29: I'd say just make these required

ocaml-lsp-server/test/e2e-new/type_enclosing.ml
```diff
@@ -0,0 +1,491 @@
...
+        | Some x -> [ ("rangeEnd", Position.yojson_of_t x) ])
+    in
+    let params = Some (Jsonrpc.Structured.t_of_yojson params) in
+    let req =
+      Lsp.Client_request.UnknownRequest
+        { meth = "ocamllsp/typeEnclosing"; params }
+    in
+    Client.request client req
+
+  let print_type_enclosing result =
```
rgrinberg, line 26: You can move this to e2e-new/test.ml since its general purpose

ocaml-lsp-server/test/e2e-new/type_enclosing.ml
```diff
@@ -0,0 +1,491 @@
+open Test.Import
+
+module Util = struct
+  let call_type_enclosing ?(verbosity = 0) ?range_end client position index =
+    let uri = DocumentUri.of_path "test.ml" in
+    let text_document = TextDocumentIdentifier.create ~uri in
+    let params =
+      `Assoc
+        ([ ("textDocument", TextDocumentIdentifier.yojson_of_t text_document)
```
rgrinberg, line 9: How about just exporting the type from `req_type_enclosing`? Writing out the json here will just make it harder to sync the test with the actual code.

ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
rgrinberg: Instead of this scheme, I'd recommend the following:

1. Either take a `range` parameter
2. Or a `position` parameter

That way you can stick to standard field names and users can use the convenient `Range.t` type to work with your request.

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,191 @@
...
+      dispatch_command pipeline command first_index real_index)
+
+let dispatch_without_range_end pipeline position index =
+  let command = make_enclosing_command position index in
+  dispatch_command pipeline command 0 index
+
+let dispatch_type_enclosing text_document_position index range_end pipeline =
+  let position =
+    Position.logical
+      text_document_position.Lsp.Types.TextDocumentPositionParams.position
```
rgrinberg, line 173: You can annotation `text_document_position` instead.

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,191 @@
...
+      raise_invalid_params ~data ~message:"Unexpected params format" ())
+
+let render_result index result =
+  let typ, enclosings =
+    match result with
+    | None -> ("<no information>", [])
+    | Some (typ, enclosings) ->
+      (typ, List.map ~f:Lsp.Types.Range.yojson_of_t enclosings)
+  in
+  `Assoc
```
rgrinberg, line 85: This needs a defined response type.

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
```diff
@@ -0,0 +1,191 @@
...
+    let range_end = range_end_of_yojson params in
+    let open Option.O in
+    let* index = index_of_yojson params in
+    let text_document_position =
+      Lsp.Types.TextDocumentPositionParams.t_of_yojson full_params
+    in
+    Some { index; range_end; verbosity; text_document_position }
+  | _ -> None
+
+let of_yojson_exn = function
```
rgrinberg, line 64: Given that this function decodes `params`:

1. Its name should be `params_of_yojson`
2. Its signature should be `Yojson.t -> params`

We must match the conventions of the code generator

rgrinberg:
Sorry for the long winded review. Regarding all the issues with json, I realize that these are issues that are essentially present in all of our custom requests. Indeed, we've added a lot of custom code doing things that are standard and all of our standard requests have a quirky behavior. I see that at least somebody tried to unify things a little bit by introducing `Custom_request`. We should expand that example. To briefly summarize the rules for custom requests:

1. The overarching rule is that hand written encoders/decoders should be interchangeable with generated ones.
2. This means that we need to follow all the conventions of generated encoders/decoders
3. It's good to have explicit request/response types for every single custom request (They're going to be very useful once we need to version things)
4. The conversion from json decoding errors to the appropriate jsonrpc error should be done once and for all somewhere in jsonrpc-fiber or lsp-fiber. E.g. we handle this for all regular lsp requests in one swoop:
```ocaml
      match In_request.of_jsonrpc req with
      | Error message ->
        let code = Jsonrpc.Response.Error.Code.InvalidParams in
        let error = Jsonrpc.Response.Error.make ~code ~message () in
        Fiber.return
          (Jsonrpc_fiber.Reply.now (Jsonrpc.Response.error req.id error), state)
```

We should strive for the same for custom requests.

In any case, it's not fair to hold your code to a higher standard to all previous code. So for now you don't need to address to make everything uniform, but I would like at least point 3 addressed with request/response module and their own `t_of_yojson` and `yojson_of_t` functions (of whatever type).

xvw:
@rgrinberg 

Thank you very much for your detailed review!

> In any case, it's not fair to hold your code to a higher standard to all previous code.

From my point of view, it's not worth merging code that we explicitly know can be improved, so I'm going to try, as best I can, to take into account all the proposed points. Then I'll try to do a pass on all the custom requests that have already been implemented.

Thank you very much, I'll take the liberty of pinging you when this PR is ready (Monday).

rgrinberg:
Sure, I'm certainly against improving things across the board. Feel free to regularize all custom requests. I'll give some background why the situation is like this in the first place. In the beginning, I was very much against custom requests and even more so against investing a lot of work into them. I thought they would require a lot duplicate work on the client side and the LSP protocol would evolve to make them all obsolete anyhow. The LSP protocol didn't really evolve much and the use cases for custom requests steadily grew. The end result is that we ended up with a lot of ad-hoc extensions.

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
xvw: `List.drop` does not exists in `Stdune`.

ocaml-lsp-server/test/e2e-new/type_enclosing.ml
xvw: Is that really general purpose? FMPOV it is really related to testing type enclosing queries.

ocaml-lsp-server/src/custom_requests/req_type_enclosing.ml
rgrinberg: Feel free to try base. And if it doesn't exist, writing it into import.ml is an option too.

xvw: Fixed into https://github.com/ocaml/ocaml-lsp/pull/1304/commits/fc297726568d120fd3a927266fdee1c90b4a8409

ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
xvw: You are right. Fixed in https://github.com/ocaml/ocaml-lsp/pull/1304/commits/89efc55328efe1b68879bbc5b388d913ad63ff0b

ocaml-lsp-server/test/e2e-new/type_enclosing.ml
rgrinberg: It appears in `e2e-new/completion.ml`, `e2e-new/hover_extended.ml`,  `e2e-new/merlin_call_compatible.ml`.  So I'd say it's general purpose. You gave it a very specific name here, but it could just be called `pretty_print_json`

rgrinberg (approved):

ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md
```diff
@@ -0,0 +1,59 @@
...
+  ```
+
+  - `index` can be used to print only one type information. This is useful to query
+    the types lazily: normally, Merlin would return the signature of all enclosing
+    modules, which can be very expensive.
+  - `verbosity` determines the number of expansions of aliases in answers.
+  - `at` :
+    - if a `Position` is given, it will returns all enclosing around the position
+    - if a `Range` is given, only enclosings that contain the range
+    `[range.start; range.end[` will be included in the answer
```
voodoos, line 42:
```suggestion
    `[range.start; range.end]` will be included in the answer
```

Not sure why I excluded the end of the range when writtng the spec.

voodoos (approved):
I think this is ready :-) 

We should open an issue in merlin about the stability of enclosings when index is 0 or 1.

voodoos:
The CI failures appear to be unrelated ? Maybe we should bump the setup-ocaml action like in https://github.com/ocaml/merlin/pull/1791 ?

PR #1318, opened 2024-06-21
hover: use `***` to print dividers.

voodoos:
vscode renderer treats the section before a `---` divider as a title (it is a [setext heading](https://spec.commonmark.org/0.31.2/#setext-heading-underline), see [example 59](https://spec.commonmark.org/0.31.2/#example-59)). This stayed undetected before the syntax doc PR because only code was printed before the divider.

Current behavior with `---`:
image

We can use stars to get a delimiter without a heading:
image

cc @PizieDust


PizieDust (approved):
Nice catch. Thank you.

PR #1331, opened 2024-07-07
chore: revert incorrect changes to Makefile

rgrinberg:
@xvw your original is still flaky even with your workaround. Your workaround also makes the tests way slower for everyone. Let's keep the test suite in a healthy state.


xvw:
Thanks a lot for the fix!

PR #1336, opened 2024-07-09
Documentation custom query

### LINKED ISSUE #1319: Documentation query
In the objective of restoring classic Merlin workflows in text-based editors we need a custom query to get a symbol's documentation.

Proposed interface:

Client capability:
```typescript
export interface GetDocClientCapabilities {
	/**
	 * Client supports the follow content formats if the content
	 * property refers to a `literal of type MarkupContent`.
	 * The order describes the preferred format of the client.
	 */
	contentFormat?: MarkupKind[];
}
```

Query:
```typescript
export interface GetDocParams extends TextDocumentPositionParams
{
    /**
     * An optional identifier. If provided, documentation for this ident 
     * is looked up from the environment at the given position. Else the 
     * server will look for the documentation of the identifier under 
     * the cursor if used.
     */
    identifier?:string;

    /**
     * Optionally override the result's format. 
     */
    contentFormat?:MarkupKind;
}
```

Response:
```typescript
result: GetDoc | null
export interface GetDoc {
    /**
     * The documentation
     */
    doc: MarkupContent;

}
```

We could also have the format choice be made a part of t capability, but I am not sure this is worth the added complexity ?

A response with `null` result is returned of the identifier doesn't have documentation.
An error is returned if the identifier is invalid.

PizieDust:
closes #1319

Add `documentation` custom query.


ocaml-lsp-server/src/custom_requests/req_documentation.ml
```diff
@@ -0,0 +1,120 @@
...
+        ]
+      | None -> [])
+
+  let t_of_yojson json =
+    let open Yojson.Safe.Util in
+    let contentFormat =
+      json |> member "contentFormat" |> to_list
+      |> List.map ~f:(fun format -> MarkupKind.t_of_yojson format)
+      |> Option.some
+    in
```
xvw, line 29: Since it is an option, we should probably match if the list is empty in order to provide a `None`.
btw @voodoos I am not sure about the level of indirection, using an option on list leads to strange behaviour, ie: `Some []`, why not directly going for a list where `[] = always plaintext` ?

ocaml-lsp-server/src/custom_requests/req_documentation.ml
```diff
@@ -0,0 +1,120 @@
...
+              | `No_documentation | `Invalid_context | `Not_found _ -> `Null
+              | `Builtin value
+              | `File_not_found value
+              | `Found value
+              | `Not_in_env value ->
+                let markup_content =
+                  match contentFormat with
+                  | Some format -> MarkupContent.create ~kind:format ~value
+                  | None ->
+                    MarkupContent.create ~kind:MarkupKind.PlainText ~value
```
xvw, line 116: I think that you can write this: 

```ocaml
let kind = match contentFormat with 
  | Some format -> format
  | None -> MarkupKind.PlainText
in MarkupContent.create ~kind ~value
```

ocaml-lsp-server/src/custom_requests/req_documentation.mli
```diff
@@ -0,0 +1,38 @@
+val meth : string
+
+val capability : string * [> `Bool of bool ]
+
+module GetDocClientCapabilities : sig
+  type t = { contentFormat : Import.MarkupKind.t list option }
+
+  val yojson_of_t :
+    t -> [> `Assoc of (string * [> `List of Yojson.Safe.t list ]) list ]
```
xvw, line 9: You can probably just use `Json.t`

ocaml-lsp-server/src/custom_requests/req_documentation.mli
```diff
@@ -0,0 +1,38 @@
...
+
+  val t_of_yojson : Yojson.Safe.t -> t
```
xvw, line 11: Not sure that this is necessary.

ocaml-lsp-server/src/custom_requests/req_documentation.mli
```diff
@@ -0,0 +1,38 @@
+val meth : string
+
+val capability : string * [> `Bool of bool ]
+
```
xvw, line 4: Are you sure that it is necessary to exports every submodules? I think that only `Request_params` is mandatory.

ocaml-lsp-server/test/e2e-new/documentation.ml
```diff
@@ -0,0 +1,100 @@
...
+  let identifier = "List" in
+  Util.test ~line ~character ~identifier source;
+  [%expect
+    {|
+    {
+      "doc": {
+        "kind": "plaintext",
+        "value": "List operations.\n\n   Some functions are flagged as not tail-recursive.  A tail-recursive\n   function uses constant stack space, while a non-tail-recursive function\n   uses stack space proportional to the length of its list argument, which\n   can be a problem with very long lists.  When the function takes several\n   list arguments, an approximate formula giving stack usage (in some\n   unspecified constant unit) is shown in parentheses.\n\n   The above considerations can usually be ignored if your lists are not\n   longer than about 10000 elements.\n\n   The labeled version of this module can be used as described in the\n   {!StdLabels} module."
+      }
+    } |}]
```
xvw, line 100: Can you add a test where you Shadow `List` for your own implementation (and documentation)?

xvw (changes requested):
Very nice work! 
I think that we can improve a little bit the code organization exposing only a `Request_params` module (for the input arguments) and exposing a type (`t`) that describes the output of the query.
I think that we can also only "produces parameters" into the body of the `Fiber.thrunk` and using a dedicated function to fetching the documentation.

ocaml-lsp-server/test/e2e-new/documentation.ml
```diff
@@ -0,0 +1,100 @@
...
+  let call_documentation ~position ?(identifier = None) ?(contentFormat = None)
+      client =
+    let uri = DocumentUri.of_path "test.ml" in
+    let text_document =
+      TextDocumentIdentifier.yojson_of_t @@ TextDocumentIdentifier.create ~uri
+    in
+    let position = Position.yojson_of_t position in
+    let params =
+      `Assoc
+        (("position", position)
```
rgrinberg, line 13: Instead of writing the encoding for the second time, you could just export what you already have and re-use that.

ocaml-lsp-server/src/custom_requests/req_documentation.mli
```diff
@@ -0,0 +1,38 @@
...
+  val yojson_of_t :
+    t -> [> `Assoc of (string * [> `List of Yojson.Safe.t list ]) list ]
+
+  val t_of_yojson : Yojson.Safe.t -> t
+end
+
+module GetDocParams : sig
+  type t =
+    { text_document : Import.TextDocumentIdentifier.t
```
rgrinberg, line 16: `open Import` should be done in the beginning of your module and not qualified explicitly

ocaml-lsp-server/test/e2e-new/documentation.ml
xvw: It still not possible because custom request are still not exposed. But it is in my current PR.

rgrinberg: I don't really see the need to wait though. Making the custom request available to the test is less code than copy pasting all this encoding stuff

ocaml-lsp-server/src/custom_requests/req_documentation.ml
PizieDust: > why not directly going for a list where [] = always plaintext
This makes sense. The next commit will implement it this way and we can eval if it's better

ocaml-lsp-server/test/e2e-new/documentation.ml
PizieDust: @xvw When I shadow `List` and run this doc query, for some reason it returns the docs for original List module. Could this be a scope issue?

xvw: With, for example, the `Base`'s one ?

ocaml-lsp-server/src/custom_requests/req_documentation.ml
```diff
@@ -0,0 +1,81 @@
...
+      | `Builtin value
+      | `File_not_found value
+      | `Found value
+      | `Not_in_env value ->
+        let markup_content =
+          match contentFormat with
+          | Some format -> GetDoc.create ~kind:format ~value
+          | None -> GetDoc.create ~kind:MarkupKind.PlainText ~value
+        in
+        GetDoc.yojson_of_t { doc = markup_content })
```
xvw, line 67: I suggest to add a `type t = GetDoc.t` (and exposes-it in the interface)

ocaml-lsp-server/test/e2e-new/documentation.ml
PizieDust: I updated the test. I added a todo comment to investigate on the Merlin side.

ocaml-lsp-server/docs/ocamllsp/documentation-spec.md
```diff
@@ -0,0 +1,40 @@
+# Documentation Request
+
+## Description
+
+Merlin has a command `document` that gets `odoc` documentation for symbols based on a cursor position or an optional identifier. This request allows documentation to be gotten using a classic Merlin workflow.
```
xvw, line 5: I think it would be better to make the definition of `merlin` agnostic and explain concretely what the query does. (By explaining that it allows you to bypass `Hover`, for example, to bring up documentation information).

ocaml-lsp-server/docs/ocamllsp/documentation-spec.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Client capability
+
+```js
+export interface GetDocClientCapabilities {
+	contentFormat: MarkupKind[];
+}
+```
+- `contentFormat`: Client supports the following content formats if the content property refers to a `literal of type MarkupContent`. The order describes the preferred format of the client.
```
xvw, line 14: We should probably add the capability relative to the request: `handleDocumentation`

ocaml-lsp-server/docs/ocamllsp/documentation-spec.md
```diff
@@ -0,0 +1,40 @@
...
+- `position`: The position of the cursor.
+- `identifier`: An optional identifier. If provided, documentation for this ident is looked up from the environment at the given position. Else the server will look for the documentation of the identifier under the cursor.
+- `contentFormat`: Optionally override the result's format. Could be `Plaintext` or `Markdown`.
+
+## Response
+
+```js
+result: GetDoc | null
+export interface GetDoc {
+    doc: MarkupContent;
```
xvw, line 34: WDYT about returning, also, the `MarkupKind` in the result?

ocaml-lsp-server/src/custom_requests/req_documentation.ml
```diff
@@ -0,0 +1,83 @@
+open Import
+module TextDocumentPositionParams = Lsp.Types.TextDocumentPositionParams
+
+let meth = "ocamllsp/documentation"
```
xvw, line 4: Are we sure about the method name? Why not `getDocumentation` (cc @pitag-ha)

ocaml-lsp-server/src/custom_requests/req_documentation.ml
```diff
@@ -0,0 +1,83 @@
...
+      | `No_documentation | `Invalid_context | `Not_found _ -> `Null
+      | `Builtin value
+      | `File_not_found value
+      | `Found value
+      | `Not_in_env value ->
+        let markup_content =
+          match contentFormat with
+          | Some format -> GetDoc.create ~kind:format ~value
+          | None -> GetDoc.create ~kind:MarkupKind.PlainText ~value
+        in
```
xvw, line 68: I think that https://github.com/ocaml/ocaml-lsp/pull/1336#discussion_r1670738570 should not be resolved no?

ocaml-lsp-server/docs/ocamllsp/documentation-spec.md
PizieDust: Do you mean in the `.md` file or the results record itself?

PizieDust: If we return `markupKind`, it becomes:
```json
 [%expect {|
    {
      "doc": { "kind": "markdown", "value": "This is another comment" },
      "markupKind": "markdown"
    } |}]
```
which is duplicated

xvw: Oh! You are right. There is already the `kind` in `MarkupContent`!

ocaml-lsp-server/src/custom_requests/req_documentation.ml
```diff
@@ -0,0 +1,112 @@
+open Import
+module TextDocumentPositionParams = Lsp.Types.TextDocumentPositionParams
+
+let meth = "ocamllsp/getDocumentation"
+let capability = "handleDocumentation", `Bool true
```
xvw, line 5:
```diff
- let capability = "handleDocumentation", `Bool true
+ let capability = "handleGetDocumentation", `Bool true
```

ocaml-lsp-server/src/custom_requests/custom_request.ml
```diff
@@ -5,3 +5,4 @@ module Switch_impl_intf = Req_switch_impl_intf
 module Typed_holes = Req_typed_holes
 module Type_enclosing = Req_type_enclosing
 module Wrapping_ast_node = Req_wrapping_ast_node
+module Documentation = Req_documentation
```
xvw, line 8: Should be `Get_documentation` (and `req_documentation.ml(i)` should be `req_get_documentation`)

ocaml-lsp-server/test/e2e-new/documentation.ml
```diff
@@ -0,0 +1,173 @@
...
+        :: [ ("textDocument", text_document) ])
+    in
+    let params = Some (Jsonrpc.Structured.t_of_yojson params) in
+    let req =
+      Lsp.Client_request.UnknownRequest
+        { meth = "ocamllsp/documentation"; params }
+    in
+    Client.request client req
+
+  let print_documentation result =
```
xvw, line 20: Non you can use `Test.print_result`

ocaml-lsp-server/test/e2e-new/documentation.ml
```diff
@@ -18,7 +18,7 @@ module Util = struct
   ;;
 
   let print_documentation result =
-    result |> Yojson.Safe.pretty_to_string ~std:false |> print_endline
```
xvw: I think, instead of using `print_documentation result` you can just use `Test.print_result` in every test

xvw (approved):
Approved! Thanks a lot @PizieDust!

PizieDust:
> Approved! Thanks a lot @PizieDust!

Thanks for the reviews

xvw:
@rgrinberg Do you want to make one last pass?

rgrinberg:
The implementation looks fine. @awilliambauer do you want to take a look at the docs? Is this something that you will be able to use?

awilliambauer:
This looks reasonable to me, and useful for providing documentation-focus functionality.

ocaml-lsp-server/docs/ocamllsp/getDocumentation-spec.md
```diff
@@ -0,0 +1,48 @@
...
+```js
+export interface GetDocParams extends TextDocumentPositionParams
+{
+    identifier?: string;
+    contentFormat?:MarkupKind;
+}
+```
+- method : `ocamllsp/getDocumentation`
+- params :
+    - `position`: The position of the cursor.
```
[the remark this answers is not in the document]
PizieDust: Yes, since the interface itself extends `TextDocumentPositionParams` which contains both the Document Uri and the Position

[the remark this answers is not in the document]
xvw: We can probably provide a link to the TextDocumentPositionParams and remove the description of `position` (since `textDocument`, the `TextDocumentIdentifier` is not mentionned in the documentation).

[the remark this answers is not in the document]
xvw: Not sure that it is right. I think that you should explain that the param should be a TextDocumentPositionParams (with a link to the specification) with additonal ones, enumerated.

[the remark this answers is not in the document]
PizieDust: @xvw I just pushed a commit with some documentation changes. Can you check if it's more structured?

ocaml-lsp-server/src/custom_requests/req_get_documentation.ml
```diff
@@ -0,0 +1,112 @@
...
+
+  let yojson_of_t { doc } = `Assoc [ "doc", MarkupContent.yojson_of_t doc ]
+
+  let t_of_yojson json =
+    let open Yojson.Safe.Util in
+    let doc = json |> member "doc" |> MarkupContent.t_of_yojson in
+    { doc }
+  ;;
+
+  let create ~kind ~value = MarkupContent.create ~kind ~value
```
voodoos, line 68: I think we are missing something here: when  the requested markup content is markdown, we should translate the documentation syntax to markown in the same way that it is done for the on_hover using `Doc_to_md.translate`.

This is how it is done in [hover_req.ml](https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/src/hover_req.ml#L230-L232)

```ocaml
Option.map doc ~f:(fun doc ->
  match Doc_to_md.translate doc with
  | Raw d -> d
  | Markdown d -> d)
```

voodoos (commented):
This PR looks good, thanks @PizieDust 

I think we do miss one thing however: when markdown is requested we should use the `Doc_to_md` module to translate ocamldoc syntax to markown.

ocaml-lsp-server/src/custom_requests/req_get_documentation.ml
PizieDust: Thank you for catching that.

voodoos (approved):
Thanks for the change ! I think this is ready then ?

PR #1346, opened 2024-07-16
Global pass on Custom_request

xvw:
The patch proposes a global pass on Custom Requests by trying to impose the `module Request_params` structure, which proposes a creation function (and projection in Yojson), and a `type t` at the module toplevel (with a conversion function from Yojson). 

- What's more, parameter deserialization relies on Yojson's API to get close to what derivation would give.
- The `Custom_request` module is exposed by `Ocaml_lsp_server` for use from, for example, tests (and also from a JavaScript client, ie: vscode).
- Some requests have been modified very little to avoid introducing breaking-changes.

I'm obviously open to feedback!

### DIFF
--- a/ocaml-lsp-server/src/custom_requests/custom_request.ml
+++ b/ocaml-lsp-server/src/custom_requests/custom_request.ml
@@ -1,30 +1,7 @@
-open Import
-
-type 't req_params_spec =
-  { params_schema : Jsonrpc.Structured.t
-  ; of_jsonrpc_params : Jsonrpc.Structured.t -> 't option
-  }
-
-let of_jsonrpc_params_exn spec params =
-  let raise_invalid_params ?data ~message () =
-    Jsonrpc.Response.Error.raise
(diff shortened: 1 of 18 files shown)


rgrinberg:
LGTM. Some further suggestions for the future:

1. Move the declarations (request/response type definition + encoders) of the custom requests to a separate library and make a package.

2. Introduce a type `('req, 'resp) custom_request` to have a uniform API

rgrinberg (approved):

xvw:
> LGTM. Some further suggestions for the future:
> 
>     1. Move the declarations (request/response type definition + encoders) of the custom requests to a separate library and make a package.
> 
>     2. Introduce a type `('req, 'resp) custom_request` to have a uniform API

Yes, it is on my todo-list!
I was also thinking about having a `Custom_request.run : ('req, 'resp) custom_request -> 'req -> 'resp` to avoid relaying on `run Unknown_request` at the user level. 

Thanks a lot for the quick review!

PR #1348, opened 2024-07-19
Construct custom query

xvw:
At the moment, construct is a completion hook, and in some cases accessing it via a one-off query can be useful.

### DIFF
--- a/ocaml-lsp-server/docs/ocamllsp/construct-spec.md
+++ b/ocaml-lsp-server/docs/ocamllsp/construct-spec.md
@@ -1,0 +1,44 @@
+# Construct Request
+
+## Description
+
+Provides commands to browse and fill typed holes (`_`). Such holes sometimes
+appear in the result of other commands like `destruct` and can also be inserted
+manually in the source. The command is already accessible via a completion hook,
+however, in certain situations, invoking `construct` on a hole via a request
+allows more control.
+
+## Client Capability
+
+There is no client capability relative to this request.
+
+## Server capability
+
+- property name: `handleConstruct`
+- property type: `boolean`
+
+## Request
+
+- method: `ocamllsp/typeEnclosing`
+- params:
+
+  ```json
+  {
+    "uri": TextDocumentIdentifier,
+    "position": Position
+    "depth?": uinteger (default value: 0)
+    "withValues?": <"local" | "none">,
+  }
+  ```
+
+## Response
+
+```json
+{
+  "position": Range,
+  "result": string[]
+}
+```
+
+The result contains the range (`position`) to be replaced (describing the hole)
+and the list of possible substitution values (`result`).
--- a/ocaml-lsp-server/docs/ocamllsp/construct-spec.md
+++ b/ocaml-lsp-server/docs/ocamllsp/construct-spec.md
@@ -19,7 +19,7 @@ There is no client capability relative to this request.
 
 ## Request
 
-- method: `ocamllsp/typeEnclosing`
+- method: `ocamllsp/construct`
 - params:
 
   ```json
@@ -30,6 +30,11 @@ There is no client capability relative to this request.
     "withValues?": <"local" | "none">,
   }
   ```
+
+The `depth` parameter allows to recursively construct terms. Note that
+when `depth > 1` partial results of inferior depth will not be
+returned. The `withValues` use values from the environment (`local`)
+or not (`none`), by default, the parameter is fixed to `none`.
 
 ## Response
 
--- a/ocaml-lsp-server/docs/ocamllsp/construct-spec.md
+++ b/ocaml-lsp-server/docs/ocamllsp/construct-spec.md
@@ -2,7 +2,7 @@
 
 ## Description
 
-Provides commands to browse and fill typed holes (`_`). Such holes sometimes
+Provides commands to fill typed holes (`_`). Such holes sometimes
 appear in the result of other commands like `destruct` and can also be inserted
 manually in the source. The command is already accessible via a completion hook,
 however, in certain situations, invoking `construct` on a hole via a request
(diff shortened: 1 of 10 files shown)


rgrinberg:
When you say "can be useful", do you mean you have a concrete user in mind? Perhaps a client where you are planning to use this.

xvw:
> When you say "can be useful", do you mean you have a concrete user in mind? Perhaps a client where you are planning to use this.

Using Lsp through Emacs (that does not rely to much on hover and interactivy on top of mouse) can make the usage of `construct` cumbersome so I'll probably rely on a custom request to use it when I'll move to emacs + lsp (instead of emacs + merlin).

rgrinberg:
I don't doubt the potential. What I'm saying is that until some client actually demonstrates real intention in using your custom request, you're just adding dead code. In the lsp standard itself, there's a requirement to demonstrate at least one client implementation for every additional protocol extension. We don't have to go that far, but how about we at least get some feedback from at least one actual client developer that is going to use this custom request before jump the gun.

xvw:
@rgrinberg 
We expect different clients to use that query:
- The Jane Street internal Emacs and Vim clients (which are not open-source). At JS, they've already enhanced the contruct query by having one code action per construct value as the current design of construct in OCaml LSP is VSCode-specifiv and can't be used by Emacs and Vim. The design of one code action per construct value isn't fully satisfying either, particularly on a long list of construct values (That's why Jane Street prefers to have a custom query). 
- We don't have concrete plans with a timeline yet, but we're certainly planning to have open-source Emacs and Vim extensions for the Emacs and Vim LSP plug-ins to support OCaml LSP specific functionalities. We're planning to use the new construct query in there as well.

rgrinberg:
Okay, so what I'm getting is that the only client developer we need to consider is from JS. So let's have somebody from there have a look at this request to do a sanity check.

rgrinberg:
cc @awilliambauer

ocaml-lsp-server/docs/ocamllsp/construct-spec.md
```diff
@@ -0,0 +1,44 @@
...
+There is no client capability relative to this request.
+
+## Server capability
+
+- property name: `handleConstruct`
+- property type: `boolean`
+
+## Request
+
+- method: `ocamllsp/typeEnclosing`
```
awilliambauer, line 22: I assume typeEnclosing is a typo

awilliambauer (changes requested):
Looks good overall, just need a few tweaks to the documentation.

ocaml-lsp-server/docs/ocamllsp/construct-spec.md
```diff
@@ -0,0 +1,44 @@
...
+
+  ```json
+  {
+    "uri": TextDocumentIdentifier,
+    "position": Position
+    "depth?": uinteger (default value: 0)
+    "withValues?": <"local" | "none">,
+  }
+  ```
+
```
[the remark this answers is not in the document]
xvw: Fixed on https://github.com/ocaml/ocaml-lsp/pull/1348/commits/b2939c252136a00bd02e561a6275dd5f30ee1756

ocaml-lsp-server/docs/ocamllsp/construct-spec.md
xvw: Fixed on https://github.com/ocaml/ocaml-lsp/pull/1348/commits/b2939c252136a00bd02e561a6275dd5f30ee1756

Nice catch, thanks!

ocaml-lsp-server/docs/ocamllsp/construct-spec.md
```diff
@@ -0,0 +1,49 @@
+# Construct Request
+
+## Description
+
+Provides commands to browse and fill typed holes (`_`). Such holes sometimes
```
voodoos, line 5:
```suggestion
Provides commands to fill typed holes (`_`). Such holes sometimes
```

ocaml-lsp-server/docs/ocamllsp/construct-spec.md
```diff
@@ -0,0 +1,49 @@
...
+    "position": Position
+    "depth?": uinteger (default value: 0)
+    "withValues?": <"local" | "none">,
+  }
+  ```
+
+The `depth` parameter allows to recursively construct terms. Note that
+when `depth > 1` partial results of inferior depth will not be
+returned. The `withValues` use values from the environment (`local`)
+or not (`none`), by default, the parameter is fixed to `none`.
```
voodoos, line 37:
```suggestion
returned. The `withValues` parameter enables the use of values from 
the environment (`local`) or not (`none`), It defaults to `none`.
```

ocaml-lsp-server/src/custom_requests/req_construct.ml
```diff
@@ -0,0 +1,111 @@
...
+;;
+
+let with_pipeline state uri f =
+  let doc = Document_store.get state.State.store uri in
+  match Document.kind doc with
+  | `Other -> Fiber.return `Null
+  | `Merlin merlin ->
+    (match Document.Merlin.kind merlin with
+     | Document.Kind.Intf ->
+       (* Construct makes so sense if its called from an interface. *)
```
voodoos, line 87:
```suggestion
       (* Construct makes no sense if its called from an interface. *)
```

voodoos (approved):

ocaml-lsp-server/docs/ocamllsp/construct-spec.md
xvw: Fixed on https://github.com/ocaml/ocaml-lsp/pull/1348/commits/b98322f1f1ed73cc9ad66b907b0c73c64c387448

ocaml-lsp-server/docs/ocamllsp/construct-spec.md
xvw: Fixed on https://github.com/ocaml/ocaml-lsp/pull/1348/commits/b98322f1f1ed73cc9ad66b907b0c73c64c387448

ocaml-lsp-server/src/custom_requests/req_construct.ml
xvw: fixed on https://github.com/ocaml/ocaml-lsp/pull/1348/commits/b98322f1f1ed73cc9ad66b907b0c73c64c387448

PR #1349, opened 2024-07-21
fix: close stderr in [run_in_directory]

rgrinberg:
Otherwise we're going to leak a file descriptor

### DIFF
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -929,6 +929,7 @@ let run_in_directory ~prog ~prog_is_quoted:_ ~args ~cwd ?stdin ?stdout ?stderr (
   in
   Unix.close stdin;
   if should_close_stdout then Unix.close stdout;
+  Option.iter stderr ~f:Unix.close;
   `Finished res
 ;;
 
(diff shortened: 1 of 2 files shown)


ocaml-lsp-server/src/ocaml_lsp_server.ml
```diff
@@ -929,6 +929,7 @@ let run_in_directory ~prog ~prog_is_quoted:_ ~args ~cwd ?stdin ?stdout ?stderr (
   in
   Unix.close stdin;
   if should_close_stdout then Unix.close stdout;
+  Option.iter stderr ~f:Unix.close;
```
voodoos, line 932: Are you trying to fix an issue you observed ? Looking at Merlin's codebase I don't think there is any usage of `run_in_directory` where `stderr` is not `None`.

But yes we should close that file descriptor when stderr it redirected to a file.

voodoos (approved):
Looks good to me. Should I re-released with the fix ? I don't think it was ever happening since `stderr` is always `None`.

ocaml-lsp-server/src/ocaml_lsp_server.ml
rgrinberg: Yes, I never observed this issue but I imagine it could be possible in the future.

rgrinberg:
No need to re-release, but it's a nice fix to have in case future versions of merlin use stderr for something

PR #1361, opened 2024-08-22
Jump Custom Requests

### LINKED ISSUE #1360: Syntactic and Semantic movement shortcuts
# Intro
- Semantic movement is navigation based on the meaning and structure of code, rather than its textual representation. E.g:
    - Jumping to the definition or declaration of a function, type, or variable.
    - Moving to the next or previous function definition.
    - Jumping to matching braces, parentheses, or other paired symbols.
- Syntactic movement involves navigating based on the syntactical structure of the code. This includes moving through the code in a way that respects its syntax rules and formatting. For example:
    - Jumping to the next or previous statement or expression.
    - Moving to the start or end of a block of code.
    - Navigating through different levels of indentation or nesting.

# Limitations of the Standard LSP Protocol
LSP protocol contains four specifications for code navigation:
- Goto Definition (implemented): This usually takes the user to where a symbol (like a variable or function) is defined.
- Goto Declaration (implemented): Similar to "Go to Definition," but often takes the user to where the symbol is first declared.
- Goto Type Definition (implemented): This might take the user to where a type is defined.
- Goto Implementation (not implemented): Resolves the implementation location of a symbol at a given text document position.

# Advanced Navigation from Merlin
Merlin finetunes navigation by understanding the semantics of OCaml code, providing features that go beyond standard navigation:
- Jump to Specific Targets: Merlin can jump to different kinds of constructs like function definitions, let bindings, or specific cases in a match expression. 
- Handling Multiple Matches: OCaml's pattern matching often means that a single pattern can be matched in multiple places. Merlin allows developers to jump to the next or previous match case, which is something standard LSP requests cannot do.
- Module System Navigation: Merlin understands the entire module system, including functor applications and module inclusions. It can navigate between these complex structures.
- Semantic Awareness: Merlin’s navigation considers the semantics of the code, not just the syntax. This means it understands the meaning of the code, not just where symbols are declared or defined.

This functionality in Merlin is provided by the [Jump](https://github.com/ocaml/merlin/blob/main/doc/dev/PROTOCOL.md#jump--target-string--position-position) command.
    
 ## Using Code Actions
Using a CodeAction in combination with LSP's [ShowDocumentRequest](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showDocument), we can be able to move the cursor within the same file to any position.

PizieDust:
closes #1360 
This PR contains an implementation for a custom requests based on Merlin's jump command.

### DIFF
--- a/ocaml-lsp-server/docs/ocamllsp/jump-spec.md
+++ b/ocaml-lsp-server/docs/ocamllsp/jump-spec.md
@@ -1,0 +1,49 @@
+# Jump Request
+
+## Description
+
+This custom request allows Merlin-type code navigation in a source buffer.
+
+## Server capability
+
+- propert name: `handleJump`
+- property type: `boolean`
+
+## Request
+
+```js
+export interface JumpParams extends TextDocumentPositionParams
+{
+    target: string;
+}
+```
+
+- method: `ocamllsp/jump`
+- params:
+    - `TextDocumentIdentifier`: Specifies the document for which the request is sent. It includes a uri property that points to the document.
+    - `Position`: Specifies the position in the document for which the documentation is requested. It includes line and character properties.
+    More details can be found in the [TextDocumentPositionParams - LSP Specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams).
+    - `Target`: A string representing the identifier within the document to search for and jump to.
+
+## Response
+
+```js
+result: Jump | String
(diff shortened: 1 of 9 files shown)


rgrinberg:
Thanks. @awilliambauer could you read the spec if you have some time? Do you plan to do anything with this custom request?

PizieDust:
Using CodeActions has made this PR obsolete. See https://github.com/ocaml/ocaml-lsp/pull/1364

PR #1363, opened 2024-08-23
Fix prefix parsing to start method completion when typing `#`

voodoos:
When completing method access like in `my_object#a|` Merlin expect the "prefix" to be `a` and not `myobject#a` as it would be for accessing a record field. 

This PR introduces a special case in the parsing of prefixes.

Additionally I found the `in` keyword completion a bit too aggressive and removed on of the cases. @ulugbekna do you remember why you made completions with an empty prefix always suggest the `in` keyword ?

### DIFF
--- a/ocaml-lsp-server/src/prefix_parser.ml
+++ b/ocaml-lsp-server/src/prefix_parser.ml
@@ -24,6 +24,15 @@ include struct
          ])
   ;;
 
+  (*  When completing module paths or record fields Merlin expects the
+      beginning of the path and the `.` to be part of the prefix. But when
+      accessing an object's methods, the prefix should not contain the `#`
+      sign. We use a sub-matching group to that effect.
+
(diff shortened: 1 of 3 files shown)


ocaml-lsp-server/src/prefix_parser.ml
```diff
@@ -24,6 +24,15 @@ include struct
          ])
   ;;
 
+  (**  When completing module paths or record fields Merlin expects the
```
rgrinberg, line 27: Probably meant a regular comment here?

rgrinberg (approved):

rgrinberg:
> Additionally I found the in keyword completion a bit too aggressive and removed on of the cases. @ulugbekna do you remember why you made completions with an empty prefix always suggest the in keyword ?

How about this is moved to a separate PR? It should also have its own CHANGES entry I believe

PR #1364, opened 2024-08-28
Merlins Jump in Ocaml-LSP

### LINKED ISSUE #1360: Syntactic and Semantic movement shortcuts

PizieDust:
closes #1360 

# Jump to Target
[Merlin's Jump](https://github.com/ocaml/merlin/blob/main/doc/dev/PROTOCOL.md#jump--target-string--position-position) command allows precise code navigation in a source code buffer. This functionality is particularly useful for developers who wish to streamline their navigation within large codebases by jumping directly to relevant code segments. However, the standard LSP protocol only allows for generalized code navigation (`goto definition`, `goto declaration`, `goto implementation`, `goto type-definition`), which is not very useful when it comes to precise movements.

The Jump command in Merlin allows us to jump to a set of targets:
- `fun`: Jump to a function
- `let`: Jump to a let definition
- `module`: Jump to a module
- `module-type`: Jump to a module type definition
- `match`: Jump to a match construct.
- `match-next-case`: Jump to the next case in the match body.
- `match-prev-case`: Jump to the previous case in the match body.

While this Jump functionality can be implemented in ocaml-lsp using a custom request, see [Closed PR - Jump | Custom Request](https://github.com/ocaml/ocaml-lsp/pull/1361), we sought to find different solutions as using a custom request requires individual client implementations for all the clients we may want to support.

This PR implements Merlin's jump command in ocaml-lsp using a new set of CodeActions.

## Implementation
The implementation of the "Jump to Target" feature uses a `CodeAction`  with a `showDocumentRequest` to move the cursor to the required position.

### 1. Document Detection

When a source file is opened, the server checks if the document is an OCaml file supported by Merlin. If the document is not compatible with Merlin, the "Jump to Target" code action is not provided.

### 2. Client Capability Check

Before generating code actions, the server verifies whether the client supports the `ShowDocument` client capability. If this capability is not supported, the "Jump to Target" feature is disabled.

### 3. Generating Code Actions with Context

For each target (`fun`, `let`, `module`, etc.), the server attempts to locate the corresponding position in the document using Merlin's `Jump` command:

- The server asynchronously sends jump requests for each target, allowing for non-blocking execution.
- The results are collected, and for each successful jump (i.e., where Merlin finds a valid target position), a `CodeAction` is created.
- Each `CodeAction` includes a title, such as "Jump to function", and a command (`ocamllsp/merlin-jump-to-target`) that instructs the client to jump to the specified location in the document.

### 4. Command Execution

When a user selects a code action, the associated command is executed. The server sends a `ShowDocument` request to the client, which asks the client to focus on the URI and range where the target was found.

### 5. Error Handling

The implementation includes error handling to ensure that if a jump request fails (e.g., Merlin does not find the target or the position is invalid), the code action may be omitted.

## Advantages of the Code Action Approach

- **Client-Agnostic**: By implementing this functionality as a `CodeAction`, the solution is broadly compatible with different clients. It avoids the need for custom client implementations that would have been required with a custom LSP request.
- **Asynchronous Execution**: The use of Fibers allows for non-blocking execution of jump requests, ensuring that the server remains responsive even when processing multiple targets in parallel.
- **Precision Navigation**: Users can quickly navigate to specific constructs within their OCaml code, improving productivity and making it easier to work within large and complex codebases.
- **Integration with Existing LSP Features**: This implementation uses existing LSP features, providing a familiar experience for users while extending the capabilities of the `ocaml-lsp-server`.

## Demo (VSCode)

https://github.com/user-attachments/assets/4659ff97-5679-46bd-bdad-6dc9597277b6

## Resources
- [LSP Protocol - ShowDocument](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showDocument)
- [LSP Protocol - ExecuteCommand](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_executeCommand)

cc @voodoos @xvw @pitag-ha


ocaml-lsp-server/src/code_actions/action_jump.ml
```diff
@@ -0,0 +1,96 @@
...
+  =
+  match Document.kind doc with
+  | `Other -> Fiber.return []
+  | `Merlin merlin ->
+    (match available capabilities with
+     | false -> Fiber.return []
+     | true ->
+       let* actions_fiber_list =
+         Fiber.parallel_map targets ~f:(fun target ->
+           let* res = process_jump_request ~merlin ~position:params.range.start ~target in
```
voodoos, line 71: If you use `let+` here you won't need all the calls to `Fiber.return`.

ocaml-lsp-server/src/code_actions/action_jump.ml
```diff
@@ -0,0 +1,96 @@
...
+  (params : CodeActionParams.t)
+  (capabilities : ShowDocumentClientCapabilities.t option)
+  =
+  match Document.kind doc with
+  | `Other -> Fiber.return []
+  | `Merlin merlin ->
+    (match available capabilities with
+     | false -> Fiber.return []
+     | true ->
+       let* actions_fiber_list =
```
voodoos, line 69: This value has type `CodeAction.t option list`, a better name would be `actions_list` or `actions` since it  is actually not a fiber.

ocaml-lsp-server/src/code_actions/action_jump.ml
```diff
@@ -0,0 +1,96 @@
...
+         Fiber.parallel_map targets ~f:(fun target ->
+           let* res = process_jump_request ~merlin ~position:params.range.start ~target in
+           match res with
+           | None -> Fiber.return None
+           | Some pos ->
+             (match Position.of_lexical_position pos with
+              | None -> Fiber.return None
+              | Some position ->
```
voodoos, line 77: These could be replaced by `Option.map`.

Even better, you can open the `Option.O` module and use the `let*` and `let+` operators:
```ocaml
let+ res = process_jump_request ~merlin ~position:params.range.start ~target in
let open Option.O in
let* lexing_pos = res in
let+ position = Position.of_lexical_position lexing_pos in
let uri = Document.uri doc in
let range = { Range.start = position; end_ = position } in
let title = sprintf "Jump to %s" target in
let command =
  let arguments = [ DocumentUri.yojson_of_t uri; Range.yojson_of_t range ] in
  Command.create ~title ~command:command_name ~arguments ()
in
CodeAction.create ~title ~kind:(CodeActionKind.Other "merlin-jump") ~command ())
        ```

File: ocaml-lsp-server/src/code_actions/action_jump.ml [hunk 4]
```diff
@@ -0,0 +1,96 @@
...
+  (capabilities : ShowDocumentClientCapabilities.t option)
+  =
+  match Document.kind doc with
+  | `Other -> Fiber.return []
+  | `Merlin merlin ->
+    (match available capabilities with
+     | false -> Fiber.return []
+     | true ->
+       let* actions_fiber_list =
+         Fiber.parallel_map targets ~f:(fun target ->
```
voodoos, line 70: It's a bit sad to have to do that. Maybe we could have a Merlin command that returns all the available jump in a more efficient way. (It's not a request for changes, but a note for later. Maybe you can add a comment about it ?)

ocaml-lsp-server/src/code_actions/action_jump.ml
```diff
@@ -0,0 +1,96 @@
...
+let process_jump_request ~merlin ~position ~target =
+  let* results =
+    Document.Merlin.with_pipeline_exn merlin (fun pipeline ->
+      let pposition = Position.logical position in
+      let query = Query_protocol.Jump (target, pposition) in
+      Query_commands.dispatch pipeline query)
+  in
+  match results with
+  | `Error _ -> Fiber.return None
+  | `Found pos -> Fiber.return (Some pos)
```
voodoos, line 55: You could use a `let+` at line 47 and then `Result.to_option`.

voodoos (commented):
Thanks Pizie, I think this is a good compromise to actually provide jump functionnality without having to provide client-side support.

For reference, here is the trace of executing such a code action:

1. First the client request the list of code actions and the server answers:
```
[Trace - 11:47:51] Sending request 'textDocument/codeAction - (71)'.
Params: {
    "textDocument": {
        "uri": "file:///.../prefix_parser.ml"
    },
    "range": {
        "start": {
            "line": 10,
            "character": 42
        },
        "end": {
            "line": 10,
            "character": 42
        }
    },
    "context": {
        "diagnostics": [],
        "triggerKind": 2
    }
}


[Trace - 11:47:51] Received response 'textDocument/codeAction - (71)' in 7ms.
Result: [
    ...,
    {
        "command": {
            "arguments": [
                "file:///.../prefix_parser.ml",
                {
                    "end": {
                        "character": 2,
                        "line": 9
                    },
                    "start": {
                        "character": 2,
                        "line": 9
                    }
                }
            ],
            "command": "ocamllsp/merlin-jump-to-target",
            "title": "Jump to let"
        },
        "kind": "merlin-jump",
        "title": "Jump to let"
    }
]
```

2. Then if the user choose this code action the client ask the server to execute the appropriate command which will make the server send a `showDocument` request.

```
[Trace - 11:47:56] Sending request 'workspace/executeCommand - (73)'.
Params: {
    "command": "ocamllsp/merlin-jump-to-target",
    "arguments": [
        "file:///.../prefix_parser.ml",
        {
            "end": {
                "character": 2,
                "line": 9
            },
            "start": {
                "character": 2,
                "line": 9
            }
        }
    ]
}


[Trace - 11:47:56] Received request 'window/showDocument - (6)'.
Params: {
    "selection": {
        "end": {
            "character": 2,
            "line": 9
        },
        "start": {
            "character": 2,
            "line": 9
        }
    },
    "takeFocus": true,
    "uri": "file:///.../prefix_parser.ml"
}
```

It's a nice trick, but it is a bit sad to have the additional round-trip when all the information is already stored in the `codeAction` response. Do you think it would be possible to have custom client support that jump directly and fallback to the `showDocument` trick for clients without custom support ?

rgrinberg:
> It's a nice trick, but it is a bit sad to have the additional round-trip when all the information is already stored in the codeAction response. Do you think it would be possible to have custom client support that jump directly and fallback to the showDocument trick for clients without custom support

Is this slowdown observable in practice? LSP already contains a ton of inefficiencies (some essential to the protocol, others are our implementation of it). I would expect such a roundtrip to be unimportant given all the LSP traffic.

ocaml-lsp-server/test/e2e-new/code_actions.ml
```diff
@@ -1337,3 +1436,16 @@ let code_action_test ~title source =
   let src, range = parse_selection source in
   Option.iter (apply_code_action title src range) ~f:print_string
 ;;
+
+let apply_code_action_no_edit ?diagnostics title source range =
+  let open Option.O in
+  let code_actions = ref None in
+  iter_code_actions ?diagnostics ~source range (fun ca -> code_actions := Some ca);
+  let* m_code_actions = !code_actions in
```
rgrinberg, line 1444: what does the `m_` prefix mean?

ocaml-lsp-server/test/e2e-new/code_actions.mli
```diff
@@ -18,6 +18,13 @@ val apply_code_action
   -> Range.t
   -> string option
 
+val apply_code_action_no_edit
+  :  ?diagnostics:Diagnostic.t list
+  -> string
```
rgrinberg, line 23: labels on these `string` arguments would be useful

ocaml-lsp-server/test/e2e-new/code_actions.mli
```diff
@@ -18,6 +18,13 @@ val apply_code_action
...
+  -> string
+  -> Range.t
+  -> unit option
```
rgrinberg, line 26: The return type is rather weird. What about a variant to explain what you're returning

rgrinberg: Actually I don't even see where this function is being used

PizieDust: I will remove it. It was "relevant" in an earlier version of this implementation but no longer needed now.

ocaml-lsp-server/src/code_actions/action_jump.ml
PizieDust: the return from Merlin is not a `Result.t`

ocaml-lsp-server/src/code_actions/action_jump.ml
PizieDust: Will create a Merlin issue and reference this comment in it.

ocaml-lsp-server/src/code_actions/action_jump.ml
```diff
@@ -0,0 +1,86 @@
...
+let command_name = "ocamllsp/merlin-jump-to-target"
+
+let targets =
+  [ "fun"; "match"; "let"; "module"; "module-type"; "match-next-case"; "match-prev-case" ]
+;;
+
+let available (capabilities : ShowDocumentClientCapabilities.t option) =
+  match capabilities with
+  | None | Some { support = false } -> false
+  | Some { support = true } -> true
```
xvw, line 14: Nitpick:
You can use `map+value` or: 
```ocaml
match capabilities with
| Some { support } -> support 
| None -> false
```

ocaml-lsp-server/src/code_actions/action_jump.ml
```diff
@@ -0,0 +1,86 @@
...
+;;
+
+let code_actions
+  (doc : Document.t)
+  (params : CodeActionParams.t)
+  (capabilities : ShowDocumentClientCapabilities.t option)
+  =
+  match Document.kind doc with
+  | `Other -> Fiber.return []
+  | `Merlin merlin ->
```
xvw, line 65: I guess that you can simplify the action invokation reordering your pattern match : 
```ocaml
match Document.kind doc with 
| `Merlin merlin when available capabilities -> 
   ...
| _ -> Fiber.return []
```
It save one level of nesting;)

ocaml-lsp-server/src/code_actions/action_jump.ml
```diff
@@ -0,0 +1,86 @@
...
+    (match available capabilities with
+     | false -> Fiber.return []
+     | true ->
+       let+ actions =
+         (* TODO: Merlin Jump command that returns all avaliable jump locations for a source code buffer. *)
```
xvw, line 70:
```diff
- avaliable
+ available
```

ocaml-lsp-server/src/code_actions/action_jump.ml
```diff
@@ -0,0 +1,86 @@
...
+         Fiber.parallel_map targets ~f:(fun target ->
+           let+ res = process_jump_request ~merlin ~position:params.range.start ~target in
+           let open Option.O in
+           let* lexing_pos = res in
+           let+ position = Position.of_lexical_position lexing_pos in
+           let uri = Document.uri doc in
+           let range = { Range.start = position; end_ = position } in
```
xvw, line 77: I maybe miss something but I do not understand why are we relaying on range? Is this because of `merlin protocol`?

xvw (changes requested):
Just two small nitpickings and one question from my review perspective. Thanks @PizieDust !

ocaml-lsp-server/src/code_actions/action_jump.ml
PizieDust: It's the LSP protocol rather. `ShowDocument` accepts a range to be able to "move" or focus the cursor when it opens the document.

voodoos (approved):
Looks good to me :-)

xvw (approved):
👍🏾

rgrinberg (approved):

voodoos:
> I think this approach precludes users being able to trigger these movements via keyboard shortcut because they'll always have to code through the code actions menu. Since all the items begin with the same letters, the user can't easily use the keyboard to jump to one of them in the code actions menu either.

So there is a way to add shortcuts to code action (for example `alt-d` is `destruct` in the vscode-ocaml-platform plugin), but since all the jump commands have the same "kind" it's indded problematic here:
image

 A solution to that problem would be to use different kinds for the different jumps "`merlin-jump-let` so that users could make their custom keybindings. wdyt @awilliambauer ?

awilliambauer:
> A solution to that problem would be to use different kinds for the different jumps "`merlin-jump-let` so that users could make their custom keybindings. wdyt @awilliambauer ?

This would be a good solution!

awilliambauer:
I've discussed this with other JS editors folks and have a couple other concerns:
- These code actions add a lot of clutter to the code actions menu. Would there be a way to suppress them from being displayed in the menu, but still make them available via keyboard shortcut?
- There are various ways editors might want to incorporate these jump commands into a large commend (e.g., a macro that jumps to a let and adds a specific annotation). We'd also want the user to be able to navigate the history of these jumps (i.e., jump backwards and forwards). Having these as code actions instead of a dedicated custom query will probably make supporting both of these things a lot harder.

voodoos:
> I've discussed this with other JS editors folks and have a couple other concerns:
> 
>     * These code actions add a lot of clutter to the code actions menu. Would there be a way to suppress them from being displayed in the menu, but still make them available via keyboard shortcut?
> 
>     * There are various ways editors might want to incorporate these jump commands into a large commend (e.g., a macro that jumps to a let and adds a specific annotation). We'd also want the user to be able to navigate the history of these jumps (i.e., jump backwards and forwards). Having these as code actions instead of a dedicated custom query will probably make supporting both of these things a lot harder.

Right, I guess that there are not other way around than a custom query if we want that kind of flexibility. cc @rgrinberg @PizieDust

PR #1366, opened 2024-09-02
Draft: Use Inlay_hint from merlin-lib

xvw:
Since https://github.com/ocaml/merlin/pull/1812, we can relay on `merlin-lib` to compute inlay hints. 

In draft because we need a new release of merlin-lib


voodoos:
Closing as part of https://github.com/ocaml/ocaml-lsp/pull/1383

PR #1368, opened 2024-09-03
selection_range: use merlin enclosing query instead of shape

voodoos:
@awilliambauer @pitag-ha I looked at the server answer to the `selectionRange` query after our discussion earlier today, and I noticed that they were, in fact, duplicated ranges:

```
[Trace - 17:40:29] Received response 'textDocument/selectionRange - (20)' in 6ms.
Result: [
    {
        "parent": {
            "parent": {
                "parent": {
                    "parent": {
                        "parent": {
                            "parent": {
                                "parent": {
                                    "parent": {
                                        "parent": {
                                            "parent": {
                                                "range": {
                                                    "end": {
                                                        "character": 38,
                                                        "line": 968
                                                    },
                                                    "start": {
                                                        "character": 0,
                                                        "line": 0
                                                    }
                                                }
                                            },
                                            "range": {
                                                "end": {
                                                    "character": 26,
                                                    "line": 409
                                                },
                                                "start": {
                                                    "character": 0,
                                                    "line": 379
                                                }
                                            }
                                        },
                                        "range": {
                                            "end": {
                                                "character": 26,
                                                "line": 409
                                            },
                                            "start": {
                                                "character": 0,
                                                "line": 379
                                            }
                                        }
                                    },
                                    "range": {
                                        "end": {
                                            "character": 26,
                                            "line": 409
                                        },
                                        "start": {
                                            "character": 2,
                                            "line": 380
                                        }
                                    }
                                },
...
            "range": {
                "end": {
                    "character": 26,
                    "line": 409
                },
                "start": {
                    "character": 4,
                    "line": 409
                }
            }
        },
        "range": {
            "end": {
                "character": 26,
                "line": 409
            },
            "start": {
                "character": 20,
                "line": 409
            }
        }
    }
]
```

With that patch these are removed:
```
[Trace - 17:34:27] Received response 'textDocument/selectionRange - (149)' in 12ms.
Result: [
    {
        "parent": {
            "parent": {
                "parent": {
                    "parent": {
                        "parent": {
                            "parent": {
                                "parent": {
                                    "parent": {
                                        "parent": {
                                            "range": {
                                                "end": {
                                                    "character": 38,
                                                    "line": 968
                                                },
                                                "start": {
                                                    "character": 0,
                                                    "line": 0
                                                }
                                            }
                                        },
                                        "range": {
                                            "end": {
                                                "character": 26,
                                                "line": 409
                                            },
                                            "start": {
                                                "character": 0,
                                                "line": 379
                                            }
                                        }
                                    },
                                    "range": {
                                        "end": {
                                            "character": 26,
                                            "line": 409
                                        },
                                        "start": {
                                            "character": 2,
                                            "line": 380
                                        }
                                    }
                                },
...
            "range": {
                "end": {
                    "character": 26,
                    "line": 409
                },
                "start": {
                    "character": 4,
                    "line": 409
                }
            }
        },
        "range": {
            "end": {
                "character": 26,
                "line": 409
            },
            "start": {
                "character": 20,
                "line": 409
            }
        }
    }
]
```

Sadly vscode still feels the need to add extraneous steps, notably when there is some indentation:

https://github.com/user-attachments/assets/004d7466-ec50-4ed9-83d7-f56a09d51d60

That's quite unfortunate... but this PR at least improves the servers' answer and does remove a few no-op steps compared to the previous behaviour.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 - Fix fd leak in running external processes for preprocessing (#1349)
 - Fix prefix parsing for completion of object methods (#1363, fixes #1358)
+- Remove some duplicates in the `selectionRange` answers (#1368)
 
 
 # 1.19.0
--- a/ocaml-lsp-server/src/ocaml_lsp_server.ml
+++ b/ocaml-lsp-server/src/ocaml_lsp_server.ml
@@ -387,41 +387,27 @@ let selection_range
   match Document.kind doc with
   | `Other -> Fiber.return []
   | `Merlin merlin ->
-    let selection_range_of_shapes
-      (cursor_position : Position.t)
-      (shapes : Query_protocol.shape list)
+    let selection_range_of_enclosings (enclosings : Warnings.loc list)
       : SelectionRange.t option
       =
-      let rec ranges_of_shape parent (s : Query_protocol.shape) =
-        let selectionRange =
-          let range = Range.of_loc s.shape_loc in
-          { SelectionRange.range; parent }
-        in
-        match s.shape_sub with
-        | [] -> [ selectionRange ]
-        | xs -> List.concat_map xs ~f:(ranges_of_shape (Some selectionRange))
+      let ranges_of_enclosing parent (enclosing : Warnings.loc) =
+        let range = Range.of_loc enclosing in
+        { SelectionRange.range; parent }
       in
-      (* try to find the nearest range inside first, then outside *)
-      let nearest_range =
-        let ranges = List.concat_map ~f:(ranges_of_shape None) shapes in
-        List.min ranges ~f:(fun r1 r2 ->
-          let inc (r : SelectionRange.t) =
-            Position.compare_inclusion cursor_position r.range
-          in
-          match inc r1, inc r2 with
-          | `Outside x, `Outside y -> Position.compare x y
-          | `Outside _, `Inside -> Gt
-          | `Inside, `Outside _ -> Lt
-          | `Inside, `Inside -> Range.compare_size r1.range r2.range)
-      in
-      nearest_range
+      List.fold_left
+        ~f:(fun parent enclosing -> Some (ranges_of_enclosing parent enclosing))
+        ~init:None
+      @@ List.rev enclosings
     in
     let+ ranges =
       Fiber.sequential_map positions ~f:(fun x ->
-        let+ shapes =
-          Document.Merlin.dispatch_exn ~name:"shape" merlin (Shape (Position.logical x))
+        let+ enclosings =
+          Document.Merlin.dispatch_exn
+            ~name:"shape"
+            merlin
+            (Enclosing (Position.logical x))
         in
-        selection_range_of_shapes x shapes)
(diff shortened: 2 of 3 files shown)


rgrinberg:
Does this work for documents that aren't well typed?

voodoos:
> Does this work for documents that aren't well typed?

I would expect it to behave as good (or as bad) as the current version. They do fairly similar things but shape returns too much information for our need.

voodoos:
I just had this déjà-vu feeling. 

I openned a very similar PR 2 years ago...

https://github.com/ocaml/ocaml-lsp/pull/860

rgrinberg:
Yup. I just read it over and I still don't really understand how much this changes. I suppose you're confident that this is a strict improvement? A few tests would make that a lot more convincing.

To recall my main desire for this feature: the criteria for what is a "selection" should be a purely syntactic one. It would make this command very useful for moving logical chunks of code around that don't need to be typed (comments, records field definitions, match statement clauses).

voodoos:
It is true that the logic is quite different, but so far I was not able to witness any difference... I will try to write more interesting tests.

The de-duplication is [done explicitly in Merlin](https://github.com/ocaml/merlin/blob/main/src/frontend/query_commands.ml#L347-L351) for the `Enclosing` command. If we want we could simply do the same thing for the `Shape` command.

Still, it looks like the output of the `Shape` command is not fitted for the `selectionRange` query since most of the result from Merlin is filtered out to "[try to find the nearest range](https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/src/ocaml_lsp_server.ml#L404-L415)". The enclosing command returns only the typedtree branch containing the cursor, which seems to correspond better to what is needed.

I will write more test, but I would argue that it is a strict improvement: the code is simpler (on both sides, Merlin and OCaml-LSP), and the results do not contain duplicates anymore.

I (still) agree with the fact that the Typedtree is probably not the best way to answer these queries, but that's out of scope for fix PR which aim is to remove duplicates.

rgrinberg:
No argument against this being an improvement. You would know more than me. I would like to see some more tests to make sure we are understanding if there's a change of behavior or a regression. After hover and completion, this is my most used request in LSP. I imagine other users are also relying on this request heavily. We would all welcome improvements in this functionality.

voodoos:
@rgrinberg I added a few more tests and so far these changes only remove duplicate entries from the results. I also tried the feature a bit around the codebase and I am quite confident that there is no bad change in behavior. (I am not a regular user however, so I might miss some differences.) If you have some suggestion for some more precise tests cases I will be happy to add them, but right now I think this is ready to be merged.

rgrinberg (approved):
Some tests demonstrating the behavior with:

1. modules containing a syntax error
2. modules that are unable to type check

Would be appropriate I think

voodoos:
I added to more cases, one with a module with a definition that fails to type check (the return type is not the expected ones), and one with an additional syntax error: the `end` keyword is misspelled.

In both cases the new results appear to be a strict improvement (whose main aspect is the remove of duplicates).

xvw (approved):
Top!

PR #1369, opened 2024-09-04
Type search custom request

PizieDust:
# TypeSearch Request

## Description

This custom request allows clients to perform a type search at a specific position within a text document based on finding functions or types that match a specific query pattern.

## Server capability

- property name: `handleTypeSearch`
- property type: `boolean`

## Request

```js
export interface TypeSearchParams extends TexDocumentPositionParams
{
    query: string;
    limit: int;
    with_doc: bool;
}
```
- method: `ocamllsp/typeSearch`
- params:
    - `TextDocumentPositionParams`: This is an existing interface that includes:
            - `TextDocumentIdentifier`: Specifies the document uri for which the request is sent.
            - `Position`: Specifies the cursor position.
        More details can be found in the [TextDocumentPositionParams - LSP Specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams).
    - `query`: The search pattern.
    - `limit`: The number of results to return
    - `with_doc`: If to return documentation information or not

## Response

```js
result: TypeSearch | null
export interface TypeSearch {
    type t = Query_protocol.type_search_result list
}
```
- `t`: A list of types that match the query.
    ```
    type Query_protocol.type_search_result =
    {
        name : string; // The fully qualified name of this result.
        typ : string;  // The signature of this result.
        loc : Range.t; // The location of the definition of this result in the source code.
        doc : string option; // Optional documentation associated with this result.
        cost : int; // A numeric value representing the "cost" or distance between this result and the query.
        constructible : string; // A constructible form or template that can be used to invoke this result
    }
    ```
- A response with null result is returned if no entries are found.


ocaml-lsp-server/src/custom_requests/req_polarity_search.ml
```diff
@@ -0,0 +1,124 @@
...
+  | EnumMember -> `Variant
+  | Field -> `Label
+  | Module -> `Module
+  | Interface -> `Modtype
+  | Method -> `MethodCall
+  | Keyword -> `Keyword
+  | Constructor -> `Constructor
+  | TypeParameter -> `Type
+  | _ -> failwith "Unknown kind"
+;;
```
xvw, line 32: There is already https://github.com/ocaml/ocaml-lsp/blob/3dc1a6633bfe2ebd34d7aa38f008c8c28300ee55/ocaml-lsp-server/src/compl.ml#L13 and since `merlin-search` always returns a kind of `Value`, I am not sure that having a full completion make sense here. WDYT?

xvw (changes requested):
Nice!

ocaml-lsp-server/src/custom_requests/req_polarity_search.ml
```diff
@@ -0,0 +1,124 @@
...
+    { Query_protocol.Compl.name; kind; desc; info; deprecated }
+  ;;
+
+  let yojson_of_entry { Query_protocol.Compl.name; kind; desc; info; deprecated } =
+    `Assoc
+      [ "name", `String name
+      ; "kind", CompletionItemKind.yojson_of_t (completion_kind kind)
+      ; "desc", `String desc
+      ; "info", `String info
+      ; "deprecated", `Bool deprecated
```
voodoos, line 80: I think we decided that we didn't want these fields in the answer. They are always "" and false. There are here because Merlin abuses the `Query_protocol.completions` type, but they should not be a part of our new interface.

ocaml-lsp-server/src/custom_requests/req_polarity_search.ml
```diff
@@ -0,0 +1,124 @@
...
+    let open Yojson.Safe.Util in
+    let name = json |> member "name" |> to_string in
+    let kind =
+      json |> member "kind" |> CompletionItemKind.t_of_yojson |> kind_of_completion_kind
+    in
+    let desc = json |> member "desc" |> to_string in
+    let info = json |> member "info" |> to_string in
+    let deprecated = json |> member "deprecated" |> to_bool in
+    { Query_protocol.Compl.name; kind; desc; info; deprecated }
+  ;;
```
voodoos, line 72: Do we really need to have the `*_of_yojson` versions of these object we never need to decode ? Is it to fullfil an imposed interface ? (same question for the following t_of_yojson).

ocaml-lsp-server/src/custom_requests/req_polarity_search.ml
xvw: In fact, `info`, `kind` and `deprecated` are constant in the polarity search implem. So I suggest to keep the output simple: 

```json
{
  path: string  // or name 
  type: string
}
```

rgrinberg:
Okay, I think we need to see that vscode client to evaluate this work.

ocaml-lsp-server/src/custom_requests/req_polarity_search.ml
rgrinberg: You don't need to write encoder/decoders if you don't use them. `ppx_yojson_conv` allows you to generate converters one way. It's only necessary to make sure the conventions of `ppx_yojson_conv` are respected so all decoders/encoders compose whether they are generated or not.

ocaml-lsp-server/test/e2e-new/polarity_search.ml
```diff
@@ -0,0 +1,320 @@
...
+let%expect_test "Search for a simple query that takes an int and returns a string" =
+  let source = "" in
+  let line = 1 in
+  let character = 0 in
+  Util.test ~line ~character ~query:"-int +string" source;
+  [%expect
+    {|
+    [
+      {
+        "name": "string_of_int",
```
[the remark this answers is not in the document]
xvw: Unfortunately no, but we can probably filter the result of test just keeping value prefixed by a given prefix? WDYT @PizieDust ?

ocaml-lsp-server/src/custom_requests/req_polarity_search.ml
voodoos: And would it be useful to allow to have the documentation in the "info" field ?

PizieDust: Yes documentation can be great from the info field. Should Merlin find and return the documentation automatically in the results?

xvw: Yes it could be useful but it need to change the polarity search function, make a new release of merlin, and affect this PR. It seems a lot of work for an hard-to-use feature ;)?

ocaml-lsp-server/docs/ocamllsp/polaritySearch-spec.md
```diff
@@ -0,0 +1,40 @@
...
+        More details can be found in the [TextDocumentPositionParams - LSP Specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams).
+    - `query`: The search pattern.
+    - `limit`: The number of results to return
+
+## Response
+
+```js
+result: PolaritySearch | null
+export interface PolaritySearch {
+    type entry = { path : string; desc : string }
```
xvw, line 35: Biksheeding here but `desc` seems not very relevant, what about `type`?

ocaml-lsp-server/src/custom_requests/custom_request.ml
```diff
@@ -6,3 +6,4 @@ module Typed_holes = Req_typed_holes
 module Type_enclosing = Req_type_enclosing
 module Wrapping_ast_node = Req_wrapping_ast_node
 module Get_documentation = Req_get_documentation
+module PolaritySearch = Req_polarity_search
```
xvw, line 9: I think that we should follow the `case convention`:  `Polarity_search`

ocaml-lsp-server/docs/ocamllsp/polaritySearch-spec.md
PizieDust: I think `type` should be a keyword that can't be used as a record field

ocaml-lsp-server/src/custom_requests/req_polarity_search.ml
rgrinberg: If this feature is hard-to-use - essentially unusable until there's a working client, what's the rush to release it? Might as well wait until you adjust the API to make it work well while there's no clients relying on it.

ocaml-lsp-server/docs/ocamllsp/polaritySearch-spec.md
xvw: We use `type` (and `type_` for the OCaml record) in Type Enclosing

ocaml-lsp-server/src/custom_requests/req_polarity_search.ml
PizieDust: I could open an issue on Merlin linking to this comment thread, and we can also subsequently implement all the updates/enhancements in future PR's no?

ocaml-lsp-server/src/custom_requests/req_polarity_search.ml
```diff
@@ -0,0 +1,102 @@
...
+let dispatch merlin position limit query =
+  Document.Merlin.with_pipeline_exn merlin (fun pipeline ->
+    let position = Position.logical position in
+    let query = Query_protocol.Polarity_search (query, position) in
+    let completions = Query_commands.dispatch pipeline query in
+    PolaritySearch.yojson_of_t
+      (List.map
+         ~f:(fun entry ->
+           { PolaritySearch.path = entry.Query_protocol.Compl.name; desc = entry.desc })
+         (if List.length completions.entries > limit
```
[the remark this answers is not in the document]
PizieDust: I can't find the `take` method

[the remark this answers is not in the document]
PizieDust: but I can use it if I do `Base.List...`

ocaml-lsp-server/src/custom_requests/req_polarity_search.ml
rgrinberg: It would make sense if this work was already in a usable state and could provide some value to users while enhancements are made, but thus far, it's not really the case. Moreover, if the API in merlin changes and this feature is released, you will need to add constraints to already released versions. Sometimes, this isn't done in time, so users end up experiencing build failures.

On the other hand, if it stays as a PR, you're free to iterate on the feature in merlin, lsp, and the client all at the same time without worrying about backwards compat or versioning. Usually, the cost of that would be potential merge conflicts. But I see that the way you've written this feature in a fairly isolated manner, so the possibility of merge conflicts is low.

Ultimately, it's not forbidden to merge WIP as sometimes the cost of merge conflicts can be rather high. Evolving 3 concurrent code bases in a way that is user conscious is also not free though. My experience hints that in this particular case, the cost of the latter will outweigh the former.

ocaml-lsp-server/test/e2e-new/polarity_search.ml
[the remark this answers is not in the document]
rgrinberg: That's definitely an improvement, but I'm kind of surprised that we can't write our fixtures. Is it because the test doesn't allow us to setup a dune project that doesn't use the stdlib?

ocaml-lsp-server/src/custom_requests/req_polarity_search.ml
xvw: I think that the `Polarity search feature` is used by merlin User, in it's current state, but I think that the feature will not change in the future (and seems the name of the command is very precise : `polarity_search`), for me, it can be useful and not "to be deprecated". So for me, the command will probably not requiring any maintenance in the future.

ocaml-lsp-server/test/e2e-new/polarity_search.ml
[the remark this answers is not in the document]
xvw: Ah yes, we can probably make a test without any dependencies (and that pass the flag `nostdlib`

ocaml-lsp-server/docs/ocamllsp/typeSearch-spec.md
```diff
@@ -0,0 +1,52 @@
...
+    type Query_protocol.type_search_result =
+    {
+        name : string; // The fully qualified name of this result.
+        typ : string;  // The signature of this result.
+        loc : Range.t; // The location of the definition of this result in the source code.
+        doc : string option; // Optional documentation associated with this result.
+        cost : int; // A numeric value representing the "cost" or distance between this result and the query.
+        constructible : string; // A constructible form or template that can be used to invoke this result
+    }
+    ```
```
voodoos, line 51: This doc should describe the returned object using json syntax, not the OCaml type.

ocaml-lsp-server/src/custom_requests/req_type_search.ml
```diff
@@ -0,0 +1,92 @@
...
+  let yojson_of_t (t : t) =
+    let yojson_of_type_search_result (res : string Query_protocol.type_search_result) =
+      `Assoc
+        [ "name", `String res.name
+        ; "typ", `String res.typ
+        ; "loc", Range.yojson_of_t (Range.of_loc res.loc)
+        ; ( "doc"
+          , match res.doc with
+            | Some d -> `String d
+            | None -> `Null )
```
voodoos, line 52: This could be replaced by `Json.option Json.string res.doc`

voodoos (approved):
Overall this looks good to me, thanks @PizieDust

voodoos:
@xvw, @PizieDust, Melrin 5.2-502 has been released so the CI should pass without pinning and could be un-drafted. Is the custom request ready for a last round of review ?

ocaml-lsp-server/docs/ocamllsp/typeSearch-spec.md
```diff
@@ -0,0 +1,55 @@
...
+        More details can be found in the [TextDocumentPositionParams - LSP Specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams).
+    - `query`: The search pattern.
+    - `limit`: The number of results to return
+    - `with_doc`: If to return documentation information or not
+
+## Response
+
+```js
+result: TypeSearch | null
+export interface TypeSearch {
```
xvw, line 36: I am still a little bit confuse by the syntactic choice fo expressing JSON

voodoos:
@PizieDust I think we should, similarly as for the `Type enclosing` custom request, provide a way to ask for different format for the documentation, like markdown, and maybe plaintext and html ?

ocaml-lsp-server/src/custom_requests/req_type_search.ml
```diff
@@ -0,0 +1,118 @@
...
+        :: ("with_doc", `Bool with_doc)
+        :: [ "query", `String query ])
+       @ doc_format)
+  ;;
+end
+
+module TypeSearch = struct
+  type t = string Query_protocol.type_search_result list
+
+  let create ~kind ~value =
```
voodoos, line 52: That's not a very informative name. `doc_to_markupContent` ?

ocaml-lsp-server/docs/ocamllsp/typeSearch-spec.md
```diff
@@ -0,0 +1,51 @@
...
+    - `with_doc`: If to return documentation information or not
+
+## Response
+```json
+{
+    [
+        "name": string,
+        "typ": string,
+        "loc": Range,
+        "doc": string,
```
voodoos, line 39: And here too

voodoos (approved):
Thanks, looks good! Let's wait for the vscode plugin implementation to catch up before merging.

voodoos:
Client support in VScode is on the way: https://github.com/ocamllabs/vscode-ocaml-platform/pull/1626

PR #1374, opened 2024-09-18
Jump Custom Request

PizieDust:
related to #1360 
This PR contains an implementation for a custom requests based on Merlin's jump command.

Using code actions introduces some noise as referenced in this comment: https://github.com/ocaml/ocaml-lsp/pull/1364#issuecomment-2353283155
> These code actions add a lot of clutter to the code actions menu. Would there be a way to suppress them from being displayed in the menu, but still make them available via keyboard shortcut?
There are various ways editors might want to incorporate these jump commands into a large commend (e.g., a macro that jumps to a let and adds a specific annotation). We'd also want the user to be able to navigate the history of these jumps (i.e., jump backwards and forwards). Having these as code actions instead of a dedicated custom query will probably make supporting both of these things a lot harder.

An alternative solution will be to use a custom request, which allows for more better finetuning.

# Merlin Jump Request

## Description

This custom request allows Merlin-type code navigation in a source buffer.

## Server capability

- propert name: `handleMerlinJump`
- property type: `boolean`

## Request

```js
export interface JumpParams extends TextDocumentPositionParams
{
    target: string;
}
```

- method: `ocamllsp/merlinJump`
- params:
    - `TextDocumentIdentifier`: Specifies the document for which the request is sent. It includes a uri property that points to the document.
    - `Position`: Specifies the position in the document for which the documentation is requested. It includes line and character properties.
    More details can be found in the [TextDocumentPositionParams - LSP Specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams).
    - `Target`: A string representing the identifier within the document to search for and jump to.
                The target can be any of `fun`, `let`, `module`, `module-type`, `match`, `match-next-case`, `match-prev-case`.

## Response

```js
result: Jump | String
export interface Jump extends TextDocumentPositionParams {
}
```

- result:
    - Type: Jump or string
    - Description: If the jump is successful, a position and document path is returned. If no relevant jump location is found, the result will be a string "no matching target" or an error message.
    - Jump:
        - Type: TextDocumentPositionParams
            - `Position`: The position to jump to
            - `TextDocumentIdentifier`: the document path which contains this position (ideally the same document as the request)

cc @voodoos @xvw @rgrinberg @awilliambauer


ocaml-lsp-server/docs/ocamllsp/merlinJump-spec.md
```diff
@@ -0,0 +1,42 @@
...
+    target: string;
+}
+```
+
+- method: `ocamllsp/merlinJump`
+- params:
+    - `TextDocumentIdentifier`: Specifies the document for which the request is sent. It includes a uri property that points to the document.
+    - `Position`: Specifies the position in the document for which the documentation is requested. It includes line and character properties.
+    More details can be found in the [TextDocumentPositionParams - LSP Specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams).
+    - `Target`: A string representing the identifier within the document to search for and jump to.
```
voodoos, line 26: I think we should list the value accepted by Merlin here.

ocaml-lsp-server/src/custom_requests/req_merlin_jump.mli
```diff
@@ -0,0 +1,21 @@
...
+    :  text_document:TextDocumentIdentifier.t
+    -> position:Position.t
+    -> target:string
+    -> unit
+    -> t
+end
+
+type t
+
+val t_of_yojson : Json.t -> t
```
voodoos, line 18: Is this a requirement for the custom request ? it doesn't look like its in use.

ocaml-lsp-server/src/custom_requests/req_merlin_jump.ml
```diff
@@ -0,0 +1,80 @@
...
+let on_request ~params state =
+  Fiber.of_thunk (fun () ->
+    let params = (Option.value ~default:(`Assoc []) params :> Yojson.Safe.t) in
+    let JumpParams.{ text_document; position; target } = JumpParams.t_of_yojson params in
+    let uri = text_document.uri in
+    let doc = Document_store.get state.State.store uri in
+    match Document.kind doc with
+    | `Other -> Fiber.return `Null
+    | `Merlin merlin ->
+      Fiber.bind (dispatch ~merlin ~position ~target) ~f:(fun res ->
```
voodoos, line 67: You could use `Fiber.map` (or `let+`) and remove all the `Fiber.return ` in the body.

ocaml-lsp-server/src/custom_requests/req_merlin_jump.mli
PizieDust: Yes this is right. I'll get rid of it.

voodoos:
We should think about what we want to do with the jump code action. The two obvious possibilities are:

1. Revert the changes
2. Add an option to disable the jump code actions

I am personally in favor of 2, since the nice trick allows direct support in any editor with the required capabilities.
It could be enabled by default in the community version of the server/plugin, until we ourselves implement an alternative way using the custom request.

PizieDust:
> We should think about what we want to do with the jump code action. The two obvious possibilities are:
> 
> 1. Revert the changes
> 2. Add an option to disable the jump code actions
> 
> I am personally in favor of 2, since the nice trick allows direct support in any editor with the required capabilities. It could be enabled by default in the community version of the server/plugin, until we ourselves implement an alternative way using the custom request.

I am also in favor of 2. 
I am currently working on the option to disable the jump code actions.

CHANGES.md
```diff
@@ -10,6 +10,8 @@
 
 - Add custom [`ocamllsp/getDocumentation`](/ocaml-lsp-server/docs/ocamllsp/getDocumentation-spec.md) request (#1336)
 
+- Add custom [`ocamllsp/merlinJump`](/ocaml-lsp-server/docs/ocamllsp/merlinJump-spec.md) request (#1374)
+
```
voodoos, line 14: This is not the correct log section, it should be part of the # Unreleased section.

voodoos (approved):
These changes looks ok to me. It's one more custom request as discussed with @awilliambauer , but hopefully we will be able to work on client implementations soon.

xvw (approved):

PR #1375, opened 2024-09-20
Draft: Move to upstream signature-help

PizieDust:
cc @voodoos @xvw

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,8 @@
 
 - Add a code-action for syntactic and semantic movement shortcuts based on Merlin's Jump command (#1364)
 
+- Upstream the implementation of signature help to merlin-lib (#1375)
+
 ## Fixes
 
 - Kill unnecessary ocamlformat processes with sigterm rather than sigint or
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,8 +15,6 @@
 
 - Add a code-action for syntactic and semantic movement shortcuts based on Merlin's Jump command (#1364)
 
-- Upstream the implementation of signature help to merlin-lib (#1375)
-
 ## Fixes
 
 - Kill unnecessary ocamlformat processes with sigterm rather than sigint or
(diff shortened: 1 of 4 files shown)


voodoos:
The diff in the testsuite shows that the documentation if not fetch as it should be. We should either have Merlin fetch the doc itself or return the position of the function identifier, I will have a look.

ocaml-lsp-server/src/signature_help.ml
```diff
@@ -195,51 +42,24 @@ let run (state : State.t) { SignatureHelpParams.textDocument = { uri }; position
...
-    let* application_signature =
-      let* inside_comment = Check_for_comments.position_in_comment ~position ~merlin in
-      match inside_comment with
-      | true -> Fiber.return None
-      | false ->
-        Document.Merlin.with_pipeline_exn ~name:"signature-help" merlin (fun pipeline ->
-          let typer = Mpipeline.typer_result pipeline in
-          let pos = Mpipeline.get_lexing_pos pipeline pos in
-          let node = Mtyper.node_at typer pos in
-          application_signature node ~prefix)
```
voodoos, line 220: I thin you should be able to use `Merlin_analysis.Signature_help.application_signature` as a drop-in replacement here and keep the rest of the `run` function identical :-)

CHANGES.md
```diff
@@ -15,6 +15,8 @@
 
 - Add a code-action for syntactic and semantic movement shortcuts based on Merlin's Jump command (#1364)
 
+- Upstream the implementation of signature help to merlin-lib (#1375)
```
rgrinberg, line 18: CHANGES are for the users. Don't think they're going to find this information useful.

xvw (approved):
👍🏾 
Since the patch depend on a release of Merlin, I took the liberty to flag-it as a draft!

rgrinberg (approved):

voodoos:
Closing as part of https://github.com/ocaml/ocaml-lsp/pull/1383

PR #1367, opened 2024-09-03
Reduce typedtree usage in infer interface

xvw:
The goal is to localize Typedtree interaction into `merlin-lib`.

PR #1383, opened 2024-09-30
Rely on upstream Merlin analyses

voodoos:
This brings together changes from [#1366](https://github.com/ocaml/ocaml-lsp/pull/1366), https://github.com/ocaml/ocaml-lsp/pull/1367 and https://github.com/ocaml/ocaml-lsp/pull/1375.

We now rely on `merlin-lib` implementations for inlay-hint and signature help.


voodoos:
NIx and windows breakages are due to the switch to ppx_expect 0.17 and are unrelated to that PR.

PR #1386, opened 2024-10-04
Compatibility with Merlin 503 preview

voodoos:


ocaml-lsp-server.opam
```diff
@@ -31,7 +31,6 @@ depends: [
   "dyn"
   "stdune"
   "fiber" {>= "3.1.1" & < "4.0.0"}
-  "ocaml" {>= "5.2.0"}
```
rgrinberg, line 34: This seems wrong. Don't we need a lower bound based on the stdlib/language features we're using?

voodoos: Yes, but I would expect that lower-bound to coincides with the one imposed by the compatible `merlin-lib` versions. Not specifying the lower-bound on OCaml here means Opam CI will fail if we don't respect that invariant.

I don't mind adding it back of course if that's a better practice. 

For the record: I plan to do releases of Merlin and LSP for 5.2 next week, and for 5.3 around the time of the final release.

rgrinberg: I think it makes sense to always specify direct deps as precisely as possible and not to rely on transitive deps. For example, what if we start using a new language feature or a new function. Our lower bound might be higher than merlin's? Conversely, what if merlin expands its support to older versions of OCaml that we can't build against?

voodoos: > Our lower bound might be higher than merlin

This is something we want to avoid for releases of LSP to be compatible with as many versions as merlin-lib as possible (which is only 5.2 and 5.3 these days anyway).

> I think it makes sense to always specify direct deps as precisely as possible and not to rely on transitive deps. 

Works for me, I will make the change.

PR #1387, opened 2024-10-04
Add support for URI fragments

Lysxia:
This is useful because in VSCode, file system URIs can contain line numbers in the fragment: `file:///path#N`.

### DIFF
--- a/lsp/src/uri0.ml
+++ b/lsp/src/uri0.ml
@@ -150,10 +150,10 @@ let to_string { scheme; authority; path; query; fragment } =
      Buffer.add_char buff '?';
      Buffer.add_string buff (encode q));
   (match fragment with
-  | None -> ()
-  | Some f ->
-    Buffer.add_char buff '#';
-    Buffer.add_string buff (encode f));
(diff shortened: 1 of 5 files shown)


voodoos:
Do you have examples of requests to the server that requires correct handling of these fragments ? Or is it that you use the `lsp` library in an other application ?

Lysxia:
I am using lsp in my own language server where I would like to send a URI with a fragment. Specifically, I'm trying to use the `DocumentLink` feature to link to another file. The target of `DocumentLink` can only be a URI, with no field for position information. The fragment lets a URI point to a file location.

rgrinberg:
Thanks for the contribution. I think you forgot to promote the test. Can you review them at [`73363fd` (#1387)](https://github.com/ocaml/ocaml-lsp/pull/1387/commits/73363fd9f87b27035db11b4a7410f28cf4eb5fe5)

If those look correct to you, this PR is good to go.

Lysxia (commented):
The tests look good to me!

PR #1391, opened 2024-10-11
Minor patch for Jump codeactions

PizieDust:
cc @voodoos

### DIFF
--- a/ocaml-lsp-server/src/code_actions.ml
+++ b/ocaml-lsp-server/src/code_actions.ml
@@ -118,7 +118,7 @@ let compute server (params : CodeActionParams.t) =
     let open_related = Action_open_related.for_uri capabilities doc in
     let* merlin_jumps =
       match state.configuration.data.merlin_jump_code_actions with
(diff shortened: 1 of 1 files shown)


ocaml-lsp-server/src/code_actions.ml
```diff
@@ -118,7 +118,7 @@ let compute server (params : CodeActionParams.t) =
     let open_related = Action_open_related.for_uri capabilities doc in
     let* merlin_jumps =
       match state.configuration.data.merlin_jump_code_actions with
-      | Some { enable = true } -> Action_jump.code_actions doc params capabilities
+      | Some { enable = true } | None -> Action_jump.code_actions doc params capabilities
       | _ -> Fiber.return []
```
voodoos, line 122:
```suggestion
      | Some { enable = false } -> Fiber.return []
```

Let's make the matching explicit since there is only one case left.

voodoos (approved):
The `None` case is when the editor client did not send a value for this configuration option and we want to have it enabled by default. LGTM.

PR #1412, opened 2024-12-02
Prepare release 1.20.1

voodoos:


yami-no-ryuu:
Is the reason for ocaml < 5.3 mentioned anywhere? Can't see a bug or comment.

voodoos:
> Is the reason for ocaml < 5.3 mentioned anywhere? Can't see a bug or comment.

That's because LSP versions for multiple OCaml versions are often incompatible between each others due to the reliance on Merlin and internal compiler libs. See https://github.com/ocaml/ocaml-lsp/pull/1386 for the branch compatible with 5.3 that is used for the preview (and soon final) releases.

PR #1416, opened 2024-12-06
Add a new option to mute the hover response.

voodoos:
As far as I know it is not possible to disable the default hover provider on the client side in vscode due to how the official LSP client implementation manage it.

However it is sometimes necessary to disable that hover provider to provide alternative information in the hover tooltips without additional clutter.

This PR introduces an option that allows the client to ask the server to not return anything when hover requests are made.

The diff is a bit larger than necessary because `dune build @lint` had not been run when options when last updated so the inlining has changed a bit.

The first use case for this option is described in this vscode-ocaml-platform PR: https://github.com/ocamllabs/vscode-ocaml-platform/pull/1675

### DIFF
--- a/ocaml-lsp-server/src/config_data.ml
+++ b/ocaml-lsp-server/src/config_data.ml
@@ -241,6 +241,78 @@ module ExtendedHover = struct
   [@@@end]
 end
 
+module StandardHover = struct
+  type t = { enable : bool [@default true] }
+  [@@deriving_inline yojson] [@@yojson.allow_extra_fields]
+
+  let _ = fun (_ : t) -> ()
+
+  let t_of_yojson =
+    (let _tp_loc = "ocaml-lsp-server/src/config_data.ml.StandardHover.t" in
+     function
+     | `Assoc field_yojsons as yojson ->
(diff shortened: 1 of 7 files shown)


ocaml-lsp-server/src/config_data.ml
```diff
@@ -241,6 +241,78 @@ module ExtendedHover = struct
...
+             let enable_value = Ppx_yojson_conv_lib.( ! ) enable_field in
+             { enable =
+                 (match enable_value with
+                  | Ppx_yojson_conv_lib.Option.None -> true
+                  | Ppx_yojson_conv_lib.Option.Some v -> v)
+             }))
+     | _ as yojson ->
+       Ppx_yojson_conv_lib.Yojson_conv_error.record_list_instead_atom _tp_loc yojson
+     : Ppx_yojson_conv_lib.Yojson.Safe.t -> t)
+  ;;
```
xvw, line 295: Is that generated code? I am a little bit confused by the semantic (and `let (_: t) = ()`?

xvw (approved):
Wow configuration capabilities are hard!

PR #1419, opened 2024-12-10
Update npm deps

smorimoto:


lsp/bin/metamodel/metaModel.json
rgrinberg, line 1: This file is just copied from upstream. No need to format it.

smorimoto: Fixed!

smorimoto:
@rgrinberg Gentle ping

PR #1421, opened 2024-12-12
Update .github/dependabot.yml

smorimoto:


smorimoto:
@rgrinberg Could you merge this?

rgrinberg:
What does this do?

smorimoto:
This change is needed because we changed the root of yarn to the repository root directory.

PR #1422, opened 2024-12-13
Add missing dev depends

smorimoto:


lsp.opam
```diff
@@ -26,7 +26,8 @@ depends: [
   "dune" {>= "3.0"}
   "jsonrpc" {= version}
   "yojson"
-  "ppx_yojson_conv_lib" {>= "v0.14"}
+  "ppx_yojson_conv" {>= "v0.14"}
```
rgrinberg, line 29: Do we need the preprocessor? We're only using the runtime

smorimoto: @rgrinberg Oops, I guess we actually needed different changes. Updated PR.

lsp/src/dune
```diff
@@ -4,8 +4,6 @@
  (name lsp)
  (public_name lsp)
  (libraries jsonrpc ppx_yojson_conv_lib uutf yojson)
- (lint
-  (pps ppx_yojson_conv))
```
rgrinberg, line 8: We need this in development to promote the generated ppx derivers

smorimoto: Should we list that dependency for development?

rgrinberg: I think that would be correct

smorimoto: Fixed.

PR #1461, opened 2025-01-31
Update Jump functionality

PizieDust:
This PR enhances jump functionality used in : 
- Jump Code Actions
- The Jump Custom request

Instead of recursively checking if each target is possible, we now send a single query and get all the possible jump targets.

The trace on computing code actions is: 

- Getting each target separately
`[Trace - 5:07:56 AM] Received response 'textDocument/codeAction - (2)' in 523ms.`
`[Trace - 5:08:35 AM] Received response 'textDocument/codeAction - (17)' in 18ms.`

- Getting all targets at once
`[Trace - 5:15:50 AM] Received response 'textDocument/codeAction - (2)' in 244ms`.
`Trace - 5:15:59 AM] Received response 'textDocument/codeAction - (11)' in 7ms.`

which indicates some performance improvement.

This Pr is dependent on the corresponding Merlin implementation  https://github.com/ocaml/merlin/pull/1891

cc @voodoos

### DIFF
--- a/ocaml-lsp-server/src/custom_requests/req_merlin_jump.ml
+++ b/ocaml-lsp-server/src/custom_requests/req_merlin_jump.ml
@@ -5,39 +5,23 @@ let meth = "ocamllsp/jump"
 let capability = "handleJump", `Bool true
 
 module JumpParams = struct
-  let targets =
-    [ "fun"
-    ; "match"
-    ; "let"
-    ; "module"
-    ; "module-type"
-    ; "match-next-case"
-    ; "match-prev-case"
-    ]
-  ;;
-
   type t =
     { textDocument : TextDocumentIdentifier.t
     ; position : Position.t
-    ; target : string option
     }
 
   let t_of_yojson json =
     let open Yojson.Safe.Util in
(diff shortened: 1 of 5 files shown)


PizieDust:
requires `no changelog` label

ocaml-lsp-server/src/custom_requests/req_merlin_jump.ml
```diff
@@ -58,47 +21,39 @@ end
...
-               | None -> None
-               | Some position -> Some (target, position))))
-      in
-      Jump.yojson_of_t (List.filter_map results ~f:Fun.id))
+      let+ res = dispatch ~merlin ~position in
+      Jump.yojson_of_t
+        (List.filter_map res ~f:(fun (target, position) ->
+           match Position.of_lexical_position position with
+           | Some pos -> Some (target, pos)
+           | None -> None)))
```
xvw, line 58:
```ocaml
Option.map (fun pos -> target, pos)
  Position.of_lexical_position position
```

xvw:
Does it change the behaviour of the custom request?

PizieDust:
> Does it change the behaviour of the custom request?

No it doesn't. The output should be the same.

xvw (approved):

PR #1483, opened 2025-02-23
chore: remove all the stdune [O] references

rgrinberg:


voodoos (approved):
Is `stdune` still a dependency after this ?

edit: Since it's only the `O` references I guess so. Did you remove these because of upstream changes ?

PR #1516, opened 2025-04-10
More control on typed holes (for navigation)

xvw:
At the moment, there is only one request to return the list of typed holes, which means that a lot of client-side logic has to be implemented. For example, in `ocaml-eglot` ([see](https://github.com/tarides/ocaml-eglot/blob/main/ocaml-eglot.el#L323)):

```elisp
;; Holes

;; TODO: It would probably be possible to improve the query to embed
;; more logic at the `merlin-lib` level rather than calculating hole
;; logic at the editor level.

(defun ocaml-eglot--first-hole-aux (holes pos comparison)
  "Return the first hole of the list HOLES since a POS using COMPARISON."
  (when (or holes (not (equal [] holes)))
    (let* ((hd (car holes))
           (tl (cdr holes))
           (h-start (cl-getf hd :start))
           (cmp (ocaml-eglot-util--compare-position h-start pos)))
      (if (funcall comparison cmp 0) hd
        (ocaml-eglot--first-hole-aux tl pos comparison)))))

(defun ocaml-eglot--first-hole-at (holes pos comparison)
  "Return the first hole of the list HOLES since a POS using COMPARISON.
If there is no available holes, it returns the first one of HOLES."
  (let ((hole (ocaml-eglot--first-hole-aux holes pos comparison)))
    (if hole hole (car holes))))

(defun ocaml-eglot--get-first-hole-in (start end)
  "Return the first hole in a given range denoted by START and END."
  (let* ((holes (ocaml-eglot-req--holes))
         (hole (ocaml-eglot--first-hole-at holes start '>=)))
    (when hole
      (let ((hole-start (cl-getf hole :start))
            (hole-end (cl-getf hole :end)))
        (when (and (>= (ocaml-eglot-util--compare-position hole-start start) 0)
                   (<= (ocaml-eglot-util--compare-position hole-end end) 0))
          hole)))))

(defun ocaml-eglot--first-hole-in (start end)
  "Jump to the first hole in a given range denoted by START and END."
  (when-let* ((hole (ocaml-eglot--get-first-hole-in start end))
              (hole-start (cl-getf hole :start)))
    (ocaml-eglot-util--jump-to hole-start)))

(defun ocaml-eglot-hole-prev ()
  "Jump to the previous hole."
  (interactive)
  (ocaml-eglot-req--server-capable-or-lose :experimental :ocamllsp :handleTypedHoles)
  (let* ((current-pos (eglot--pos-to-lsp-position))
         (holes (reverse (ocaml-eglot-req--holes)))
         (hole (ocaml-eglot--first-hole-at holes current-pos '<)))
    (when hole (ocaml-eglot-util--jump-to-range hole))))

(defun ocaml-eglot-hole-next ()
  "Jump to the next hole."
  (interactive)
  (ocaml-eglot-req--server-capable-or-lose :experimental :ocamllsp :handleTypedHoles)
  (let* ((current-pos (eglot--pos-to-lsp-position))
         (holes (ocaml-eglot-req--holes))
         (hole (ocaml-eglot--first-hole-at holes current-pos '>)))
    (when hole (ocaml-eglot-util--jump-to-range hole))))
```

This patch allows the next/previous hole to be related to a given piece of contextual information. It is possible to include a range to return a hole only in a given range.

(cc @awilliambauer)


ocaml-lsp-server/docs/ocamllsp/jumpTypedHoles-spec.md
```diff
@@ -0,0 +1,57 @@
+# Jump To Typed Holes Requet
+
+### Description
+
+Returns the next or previous typed hole at a given position (included
+in a range or not).
```
voodoos, line 6:
```suggestion
Returns the next or previous typed hole at a given position. An optional
range can be used to restrict the search.
```

ocaml-lsp-server/src/custom_requests/req_jump_typed_hole.ml
```diff
@@ -0,0 +1,136 @@
...
+
+let holes_in_range range holes =
+  match range with
+  | None -> holes
+  | Some range -> List.filter ~f:(Range.contains range) holes
+;;
+
+let find_prev_hole ~range ~position holes =
+  let holes = holes_in_range range holes in
+  Base.List.fold_until
```
voodoos, line 76: I expect that Base is already open right ? (see `find_next_hole`)

```suggestion
  List.fold_until
```

CHANGES.md
```diff
@@ -3,6 +3,7 @@
 ## Features
 
 - Make `inlay-hint` for function parameters configurable (#1515)
+- Add custom `ocamllsp/jumpTypedHole` to navigate through typed holes (#1516)
```
voodoos, line 6: Suggestion...
```suggestion
- Add custom `ocamllsp/jumpToTypedHole` to navigate through typed holes (#1516)
```

ocaml-lsp-server/src/custom_requests/req_jump_typed_hole.ml
xvw: No, `Base` is not part of `Import` so it is not opened.

voodoos (approved):
Thanks, I agree that moving this logic to the server is useful: these position based tricks are frequent footguns so it's good that client implementers can just trust the server instead of re-implementing it.

PR #1518, opened 2025-04-11
Handle error cases in `definition_query`

xvw:
In order to have best diagnostics at the editor level.

### DIFF
--- a/ocaml-lsp-server/src/definition_query.ml
+++ b/ocaml-lsp-server/src/definition_query.ml
@@ -2,19 +2,23 @@ open Import
 open Fiber.O
 
 let location_of_merlin_loc uri : _ -> (_, string) result = function
-  | `At_origin -> Ok None
-  | `Builtin _ -> Ok None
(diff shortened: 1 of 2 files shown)


ocaml-lsp-server/src/definition_query.ml
```diff
@@ -2,19 +2,23 @@ open Import
 open Fiber.O
 
 let location_of_merlin_loc uri : _ -> (_, string) result = function
-  | `At_origin -> Ok None
-  | `Builtin _ -> Ok None
+  | `At_origin -> Error "Already at definition point"
+  | `Builtin s ->
+    Error
+      (sprintf
+         "%S is a builtin, and it is therefore impossible to jump to its definition"
```
voodoos, line 9: That's a bit long isn't it ?
```suggestion
         "%S is a builtin, it is not possible to jump to its definition"
```

voodoos (approved):

PR #1528, opened 2025-05-07
Add ocamlformat mlx

davesnx:


ocaml-lsp-server/src/document.ml
```diff
@@ -436,13 +440,18 @@ let get_impl_intf_counterparts m uri =
     in
     match Syntax.of_fname fname with
     | Dune | Cram -> []
+    (* TODO: Unsure about this, keeping it empty for now *)
+    | Mlx ->
+      (match kind with
+       | Intf -> [ re; ml; mly; mll ]
```
andreypopp, line 446:
```suggestion
       | Intf -> [ re; ml; mly; mll; mlx ]
```
as we want to be able to switch to .mlx from .mli if it is present

voodoos: But here we are matching on an "Mlx Intf", which is not an `mli` file, right ? I am not sure that case happens at all. But nothing really bad could happen from having too many of such rules...

andreypopp: looking at the surrounding code, I thought the `kind` belongs to the origin, so we are on interface file and want to switch to the implementation, which can be .re, .ml, .mly, mll, and now .mlx

ocaml-lsp-server/src/document.ml
```diff
@@ -436,13 +440,18 @@ let get_impl_intf_counterparts m uri =
     in
     match Syntax.of_fname fname with
     | Dune | Cram -> []
+    (* TODO: Unsure about this, keeping it empty for now *)
```
andreypopp, line 443: not clear what's the TODO about

davesnx:
Embarrassing, I would push this forward. Thanks for the review @voodoos.

I wanted to test this PR manually before merging as well, is it as easy as pinning to ensure everything works or do I need something else?

voodoos:
> I wanted to test this PR manually before merging as well, is it as easy as pinning to ensure everything works or do I need something else?

I would expect so, please report if it doesn't :-)

voodoos:
There are still two places int he code were the build fails because of missing match cases.

davesnx:
Yes, tested manually in https://github.com/davesnx/html_of_jsx/pull/32 and works great

voodoos (approved):
Thanks @davesnx looks good to me ! Just a small fix needed in the changelog

PR #1529, opened 2025-05-19
Delegate outline generation to Merlin

Tim-ats-d:
Replacement of the in-house solution for generating outlines by Merlin.

I had to regenerate the tests because merlin and ocaml-lsp don't calculate the positions of elements in the outline in the same way. Fir instance, the `y` loc was the enclosing location for ocaml-lsp:
```ocaml
let x = object
  method y = 10
(*^^^^^^^^^^^^^*)  
end
```
While Merlin indicates the exact position:
```ocaml
let x = object
  method y = 10
      (* ^ *)  
end
```

The PR diff also indicates that several items `kind` value have changed from 7 to 13. Indeed, Merlin [treats instance variables as `Values`](https://github.com/ocaml/merlin/blob/018ad037e4646efe0fc3f091463ebbdb2947516a/src/frontend/query_protocol.ml#L57) and not as `Property` like [ocaml-lsp does](https://github.com/ocaml/ocaml-lsp/blob/master/lsp/src/types.mli#L77). This change simply means that the icon displayed on the vscode side will be different

Could you please mark this PR has draft? We have to wait for the next Merlin next because we need this patch https://github.com/ocaml/merlin/pull/1936


xvw (approved):
What a nice PR! Thanks!

xvw:
I guess that the PR can be `undrafted` yes?

PR #1545, opened 2025-07-16
Refactor extract custom query

Tim-ats-d:
Custom query for the corresponding Merlin command https://github.com/ocaml/merlin/pull/1948.


ocaml-lsp-server/src/custom_requests/req_refactor_extract.ml
```diff
@@ -0,0 +1,91 @@
+open Import
+
+let capability = "refactorExtract", `Bool true
```
xvw, line 3: Here, it should be `handleRefactorExtract`

PR #1547, opened 2025-07-21
Add wrap-type-in-module code action

fprasx:
Adds a code action that takes type declarations and wraps them in a module that includes a type  `t`.


rgrinberg:
@xvw did merlin end up gaining this feature?

Issue #1552, opened 2025-09-01
hover on method call replies "No information available"


trefis:
Tests were run from neovim, the exact error message might be different from other editors.

I tried with merlin on the command line (command `type-enclosing`) and got the expected result (the type of the method).

---
PS: ocamllsp v1.23.0 on ocaml 5.3.0

voodoos:
I was able to reproduce, that's quite surprising...

PR #1553, opened 2025-09-09
Fix hover on method calls

voodoos:
Fixes https://github.com/ocaml/ocaml-lsp/issues/1552

Now object method types are correctly shown.

The issue was due to a missing case in the hover filtering logic (cc @awilliambauer).

### DIFF
--- a/ocaml-lsp-server/test/e2e-new/hover.ml
+++ b/ocaml-lsp-server/test/e2e-new/hover.ml
@@ -30,7 +30,8 @@ let f (o : <  g : int -> unit >) = o#g 4
     Fiber.return ()
   in
   Helpers.test source req;
-  [%expect {|
+  [%expect
+    {|
     {
       "contents": { "kind": "plaintext", "value": "int -> unit" },
       "range": {


trefis (commented):
Perhaps the "filtering" match could be made non-fragile to avoid such issues in the future (and so every situation is properly considered)?

Side note: I feel like this code kind of duplicates what the browse tree is used for in merlin...

voodoos:
> Side note: I feel like this code kind of duplicates what the browse tree is used for in merlin...

My feeling is that the goal of that code is distinct from the one of mbrowse. Merlin was designed around a protocol and a set of editors that require a deliberate action from users to call the type-enclosing command: thus, if the user asked, she probably expects an answer. In visual LSP-based editors, such as vscode, type-enclosing is called continuously, everywhere around the code, as soon as the cursor moves. The best effort strategy of Merlin can result in a lot of clutter, and I think it makes sense to rely on stricter filtering before answering. (I do agree that it could be worth investigating how to re-conciliate both usages with the same code paths internally.)

It's also true that the opt-in nature of the current implementation make it sensible to syntax changes, additions (and omissions such as this one). However it is not a trivial task to reverse this behavior, and outside the scope of this PR.

Meanwhile, we might want to check that other parsetree constructions where an ident is not actually represented by a `Pexp_ident` node are not suffering from the same omission...

xvw (approved):
I agree with the fix
(Even, I think that, in a future, providing `hover` as part of the Merlin protocol make sense).

PR #1559, opened 2025-10-04
Enable support for OCaml 5.4 and prepare release 1.24.0

voodoos:


voodoos:
Tests pass on my machine but ocamlformat and bisect_ppx are not yet released for 5.4.

PR #1564, opened 2025-10-14
Add autocompletion for labelled args

lessp:
This adds auto-completion for labelled arguments. I decided to disable them for primitives like `int` or `bool` as it felt a bit iffy, but it'd work for those too. For int merlin seems to suggest `0` etc.

## Examples

Here's an example with variants:

https://github.com/user-attachments/assets/2e27725e-1907-4658-b3bd-7a991374cb9e

And records:

https://github.com/user-attachments/assets/094254fc-c4a2-472b-b7fa-f7c2964cfc33

### DIFF
--- a/ocaml-lsp-server/src/compl.ml
+++ b/ocaml-lsp-server/src/compl.ml
@@ -273,8 +273,18 @@ let is_primitive_type argument_type =
   List.mem primitives argument_type ~equal:String.equal
 ;;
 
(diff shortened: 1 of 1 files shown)


xvw:
Hi ! 
I totally agree with one, but I think that `construct` is usually _to much used as an hack entry door_ so I guess that just relaying on completion is sufficient?

PR #1565, opened 2025-10-16
Improve precision of duration field in `view-metrics`

Tim-ats-d:
Previously, current time was gathered with `Unix.time` which has a resolution of one second. Since each LSP request responds take less than one second to be computed, the `"dur"` filed is always set to `0` in reported metrics.
This patch uses `Unix.gettimeofday` to compute more precise timestamps.

cc @voodoos

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+## Features
+
+- Make `code-lens` for nested let bindings configurable (#1567)
+
 ## Fixes
 
 - Improve precision of collected metrics timestamps. (#1565)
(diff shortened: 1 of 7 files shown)


CHANGES.md
```diff
@@ -4,6 +4,10 @@
 
 - Support for OCaml 5.4 (#1559)
 
+## Fixes
+
+- Improve precision of collected metric timestamps. (#1565)
+
```
voodoos: This is not the correct changelog section, you should create a new "unreleased" section.

voodoos (approved):
LGTM but the changelog is wrong.

Issue #1561, opened 2025-10-06
[code lenses] really annoyed by code lenses on internal let-bindings


correnson:
Since version 1.23.0 there are extended code lenses on internal let-bindings of function, in additional to those for top-level definitions.
Personally, I really appreciate lenses for top-level definitions.
But internal ones really clutter my understanding of the code!

I'm speaking about this in 1.23.0 :

<img width="199" height="136" alt="Image" src="https://github.com/user-attachments/assets/45d4b2e3-140b-458e-b817-f54b00bc9d8f" />

Versus this in 1.22.0 :

<img width="182" height="101" alt="Image" src="https://github.com/user-attachments/assets/8c034bb5-2bda-405f-95b2-c99527b12cae" />

It would be *awesome* to be able to deactivate *internal* code lenses from VSCode settings.

voodoos:
Thank you for taking the time to report that @correnson. That sounds very legitimate. I would even suggest having these code lenses opt-in for the editors that won't benefit from this option.

It should be fairly easy to implement by looking at other configuration options, if someone wants to have a look ?

PR #1567, opened 2025-10-22
Make `code-lens` for toplevel let binding configurable

Tim-ats-d:
This PR adds an `only_toplevel` field to CodeLens settings, allowing CodeLens to display only for toplevel `let` bindings.

Solving https://github.com/ocaml/ocaml-lsp/issues/1561

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-## Fixes
+## Features
 
 - Make `code-lens` for nested let bindings configurable (#1567)
 


xvw (changes requested):
Can you add some `e2e-new` test to validate the behaviour? Thanks!

ocaml-lsp-server/docs/ocamllsp/config.md
```diff
@@ -15,12 +15,21 @@ interface config {
...
+    * Enable/Disable CodeLens
+    * @default false
+    * @since 1.16
+    */
+    enable : boolean,
+
+    /**
+    * Enable CodeLens only for toplevel let binding
+    * @default false
+    * @since 1.24
```
xvw: since `1.25` (I guess)

voodoos (changes requested):
Sorry to be that guy, but I think the choice of option is unnecessary restrictive here. What if one day we add lenses for other things and want to have it finely configurable ? Then with an option named "only_toplevel" we are stuck.

Let's make it additive instead (there might be a better name for it): "for_nested_bindings" with a default to false.

Does that make sense ?

Tim-ats-d:
> The semantic of the option itself is too restrictive.
Okay I think I see what you mean. I've changed the semantic of the option in my last commit. What do you think?

PR #1568, opened 2025-10-27
Rename parameter used to configure nested bindings

Tim-ats-d:
Previously the parameter was named with underscore (`for_nested_bindings`) which cause a malformed displaying in the VSCcode options panel. This PR renames it `forNestedBindings` for the sake of uniformity with other parameters.


voodoos:
Thanks !

Also, I think we want a capability to ensure good support from the client side, wdyt @xvw

PR #1573, opened 2025-11-25
Add instructions to readme for installing with dune pkg

gridbugs:


xvw (approved):
Thanks! Is the version with `eval (dune env)` is already released?

gridbugs:
Good catch, no it's not! I didn't realise that. This command will be added in dune.3.21.0 which should be released in the next few weeks.

xvw:
> Heads up that the documentation [here](https://github.com/tarides/ocaml-eglot?tab=readme-ov-file#usage-with-dune-pkg) for ocaml-eglot suggests using `dune tools env`.

Yes but there is a clear mention of the needed version: 

> If you are using [dune for package management](https://dune.readthedocs.io/en/stable/explanation/package-management.html) on the [latest nightly bulid](https://preview.dune.build/) or version 3.21

I think your instructions deserve this mention of version number too :)

PR #1575, opened 2025-12-15
Update config.md

zaclegarssure:
Update `ocaml-lsp-server/docs/ocamllsp/config.md` to match `ocaml-lsp-server/src/config_data.ml`.


voodoos:
Thank you @zaclegarssure ! Looks good to me.

PR #1576, opened 2025-12-16
Add some missing custom request

### LINKED ISSUE #1570: [protocol] add an optional parameter for looking for a given identifier for `definiton/declaration` request
At present, the `find definition/declaration` function only allows searching by position but Merlin is perfectly capable of searching for identifiers (`locate-ident` or `occurrences` with `prefix`). 

Even though it's not particularly protocol-compliant, I suggest (on behalf of @catern, https://github.com/tarides/ocaml-eglot/pull/72#discussion_r2495462145) adding an optional field in the `references` request to pass a prefix.

Since the parameter is optional, it does not disrupt the usual functioning of the query and allows Emacs to have a more refined `xref` backend that does not rely on tunneling (`merlinCallCompatible`).

xvw:
This PR adds three custom request: 
- `ocamllsp/typeExpression`
- `ocamllsp/phrase`
- `ocamllsp/locate` (fix #1570)

Adding an optional parameter to definition/declaration/type-definition search queries posed problems because the parameter types are generated by the Cinaps preprocessor, hence the addition of a custom request (based on the LSP request code).

### DIFF
--- a/ocaml-lsp-server/test/e2e-new/locate.ml
+++ b/ocaml-lsp-server/test/e2e-new/locate.ml
@@ -16,12 +16,32 @@ module Util = struct
     Client.request client req
   ;;
 
+  let sanitize_path = function
+    | `List ranges ->
+      `List
+        (List.map
+           ~f:(function
+             | `Assoc [ range; ("uri", `String value) ] ->
+               let new_value =
+                 value
+                 |> String.split_on_char ~sep:'/'
+                 |> List.last
+                 |> function
+                 | None -> "unknown"
+                 | Some x -> "file:///" ^ x
+               in
+               `Assoc [ range; "uri", `String new_value ]
+             | x -> x)
+           ranges)
+    | x -> x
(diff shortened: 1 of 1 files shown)


ocaml-lsp-server/src/custom_requests/req_type_expression.ml
```diff
@@ -0,0 +1,64 @@
...
+
+let make_type_expr_command position expression =
+  Query_protocol.Type_expr (expression, position)
+;;
+
+let dispatch_type_expr position expression pipeline =
+  let position = Position.logical position in
+  let command = make_type_expr_command position expression in
+  let result = Query_commands.dispatch pipeline command in
+  `String result
```
voodoos: Shouldn't we try to format the result with ocamlformat's rpc ?

xvw: Here it make sense :)

PR #1574, opened 2025-11-26
Update 4.14 branch with master

davesnx:
This PR brings additional features from master to work with OCaml 4.14, building on top of the [414-LTS-upgrade PR #1536](https://github.com/ocaml/ocaml-lsp/pull/1536).

### Cherry-picked commits (on top of 414-LTS)

- `df3a9a82` [#1534](https://github.com/ocaml/ocaml-lsp/pull/1534) Yojson 3 compat
- `f71ba500` [#1553](https://github.com/ocaml/ocaml-lsp/pull/1553) Fix hover on method calls
- `3226c927` [#1567](https://github.com/ocaml/ocaml-lsp/pull/1567) Make code-lens for toplevel let binding configurable
- `db3c0ea5` [#1565](https://github.com/ocaml/ocaml-lsp/pull/1565) Improve precision of duration field in view-metrics
- `d73c8f95` [#1568](https://github.com/ocaml/ocaml-lsp/pull/1568) Rename parameter used to configure nested bindings
- `5f4308f2` [#1528](https://github.com/ocaml/ocaml-lsp/pull/1528) Add ocamlformat-mlx and most OCaml features for .mlx 
- `8fb19aca` [#1563](https://github.com/ocaml/ocaml-lsp/pull/1563) More natural fallback on .merlin files when no dune-project is found. 
- `19a5c2dc` [#1400](https://github.com/ocaml/ocaml-lsp/pull/1400) refactor: get rid of stdune from code generator
- `e8d57f6b` [#1404](https://github.com/ocaml/ocaml-lsp/pull/1404) refactor: remove [Stdune.Io]
- `6d5aaa92` [#1480](https://github.com/ocaml/ocaml-lsp/pull/1480) refactor: remove uses of [Stdune.List]

### Merlin-lib 4.18 compatibility fix

The final commit (`976749cd`) fixes compatibility with merlin-lib 4.18:

- Restored local `Typed_hole` module with `can_be_hole` and `is_a_hole` functions (merlin-lib 4.18 doesn't have `Merlin_analysis.Typed_hole`)
- Removed `hint_function_params` from `Inlay_hints` (6th argument not available in merlin-lib 4.18)
- Uses 5-argument form of `Query_protocol.Inlay_hints`

PR #1577, opened 2025-12-20
414 LTS Upgrade

voodoos:
Supplants https://github.com/ocaml/ocaml-lsp/pull/1574

Several fixes require changes to Merlin. Unless there is a good reason not to I suggest postponing the release of the LTS version until that happens.


rgrinberg:
Please re-open if this is still relevant.

PR #1579, opened 2025-12-26
use correct -short-paths config for merlin

### LINKED ISSUE #1395: Cannot disable short-paths
As stated in issue #1227, short-paths usually gives weird behaviors for code using type aliases.

In the OCaml compiler, this issue can sometimes be mitigated by not using the `-short-paths` option to force displaying full paths in error messages. Such types are usually longer but at least they are similar to the declared function signature.

With OCaml LSP, such a mitigation seems inaccessible with the current implementation. The `-short-path` option is always enabled by default. If I understand correctly, passing the `real_path` option to the underlying Merlin should disable this path shortening and give a behavior that is similar to the compiler without the `-short-paths` option.

Ideally, I would like to be able to use OCaml LSP with Merlin in `real_paths` mode for some projects, and in the current `short_paths` mode for others.

SkySkimmer:
Fix #1395

The default needs to be `real_paths = true` (from `Mconfig.initial`), because `get_external_config` (called in `config`) will modify the config from the ocaml flags which will not contain a negated `-short-paths` when short paths are not wanted.

Should backport cleanly to 414-LTS (or rather I tested on 414-LTS then moved the commit on top of master ;))

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,7 @@
 ## Fixes
 
 - Enable "hover" on more locations, notably class and object related (#1599)
+- do not force `-short-paths` (#1579)
 
 # 1.25.0
 
--- a/ocaml-lsp-server/src/merlin_config.ml
+++ b/ocaml-lsp-server/src/merlin_config.ml
@@ -314,8 +314,7 @@ let create db path =
     let filename = Filename.basename path in
     let init = Mconfig.initial in
     { init with
-      ocaml = { init.ocaml with real_paths = false }
-    ; query = { init.query with filename; directory; verbosity = Mconfig.Verbosity.Smart }
+      query = { init.query with filename; directory; verbosity = Mconfig.Verbosity.Smart }
     }
   in
   { path; directory; initial; db; entry = None }


SkySkimmer:
Not sure how adding tests works here but a pair of tests (with / without short paths) should probably be added.

voodoos (commented):
Thanks ! It's not unreasonable, and would make ocaml-lsp's default behavior the same as merlin's.

I re-read Merlin's configuration processing, and I think I have a way to disable short-paths without modifying the default behavior here:

```
(env
 (_
  (flags ((:standard \ -short-paths) -real-paths))))
```

Could you try it ? Would you be happy with such a solution ? I am a bit hesitant to change the default behavior here, but we do need it to be configurable.

voodoos (approved):
Let's move back to the same defaults as the rest of the stack.

@SkySkimmer  can you rebase and add a changelog entry ?

SkySkimmer:
ping @voodoos

SkySkimmer:
ping @voodoos

voodoos:
Thanks @SkySkimmer, and sorry for the latency !

PR #1580, opened 2026-01-08
Improving the readability of signature_help for long signatures

Lucccyo:
On long signatures in the response, signature-help might be hard to read.
This PR applies `ocamlformat` to the long signatures to ensure better readability.


Lucccyo:
Why should we ignore labels ?

ocaml-lsp-server/src/signature_help.ml
```diff
@@ -63,19 +63,56 @@ let run (state : State.t) { SignatureHelpParams.textDocument = { uri }; position
...
+           if application_signature.signature = formatted_signature
+           then
+             List.map
+               application_signature.parameters
+               ~f:(fun (p : Merlin_analysis.Signature_help.parameter_info) ->
+                 let label = `Offset (offset + p.param_start, offset + p.param_end) in
+                 ParameterInformation.create ~label ())
+           else (
+             let explode s = List.init (String.length s) ~f:(fun e -> String.get s e) in
+             let rec parse_formatted_parameters pdepth acc strl res c : (int * int) list =
```
voodoos: Your function have four different accumulators: pdepth, acc, res and c. 

it would make the function easier to read and understand to have less cryptic names for these and maybe a short comment stating their purpose.

PR #1581, opened 2026-01-11
Allow to configure merlin config server

andreypopp:
Before this commit it was only possible to configure ocamllsp either with .merlin files or with 'dune ocaml-merlin' command which starts a configuration server.

This commit extends ocamllsp to allow to be configured with merlin config server, other than `dune ocaml-merlin`.

This is done by reading the following environment variables:
- `$OCAMLLSP_PROJECT_BUILD_SYSTEM` specifies the project's build system, which should be an executable available on $PATH. The convention is there's a subcommand `$OCAMLLSP_PROJECT_BUILD_SYSTEM ocaml-merlin` which starts merlin configuration server.
- `$OCAMLLSP_PROJECT_ROOT` specifies project root
- in the absence of the former, `$OCAMLLSP_PROJECT_ROOT_MARKERS` (a list of `,` comma separated filenames) is read and used to find a project root by checking if a dir (from the current one up the hierarchy) contains one of the markers

In absence of these new environment variables ocamllsp falls back to the current behaviour.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,16 @@
 
 - Add `destruct` custom request (#1583)
 
+- Allow to configure merlin using build systems other than dune.
+
+  If `$OCAMLLSP_PROJECT_BUILD_SYSTEM` environment variable is set then OCaml
+  LSP will execute `$OCAMLLSP_PROJECT_BUILD_SYSTEM ocaml-merlin` command to get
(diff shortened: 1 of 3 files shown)


rgrinberg (approved):

xvw (approved):
But I think it deserve a CHANGE entry ! 
Thanks !

PR #1600, opened 2026-04-03
Update changelog for 1.26.0-504

PizieDust:
cc @voodoos


voodoos:
Thanks @PizieDust We also need to update the constraints on merlin-lib before releasing 😉

Issue #1602, opened 2026-05-04
The compilation of ocaml-lsp-server.1.26.0-5.5~preview failed (because of merlin?)


juliancoffee:
Well, I tried to follow the instructions in README.md, I did update submodules with recursive flag, but the moment I get to `opam switch --yes create .`, it fails with the following error. Cropped, to, I believe, relevant output.
```
[ERROR] The compilation of ocaml-lsp-server.1.26.0-5.5~preview failed at "dune build -p
        ocaml-lsp-server -j 7 @install".

#=== ERROR while compiling ocaml-lsp-server.1.26.0-5.5~preview ================#
# context     2.5.0 | macos/arm64 | ocaml.5.4.1 | pinned(git+file:///Users/illiadenysenko/Workspace/repos/ocaml-lsp#master#63e16a5353c80203ee63a933d3debde0bb0e1e37)
# path        ~/Workspace/repos/ocaml-lsp/_opam/.opam-switch/build/ocaml-lsp-server.1.26.0-5.5~preview
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p ocaml-lsp-server -j 7 @install
# exit-code   1
# env-file    ~/.opam/log/ocaml-lsp-server-47606-bbfb82.env
# output-file ~/.opam/log/ocaml-lsp-server-47606-bbfb82.out
### output ###
# [...]
# submodules/lev/lev/vendor/ev.c:1734:3: warning: unused function 'ecb_float_to_binary16' [-Wunused-function]
#  1734 |   ecb_float_to_binary16 (float x)
#       |   ^~~~~~~~~~~~~~~~~~~~~
# submodules/lev/lev/vendor/ev.c:1742:3: warning: unused function 'ecb_binary16_to_float' [-Wunused-function]
#  1742 |   ecb_binary16_to_float (uint16_t x)
#       |   ^~~~~~~~~~~~~~~~~~~~~
# 63 warnings generated.
[...]

PR #1603, opened 2026-05-05
Pin merlin in ocaml-lsp-server.opam.template

juliancoffee:
Github issue: #1602

So, basically, it moves the CI merlin pin to `ocaml-lsp-server.opam.template`, and removes the dune command, which was copying `lsp.opam.template`, instead doing it manually.

But I guess it's not enough, since both CI and local switch creation fail ^^'

UPD: ok, I just needed to run dune promote and update `ocaml-lsp-server.opam` file, I guess, not just the template ^^'

Does this need a changelog too?

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -47,10 +47,6 @@ jobs:
         with:
           ocaml-compiler: "5.4"
 
-      # Remove this pin once a compatible version of Merlin has been released
-      - name: Pin dev Merlin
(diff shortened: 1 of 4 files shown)


voodoos:
I see that you are in the CI Death Loop©️

Don't worry too much if it continues to fail, I will have a look at some point !

voodoos:
Well done 👍

voodoos:
Thanks you @juliancoffee it's much more robust than that. (But we now need to think about removing the pin when releasing.)

Issue #717, opened 2022-06-11
Add hover support for warning attributes


ulugbekna:
A user hovering over `[@warning "-27"]` could see that this attribute is disabling warning 27 and description of what that warning is

xvw:
To achieve this, you would need to refer to the function integrated via this issue: https://github.com/ocaml/merlin/issues/1775 for linking between warning id and warning description.

PR #1606, opened 2026-05-26
Hover on warnings

PizieDust:
This PR attempts to solve https://github.com/ocaml/ocaml-lsp/issues/717 and requires https://github.com/ocaml/merlin/pull/2088 to work


ocaml-lsp-server/src/hover_req.ml
```diff
@@ -13,6 +13,97 @@ let environment_mode =
...
+  | Disable of int
+  | Enable_as_error of int
+  | Enable_range of int * int
+  | Disable_range of int * int
+  | Enable_as_error_range of int * int
+  | Enable_letter of char
+  | Disable_letter of char
+  | Enable_as_error_letter of char
+
+let parse_warning_payload s =
```
xvw: This function is a little bit stressful. Could we imagine some unit-tests for the specific function? Thanks :)

voodoos: I suspect there is a similar function somewhere in the compiler. We should reuse it!

PizieDust: We have this function `parse_options` https://github.com/ocaml/ocaml/blob/trunk/utils/warnings.ml#L876 which is the function exposed in the compiler library interface (Warnings.mli). This function calls `parse_opt` parser which scans the string and mutates the compiler's warning flags

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -306,11 +365,13 @@ let format_type_enclosing
...
          in
-         print_dividers (List.filter_opt [ type_info; syntax_doc; doc ])
+         print_dividers (List.filter_opt [ type_info; syntax_doc; doc; warnings_doc ])
        in
        { MarkupContent.value; kind = MarkupKind.Markdown })
      else (
-       let value = print_dividers (List.filter_opt [ Some typ; syntax_doc; doc ]) in
+       let value =
+         print_dividers (List.filter_opt [ Some typ; syntax_doc; doc; warnings_doc ])
+       in
```
voodoos, line 353: This, and the related code, is unrelated to the initial goal of this PR which was to implement the feature suggested in https://github.com/ocaml/ocaml-lsp/issues/717

Unlike #717 adding this information to the hover query when over a warnings' loc has not been discussed and IMO need more design time and discussion: 
- Hovers are already severely overloaded by ocaml-lsp-server.
- Editors already show the short warning description on hover
- There are other ways we could decide to add more infos to the diagnostics: by adding it to the message, or via the related_information field, or by having a code action on highlighted warnings using the data field, etc.

I suggest moving this ot another branch and opening an issues to discuss options for this feature.

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -524,6 +632,19 @@ let handle server { HoverParams.textDocument = { uri }; position; _ } mode =
...
+           let client_capabilities = State.client_capabilities state in
+           let markdown_support =
+             ClientCapabilities.markdown_support client_capabilities ~field:(fun td ->
+               Option.map td.hover ~f:(fun h -> h.contentFormat))
+           in
+           if markdown_support
+           then `MarkupContent (MarkupContent.create ~kind:Markdown ~value:markdown)
+           else `MarkedString { Lsp.Types.MarkedString.value = markdown; language = None }
+         in
+         let range = None in
```
voodoos: I think we want the attribute's position here.

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -243,6 +274,32 @@ let hover_at_cursor parsetree (`Logical (cursor_line, cursor_col)) =
...
+                       ; _
+                       }
+                     , _ )
+               ; _
+               }
+             ] ->
+           let actions = Ocaml_utils.Warnings.parse_warnings payload in
+           let markdown_lines = List.map ~f:format_warning_action actions in
+           let markdown = String.concat ~sep:"\n" markdown_lines in
+           if markdown <> "" then result := Some (`Warning_attribute markdown)
```
voodoos: The spirit of the hover_at_cursor function is to do the triaging before processing. This should be moved later and we should just use `'Warning_attribute` to transmit the necessary information for furture processing, that is the raw payload and the loc.

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -243,6 +274,32 @@ let hover_at_cursor parsetree (`Logical (cursor_line, cursor_col)) =
...
+         | _ -> Ast_iterator.default_iterator.attribute self attr)
+      | _ -> Ast_iterator.default_iterator.attribute self attr)
+    else Ast_iterator.default_iterator.attribute self attr
```
voodoos:
```suggestion
         | _ -> ())
      | _ -> ();
    Ast_iterator.default_iterator.attribute self attr
```

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -243,6 +274,32 @@ let hover_at_cursor parsetree (`Logical (cursor_line, cursor_col)) =
       result := Some `Type_enclosing
     | _ -> Ast_iterator.default_iterator.signature_item self item
   in
+  let attribute (self : Ast_iterator.iterator) (attr : Parsetree.attribute) =
+    if is_at_cursor attr.attr_name.loc || is_at_cursor attr.attr_loc
```
voodoos: Why test both ? If you want the popup to appear only when the cursor is on the attribute name you should remove the second case. If not remove the first one.

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -243,6 +274,32 @@ let hover_at_cursor parsetree (`Logical (cursor_line, cursor_col)) =
...
+                   Pstr_eval
+                     ( { pexp_desc =
+                           Pexp_constant
+                             { pconst_desc = Pconst_string (payload, _, _); _ }
+                       ; _
+                       }
+                     , _ )
+               ; _
+               }
+             ] ->
```
voodoos: This looks like `Ast_helpers.extract_str_payload`

voodoos (changes requested):

ocaml-lsp-server/src/hover_req.ml
PizieDust: Thanks. This should be done in the latest commit. Is there more needed for this PR?

PR #1608, opened 2026-06-20
Avoid Stdune.Pid for process ids

anmonteiro:
Avoid using `Stdune.Pid` for process ids, matching the Dune API change in ocaml/dune@39d8a28027768119a3ced17aa8d28a187891124f.


rgrinberg:
How about we stop using this module altogether in lsp? or is that non-trivial?

anmonteiro:
made it an `int`, do you think that's a good compromise here?

PR #1612, opened 2026-06-30
[codex] Use string for diagnostic messages

aspeddro:
## Summary

This changes LSP diagnostics so `Diagnostic.message` is represented as a plain string again instead of accepting either a string or markup content.

Callers that inspected diagnostic messages now use the string directly, and diagnostic construction sites no longer wrap messages in `` `String``.

See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic

## Impact

This aligns the typed API with the expected diagnostic message shape and removes unnecessary markup handling in diagnostic consumers and tests.

## Validation

- `make check`

### DIFF
--- a/lsp/src/types.ml
+++ b/lsp/src/types.ml
@@ -16865,34 +16865,13 @@ module CodeDescription = struct
 end
 
 module Diagnostic = struct
-  type message_pvar =
-    [ `String of string
-    | `MarkupContent of MarkupContent.t
-    ]
-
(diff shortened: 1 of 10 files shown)


rgrinberg:
markdown is allowed since 3.18 https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#diagnostic

PR #1613, opened 2026-07-04
Fix JSON format typos

divarvel:
quick question wrt

> A  response with null result is returned if no entries are found.

is this a LSP thing? why not return an empty list in that case?


voodoos:
Thanks @divarvel.
@xvw can correct me, but since this is a custom request it's not subject to the LSP protocol, not sure why this choice was made in the first place.

xvw:
I can't remember. 
It was maybe because my goal was to fit the same behaviour or `polarity search`? Btw, I think that we can merge this. Thanks.

PR #1615, opened 2026-07-07
Use merlins ppx expand command

### LINKED ISSUE #1614: Broken ppx deriving expansion
On OCaml 5.5.0 with ocaml-lsp 1.27.0 and the following code
```ocaml
type mrec_variant =
  | MrecFoo of string
  | MrecBar of int

and mrec_variant_list = mrec_variant list [@@deriving eq]
```
from https://github.com/ocaml-ppx/ppx_deriving/blob/c99dd3c44437669d4234df0eb9d0a5c9a7e95bac/src_test/eq/test_deriving_eq.cppo.ml#L77-L81.

The hover on `@@deriving` shows the following broken expansion:
```ocaml
(* ppx deriving expansion *)
;;eq
let _ = fun (_ : mrec_variant_list) -> ()
;;"-39"
;;"-A"
open! ((Ppx_deriving_runtime)[@ocaml.warning "-A"])[@@ocaml.warning "-A"]
;;"-39"
```

The actual preprocessed document contains
```ocaml
include
  struct
    let _ = fun (_ : mrec_variant) -> ()
    let _ = fun (_ : mrec_variant_list) -> ()
    let rec (equal_mrec_variant :
              mrec_variant -> mrec_variant -> Ppx_deriving_runtime.bool)
      =
      ((let open! ((Ppx_deriving_runtime)[@ocaml.warning "-A"]) in
          fun lhs rhs ->
            match (lhs, rhs) with
            | (MrecFoo lhs0, MrecFoo rhs0) ->
                ((fun (a : string) b -> a = b)) lhs0 rhs0
            | (MrecBar lhs0, MrecBar rhs0) ->
                ((fun (a : int) b -> a = b)) lhs0 rhs0
            | _ -> false)
      [@ocaml.warning "-39"][@ocaml.warning "-A"])[@@ocaml.warning "-39"]
    and (equal_mrec_variant_list :
          mrec_variant_list -> mrec_variant_list -> Ppx_deriving_runtime.bool)
      =
      ((let __0 = equal_mrec_variant in
        ((let open! ((Ppx_deriving_runtime)[@ocaml.warning "-A"]) in
            fun x ->
              (let rec loop x y =
                 match (x, y) with
                 | ([], []) -> true
                 | (a::x, b::y) -> (__0 a b) && (loop x y)
                 | _ -> false in
               fun x y -> loop x y) x)
          [@ocaml.warning "-A"]))
      [@ocaml.warning "-39"])[@@ocaml.warning "-39"]
    let _ = equal_mrec_variant
    and _ = equal_mrec_variant_list
  end[@@ocaml.doc "@inline"][@@merlin.hide ]
```

PizieDust:
fixes https://github.com/ocaml/ocaml-lsp/issues/1614

This PR refactors ppx hover functionality to use the output of Merlin's `expand-ppx` command rather than trying to construct what the generated ppx should be.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+## Features
+ - Refactor ppx hover to use merlin directly (#1606)
+
 # 1.27.0
 
 ## Features
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 # Unreleased
 
-## Features
- - Refactor ppx hover to use merlin directly (#1606)
+## Fixes
+ - Refactor ppx hover to use merlin directly (#1606, fixes #1614)
 
 # 1.27.0
 
--- a/ocaml-lsp-server/src/hover_req.ml
+++ b/ocaml-lsp-server/src/hover_req.ml
@@ -109,8 +109,7 @@ let hover_at_cursor parsetree (`Logical (cursor_line, cursor_col)) =
         if is_at_keyword
         then result := Some `Type_enclosing
         else Ast_iterator.default_iterator.expr self expr
-      | Pexp_extension (ppx, _) when is_at_cursor ppx.loc ->
-        result := Some (`Ppx_expr ppx.txt)
+      | Pexp_extension (ppx, _) when is_at_cursor ppx.loc -> result := Some (`Ppx ppx.txt)
       | _ -> Ast_iterator.default_iterator.expr self expr)
   in
   (* Hover a value declaration in a signature *)
@@ -135,7 +134,7 @@ let hover_at_cursor parsetree (`Logical (cursor_line, cursor_col)) =
         List.find decl.ptype_attributes ~f:(fun attr -> is_at_cursor attr.attr_loc)
       in
       match attribute_at_cursor with
-      | Some attr -> result := Some (`Ppx_typedef_attr attr.attr_name.txt)
+      | Some attr -> result := Some (`Ppx attr.attr_name.txt)
       | None -> Ast_iterator.default_iterator.type_declaration self decl)
   in
   (* Classes and objects *)
@@ -419,7 +418,7 @@ let handle server { HoverParams.textDocument = { uri }; position; _ } mode =
            | Some _ | None -> false
          in
          type_enclosing_hover ~server ~doc ~merlin ~mode ~uri ~position ~with_syntax_doc
-       | Some ((`Ppx_expr _ | `Ppx_typedef_attr _) as ppx_kind) ->
+       | Some (`Ppx ppx) ->
          let+ ppxed_source =
            Document.Merlin.dispatch_exn
              ~name:"expand-ppx"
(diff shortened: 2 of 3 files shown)


PizieDust:
cc @voodoos 
will this need a changelog?

voodoos:
> cc @voodoos will this need a changelog?

Yes. And there is a change in the testsuite that needs checking.

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -135,11 +135,7 @@ let hover_at_cursor parsetree (`Logical (cursor_line, cursor_col)) =
         List.find decl.ptype_attributes ~f:(fun attr -> is_at_cursor attr.attr_loc)
       in
       match attribute_at_cursor with
-      | Some attr ->
-        (* Produce a hover for the attribute, if it's name is hovered, otherwise
-           bail. *)
-        if is_at_cursor attr.attr_name.loc
-        then result := Some (`Ppx_typedef_attr (decl, attr))
```
voodoos, line 142: Why did you remove the `is_at_cursor attr.attr_name.loc` check ?

ocaml-lsp-server/src/hover_req.ml
```diff
@@ -525,14 +420,24 @@ let handle server { HoverParams.textDocument = { uri }; position; _ } mode =
...
-         (match ppx_kind with
-          | `Ppx_expr (expr, ppx) -> ppx_expression_hover ~ppx_parsetree ~expr ~ppx
-          | `Ppx_typedef_attr (decl, attr) ->
-            typedef_attribute_hover ~ppx_parsetree ~decl ~attr)))
+         (match ppxed_source with
+          | `Found { Query_protocol.code; attr_start; attr_end } ->
+            let ppx =
+              match ppx_kind with
+              | `Ppx_expr ppx -> ppx
+              | `Ppx_typedef_attr ppx -> ppx
```
voodoos: Since both cases are treated the same there is no reason to keep both variants, we should replace it by a single `'Ppx` one. This should allow you to simplify the code a bit.

voodoos (commented):
Look good ! Just one removed check to check, the CI failure to investigate, and a small refactor.

ocaml-lsp-server/src/hover_req.ml
PizieDust: This was so that it only activates when the user hovers on the `deriving` keyword but now i think it should hover anywhere within the attribute and not just the attribute name

PizieDust: now we let merlin do the heavy lifting and give us where it thinks the ppx starts and ends

CHANGES.md
```diff
@@ -1,3 +1,8 @@
+# Unreleased
+
+## Features
+ - Refactor ppx hover to use merlin directly (#1606)
```
voodoos:
```suggestion
## Fixes
 - Refactor ppx hover to use merlin directly (#1606, fixes #1614)
```

voodoos:
Thanks ! I fixed the change section.

PR #1621, opened 2026-07-12
Restrict Dune formatting to Dune documents

smorimoto:
## Summary

- distinguish Dune documents from other non-Merlin documents in `Document.t`
- expose an abstract `Document.Dune.t` and require it throughout the Dune formatting path
- restrict whole-document OCaml formatting to `Document.Merlin.t`
- preserve OCamllex and Menhir semantic-action range formatting while returning no edits for unsupported requests
- add regression coverage for OCamllex, Menhir, Cram and Dune documents

## Motivation

`Document.kind` intentionally presents every non-Merlin document as `Other`. The formatting handler treated that compatibility view as a formatting classification and sent every `Other` document to Dune's S-expression formatter. As a result, OCamllex, Menhir and Cram documents could be misrouted through Dune RPC.

This change makes the Dune formatter's contract explicit in the type system, so non-Dune documents cannot be passed to it accidentally. Unsupported formatting requests now return no edits without a warning. Dune documents retain the existing `InvalidRequest` behaviour when no Dune RPC instance is available.

Related:

- https://github.com/ocamllabs/vscode-ocaml-platform/issues/1159
- https://github.com/ocamllabs/vscode-ocaml-platform/issues/1487

## Validation

- `make fmt`
- `make check`
- `make test`

### DIFF
--- a/ocaml-lsp-server/src/document.ml
+++ b/ocaml-lsp-server/src/document.ml
@@ -205,10 +205,15 @@ type merlin =
   ; kind : Kind.t option
   }
 
-type dune = Text_document.t
+module Dune = struct
+  type t = Text_document.t
+
+  let uri = Text_document.documentUri
+  let text = Text_document.text
+end
 
 type t =
-  | Dune of dune
+  | Dune of Dune.t
   | Other of
       { tdoc : Text_document.t
       ; syntax : Syntax.t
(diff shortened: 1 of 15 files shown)


ocaml-lsp-server/src/document.ml
```diff
@@ -205,21 +205,26 @@ type merlin =
   ; kind : Kind.t option
   }
 
+type dune = Text_document.t
```
rgrinberg: Do you need this alias? Why can't you move this `Dune` module before the definition of `t`?

smorimoto: Good point. I have removed the alias and reorganised the definitions by moving the `Dune` module before `t`; the constructor now uses `Dune.t` directly.

rgrinberg:
Thanks.

PR #1730, opened 2026-07-18
test(code-actions): reproduce extraction of scoped constructors

rgrinberg:
## Summary

- attempt function extraction around a local exception constructor
- record the currently emitted out-of-scope function

The expect output is promoted to the current faulty behavior so this test-only change passes CI. The production fix and changelog entry will follow separately.

## Testing

- `make test`


rgrinberg:
Closing this test-only PR because the existing `extract function with local exception` expect test already reproduces and documents the bug. This change only renamed and reformatted that test, so it does not add useful regression coverage. The production fix can update the existing expectation in its follow-up PR.

PR #1749, opened 2026-07-19
feature(lsp): update spec to 3.18

rgrinberg:
- Update the bundled meta-model from LSP 3.17 to the official LSP 3.18.0 model and regenerate the OCaml types.
- Add the 3.18 workspace text-document-content requests, named capability records, open language kinds, code-action tags/documentation, completion defaults/apply kinds, snippet edits, workspace-edit metadata, and related protocol additions.
- Follow the generated 3.18 API shapes, including named helper records, partial/whole-document change variants, and `LocationUriOnly` workspace symbols; downstream OCaml call sites are updated rather than compatibility-shimmed.
- Retain the wire-level `"compact"` trace extension and the document-selector/document-link URI protocol corrections.

### DIFF
--- a/lsp/bin/cinaps.ml
+++ b/lsp/bin/cinaps.ml
@@ -8,75 +8,27 @@ let preprocess_metamodel =
       match
         List.filter_map types ~f:(function
           | Metamodel.Literal (Record []) -> None
-          | _ as t -> Some (self#type_ path t))
+          | type_ -> Some (self#type_ path type_))
       with
       | [] -> assert false
-      | [ t ] -> t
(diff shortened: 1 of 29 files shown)


voodoos:
This should have a changelog entry doesn't it ?

rgrinberg:
I only maintain entries relevant to ocaml-lsp itself. I don't think this change has any user facing impact. Though I don't mind mentioning it regardless if you think users would care.

voodoos:
Ah I see, we don't have a changelog for the LSP package itself.

PR #1778, opened 2026-07-21
refactor(test): share e2e client lifecycle helpers

rgrinberg:
Factor the e2e lifecycle cleanup into four focused refactors:

- add an initialized runner and use it in the shared request helper;
- reuse initialized setup for tests with explicit or diagnostic-synchronized teardown;
- share the graceful `shutdown`-then-stop sequence;
- share the LSP `shutdown`-and-`exit` sequence.

Each migrated test retains its existing shutdown, exit, and diagnostic synchronization order. The raw runner remains available for lifecycle-sensitive tests.


rgrinberg:
Superseded by the split series, which will be submitted one PR at a time starting with #1780.

PR #1826, opened 2026-07-24
test(e2e): share ignored formatting fixture

rgrinberg:
Share the ignored-file OCaml source between full- and range-formatting tests. This also replaces the full-formatting fixture's accidental unmatched leading quote with the valid shared source.


rgrinberg:
Closing to keep this cleanup focused on duplicated test code rather than fixture data.

PR #1857, opened 2026-07-25
feat(semantic-highlighting): classify function parameters

rgrinberg:
Function parameters are currently highlighted as ordinary variables, and labeled and optional arguments lose their syntactic distinction.

Emit the standard `parameter` token type for parameter bindings and resolved references, including references in call position. Typed-tree identifier resolution distinguishes captured parameters from shadowing local bindings. Advertise custom `labeled` and `optional` modifiers for OCaml argument labels, and correct modifier decoding for the extended legend.

Retain syntax-based highlighting in ill-typed documents and fall back to variable/function tokens for unresolved references. Extend recovery coverage to patterns, guards, defaults, annotations, interfaces, shadowing, and incremental break/repair deltas. Avoid overlapping tokens for constrained labeled and optional parameter bindings.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,8 @@
 - Add a code action to open the closest Dune file for the current document.
   (#1817, fixes #1491, @rgrinberg)
 - Distinguish operators from functions in semantic highlighting. (#1831, @rgrinberg)
+- Classify function parameters and labels in semantic highlighting.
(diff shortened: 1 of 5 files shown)


rgrinberg:
Not entirely sure about this. It reaches for the typer to provide this highlighting.

Issue #1139, opened 2023-06-14
Feature request: restore some syntactic distinctions in semantic highlights


ddickstein:
I liked some of the distinctions that the vim syntax highlighting offered that are erased with semantic highlights:

- Operators are distinguished from functions
- Built-in types are identified separately
- Labeled vs. unlabled arguments in function definitions (would be nice to have this in declarations too)

None of these are really principled distinctions to draw, but I think people's aesthetic preferences for reading code don't align exactly with language semantics.

rgrinberg:
> Operators are distinguished from functions

We should be able to restore this one

> Built-in types are identified separately

This relied on hacks to determine the built in types. To do this reliably we would need to type check the buffer, but syntax highlighting today only relies on the parsetree for performance. We would have to confirm that this doesn't impact performance.

> Labeled vs. unlabled arguments in function definitions (would be nice to have this in declarations too)

What token types would you use to distinguish these?

dlesbre:
Another distinction that is lost is highlighting special character in strings (namely escapes `\n`, `\t` and printf `%d`...).
They used to be of a different color but aren't (since semantic highlighting set their token type to `string`, which is less precise than the syntactic token type `constant.character.printf` but has higher precedence):

![image](https://github.com/ocaml/ocaml-lsp/assets/53176929/4882057d-10be-4f87-9a46-fd448d2c3dc8)

GrayJack:
A distinction that I would like to see (besides the operator from function) is function parameters having the semantic token of `parameter`

> What token types would you use to distinguish these?

I'm not an expert on LSP, but I believe you can use custom tokens and token modifiers. Rust-analyzer (the only LSP server that I contributed), uses them in a few places, more on modifiers than tokens

So maybe all parameters have `parameter` token and `labeled` or `unlabeled` modifier depending on the case?

PR #1858, opened 2026-07-25
feat(semantic-highlighting): classify built-in types

rgrinberg:
Predeclared OCaml types are currently indistinguishable from user-defined types in semantic highlighting.

Mark references to compiler predeclared types with the standard `defaultLibrary` modifier. Resolve type-constructor paths through the typed tree rather than matching names, so a local type that shadows `int` remains an ordinary user-defined type.

Regression coverage was merged in #1849. The ill-typed-buffer coverage in #1877 additionally verifies that resolved built-ins retain the modifier across errors, unresolved types degrade to ordinary `type` tokens, and shadowed names are not marked. This is stacked on the parameter-classification change in #1857 and completes #1139.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,8 @@
 - Distinguish operators from functions in semantic highlighting. (#1831, @rgrinberg)
 - Classify function parameters and labels in semantic highlighting.
   (#1857, @rgrinberg)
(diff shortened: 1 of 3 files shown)


rgrinberg:
This change also relies on the type for highlighting. But given that the task is to classify built-in types, I don't think there's a way around that.

PR #1889, opened 2026-07-25
fix(completion): respect negotiated position encoding

rgrinberg:
Completion currently mixes LSP character positions with Merlin byte offsets and computes edit ranges using UTF-8 byte lengths. Convert request positions through the document and calculate completion, construct, and resolve ranges using the negotiated UTF-8 or UTF-16 encoding.


rgrinberg:
Superseded by #2011, which uses document-bound UTF-8 span conversion and keeps negotiated encoding out of completion feature code.

PR #1943, opened 2026-07-27
fix(lsp): escape snippet text

rgrinberg:
Plain snippet text currently emits `$`, `}`, and backslashes without escaping them, which can change how an editor parses the snippet.

Render text through the snippet escaping function and update the focused regression expectation to the valid serialization.


rgrinberg:
Closing this stacked PR so the remaining changes can be submitted directly against master.

PR #1948, opened 2026-07-27
fix(jsonrpc-fiber): permit notification batches after stop

rgrinberg:
Permit notification-only batches after the session stops accepting requests, while continuing to reject batches containing requests.

This fixes a regression introduced by #1938 that broke shutdown-time notification draining in the end-to-end server tests.


rgrinberg:
Superseded by the full revert in #1949 while we reconsider the stopped-session output semantics.

PR #1964, opened 2026-07-28
fix(completion): replace incorrect identifier suffixes

### LINKED ISSUE #838: Fix completion with incorrect suffix
We should support completion for `String.is_prefix` given `String.is_pr|word` (`|` is the cursor here).

First, we should check whether it's on the merlin or our side.

rgrinberg:
Extend completion text edits through the identifier suffix after the cursor. In the #838 reproduction, selecting `is_prefix` now replaces all of `is_prword` instead of leaving `word` behind.

Fixes #838.


rgrinberg:
This fix works in vscode but not neovim.

PR #1991, opened 2026-07-30
test(code-actions): reproduce destruct assertion on malformed input

rgrinberg:
Adds an expect-test reproducer for a destruct code-action request on `object method x with|0`.

A valid zero-width request at `(0,21)` currently returns JSON-RPC `InternalError` after reaching an assertion in Merlin's destruct analysis. The unstable assertion location and backtrace are censored. This is a test-only, test-first change for a follow-up fix.


rgrinberg:
cc @voodoos

PR #2013, opened 2026-07-31
fix(server): enable function param inlay hints without configuration

### LINKED ISSUE #1371: inlayHints config doc is out of date
https://github.com/ocaml/ocaml-lsp/blob/3dc1a6633bfe2ebd34d7aa38f008c8c28300ee55/ocaml-lsp-server/docs/ocamllsp/config.md?plain=1#L37

https://github.com/ocaml/ocaml-lsp/blob/3dc1a6633bfe2ebd34d7aa38f008c8c28300ee55/ocaml-lsp-server/src/config_data.ml#L5-L9

dayangac:
b7faaf23 (#1517) made function parameter inlay hints default to enabled, but it
only changed the `[@default true]` used when deserializing an `inlayHints`
object. `Config_data.default`, which applies when the client sends no
configuration at all, still had `hint_function_params = false`.

Clients that never send `workspace/didChangeConfiguration` therefore saw no
function parameter hints, while clients sending even an empty `inlayHints`
object saw them. `config.md` documented the pre-#1517 default, so it disagreed
with both paths. Set `hint_function_params` to `true` in `Config_data.default`
and update the documentation to match.

The inlay hint test helper always sends a configuration, so no test covered the
default path; add a `configure` flag to the helper and a test for it.

If you would rather not change the no-configuration default, the `config.md`
change alone also closes the issue and I am happy to drop that commit.

Fixes #1371.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,8 @@
 - Preserve deprecation status in workspace-symbol results. (#2044, @rgrinberg)
 - Render empty odoc tables in hover documentation without raising an internal
   error. (#2038, @rgrinberg)
+- Enable function parameter inlay hints when the client sends no configuration.
+  (#2013, fixes #1371, @dayangac)
 - Respect client support for signature-help parameter-label offsets. (#2009,
   @rgrinberg)
 - Keep Merlin diagnostic ranges within document bounds. (#2008, @rgrinberg)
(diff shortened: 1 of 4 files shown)


rgrinberg:
Merged in https://github.com/ocaml/ocaml-lsp/pull/2097. Thanks

PR #2032, opened 2026-07-31
test(definition): expose Not_found for nullary exception

rgrinberg:
Adds a regression test for go-to-definition on a nullary exception used as a structure expression:

```ocaml
exception E
E
```

A definition request on the second `E` currently leaks `Not_found` from `Merlin_analysis.Locate.find_definition_uid` as JSON-RPC `InternalError`. The test censors the unstable backtrace and otherwise snapshots the error unchanged.


rgrinberg:
Seems like a merlin issue @voodoos

PR #2033, opened 2026-07-31
test(rename): expose Not_found on incomplete local module

rgrinberg:
Adds a regression test for occurrence lookup on the incomplete local-module expression `let module X`.

`textDocument/prepareRename` at the valid EOF position currently leaks `Not_found` from `Merlin_analysis.Occurrences.uid_and_loc_of_node` as JSON-RPC `InternalError`. The test censors the unstable backtrace and preserves the exception and response code for a test-first fix.


rgrinberg:
Seems like a merlin issue @voodoos

PR #2046, opened 2026-07-31
test(rename): expose duplicate edits on incomplete binding

rgrinberg:
Renaming the incomplete binding `let rec ma` at its valid EOF position returns the same text edit twice. The raw Merlin `occurrences` query already returns the same location twice, and ocamllsp forwards both without deduplication. The primary root is in Merlin occurrence recovery, although the LSP adapter could defensively deduplicate before constructing a workspace edit.

This adds a static test-only regression and snapshots both duplicate edits.


rgrinberg:
cc @voodoos. There seems to be a merlin issue hiding in here.

PR #2088, opened 2026-08-02
fix(code-actions): hide inferred-intf action when nothing is missing

### LINKED ISSUE #347: "insert inferred interface" always shown in valid non-empty mli file
The code action "insert inferred interface" is always shown in a valid non-empty mli file. This isn't very helpful and is probably a bug. 

image

dayangac:
The inferred-interface code action was gated only on the document being an
interface and on the implementation opening successfully. It never checked
whether anything was left to insert, so on a `.mli` that already declares
everything the `.ml` exports the action still appeared in the menu and applied
an edit whose `newText` was empty.

Skip the action when the inferred signature is blank. Checking before the
ocamlformat round-trip also avoids an rpc call that cannot change the outcome.

Note this is a different code path from #2051, which made the
`ocamllsp/inferIntf` custom request reject interface documents. The code action
goes through `Inference.infer_intf` and was unaffected.

Existing tests covered an empty `.mli` and a partially populated one; the new
test covers a complete one.

Fixes #347.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@
 
 ## Fixes
 
+- Stop offering the inferred-interface code action when the interface already
+  declares everything the implementation exports. (#2088, fixes #347, @dayangac)
 - Unregister Dune promotion code actions when an RPC instance finishes.
   (#2076, @rgrinberg)
 - Avoid unregistering promotion code actions when clearing the last promotion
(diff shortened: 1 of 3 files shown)


rgrinberg:
Merged in https://github.com/ocaml/ocaml-lsp/pull/2096

In subsequent PR's, please make sure that every commit is passing (even tests that demonstrate undesirable behavior)

PR #2099, opened 2026-08-03
fix(folding-range): respect client folding range capabilities

### LINKED ISSUE #911: Respect Client Capabilities

dayangac:
Folding ranges were always built with character positions and a `Region` kind,
and the full list was always returned, without consulting the client's
`FoldingRangeClientCapabilities`. `foldingRangeProvider` is advertised
unconditionally, and `compute` already had `State.t` in scope.

Omit the character positions when the client sets `lineFoldingOnly`, omit the
kind when `Region` is absent from `foldingRangeKind.valueSet`, and truncate the
result to `rangeLimit`.

These fields are hints rather than hard requirements, so this is about not
sending data the client has said it will ignore, in line with the recent
capability-negotiation work.

`import.ml` gains one line: `Import` re-exports LSP types selectively and
`FoldingRangeKind` was not among them.

Each commit passes on its own: the test commit records the current output, and
the fix commit updates it. No other expectation in the suite changed.

Fixes #911.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,9 @@
 
 ## Fixes
 
+- Respect the client's folding range capabilities: omit character positions for
+  `lineFoldingOnly` clients, omit unsupported folding range kinds, and honor
+  `rangeLimit`. (#2099, fixes #911, @dayangac)
 - Return no location instead of failing the request when a definition,
   declaration or type definition cannot be found, so clients that request them
   eagerly no longer surface spurious errors. (#2100, fixes #1161, @dayangac)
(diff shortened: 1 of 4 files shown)


rgrinberg:
Do you mind putting the CHANGES entry in the same commit as the fix for next time? Makes it easier to revert things if there's a bug.

dayangac:
Of course, sorry for the trouble.

PR #2102, opened 2026-08-03
fix(hover): resolve constructors and exceptions at their declaration

### LINKED ISSUE #1300: Exceptions with record arguments should  not be functors.
Record-like arguments are important. OCaml 4.14, release 1.17.0 from OPAM

image

dayangac:
The hover iterator overrides `type_declaration` and `value_description` but
neither `constructor_declaration` nor `extension_constructor`, so hovering a
variant constructor or an exception where it is declared produced no response,
while the same name at a use site resolved.

Override both, so the declaration site reports the constructor type.

The issue title refers to the type being rendered as a functor. That rendering
is already gone; what remains at the same cursor position is no response at all,
which is what this fixes.

Fixes #1300.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,8 @@
   `rangeLimit`. (#2099, fixes #911, @dayangac)
 - Classify module names in `with` constraints and in signature `open`s in
   semantic highlighting. (#2098, fixes #808, @dayangac)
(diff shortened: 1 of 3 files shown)


rgrinberg:
Do you mind rebasing this?

dayangac:
rebased!

PR #2115, opened 2026-08-06
fix(code-actions): only offer to create a counterpart when the client can

### LINKED ISSUE #2114: Create <file>.mli code action emits a resource operation without checking client support
The "open related source" code action attaches a `CreateFile` workspace edit
when the counterpart file does not exist, without consulting
`workspace.workspaceEdit.documentChanges` or
`workspace.workspaceEdit.resourceOperations`.

`ocaml-lsp-server/src/code_actions/action_open_related.ml:56-62`:

```ocaml
let edit =
  match exists with
  | true -> None
  | false ->
    let documentChanges = [ `CreateFile (CreateFile.create ~uri ()) ] in
    Some (WorkspaceEdit.create ~documentChanges ())
in
```

A client advertising neither capability can only apply plain `TextEdit`s through
`WorkspaceEdit.changes`, so it must ignore this edit. The file is never created,
and the `ocamllsp/open-related-source` command on the same action then calls
`window/showDocument` on a path that does not exist, reaching the
`failed to open %s` branch at `action_open_related.ml:24-26`.

This is visible in the repository's own expectations.
`ocaml-lsp-server/test/e2e-new/code_actions_annotations.ml:55-64` runs with the
default `ClientCapabilities.create ()`, advertising no workspace capabilities,
and records:

```json
{ "command": { "arguments": [ "file:///foo.mli" ],
               "command": "ocamllsp/open-related-source",
               "title": "Create foo.mli" },
  "edit": { "documentChanges": [ { "kind": "create", "uri": "file:///foo.mli" } ] },
  "kind": "switch",
  "title": "Create foo.mli" }
```

No test advertises `resourceOperations`, and only `rename.ml` advertises
`workspaceEdit` at all, so this path is unexercised.

For contrast, `ocaml-lsp-server/src/rename.ml:147-173` consults
`workspace.workspaceEdit.documentChanges` and falls back to `changes`.

dayangac:
The open-related code action attaches a `CreateFile` workspace edit when the implementation or interface counterpart does not exist. Clients that do not advertise both `workspace.workspaceEdit.documentChanges` and the `create` resource operation cannot apply that edit; they then run the accompanying open command against a file that was never created.

Offer the action when the counterpart already exists, but offer creation only when the client supports both required capabilities. The capability coverage merged in #2119 now exercises supported and unsupported clients; this PR also covers clients advertising only one of the two requirements.

Fixes #2114.

### DIFF
--- a/ocaml-lsp-server/test/e2e-new/code_actions_annotations.ml
+++ b/ocaml-lsp-server/test/e2e-new/code_actions_annotations.ml
@@ -463,3 +463,66 @@ let my_fun x y : int = 1
   print_code_actions source range ~filter:find_remove_annotation_action;
   [%expect {| No code actions |}]
 ;;
+
+let capabilities_with_create_file =
+  let window =
+    let showDocument = ShowDocumentClientCapabilities.create ~support:true in
+    WindowClientCapabilities.create ~showDocument ()
+  in
+  let workspace =
+    let workspaceEdit =
+      WorkspaceEditClientCapabilities.create
+        ~documentChanges:true
+        ~resourceOperations:[ ResourceOperationKind.Create ]
+        ()
+    in
+    WorkspaceClientCapabilities.create ~workspaceEdit ()
+  in
+  ClientCapabilities.create ~window ~workspace ()
+;;
+
+let%expect_test "create counterpart action, client supports resource operations" =
+  let range = range ~start_line:0 ~start_character:4 ~end_line:0 ~end_character:5 in
+  print_code_actions
+    ~capabilities:capabilities_with_create_file
+    ~filter:(find_action "switch")
+    "let x = 1\n"
+    range;
+  [%expect
+    {|
+    Code actions:
+    {
+      "command": {
+        "arguments": [ "file:///foo.mli" ],
+        "command": "ocamllsp/open-related-source",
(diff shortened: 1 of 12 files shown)


ocaml-lsp-server/test/e2e-new/code_actions_annotations.ml
```diff
@@ -168,19 +157,7 @@ let iiii = 3 + 4
...
-        "title": "Create foo.mli"
-      },
-      "edit": {
-        "documentChanges": [ { "kind": "create", "uri": "file:///foo.mli" } ]
-      },
-      "kind": "switch",
-      "title": "Create foo.mli"
-    }
-     |}]
+    |}]
```
rgrinberg, line 160: Do we still have tests that demonstrate we are offering this command when appropriate?

dayangac:
No, and that's a gap. The code-action tests only advertise window.showDocument(Lsp_helpers.capabilities_with_show_document), so nothing exercises a client that can create files, the two expectations I changed were the only coverage, and both were the negative case.

rgrinberg:
Okay, let's fill that a gap with a test PR first. So that this PR demonstrates that we show CreateFile when appropriate and not show the code action when it is not.

dayangac:
will do.

PR #2133, opened 2026-08-07
fix(ocamlformat-rpc): degrade gracefully when the process dies at startup

rgrinberg:
An `ocamlformat-rpc` process that dies or produces garbage during startup
version negotiation used to raise an uncaught exception ("fd closed
unflushed") that failed the hover request with an internal error.

Wrap `Process.create` in `Fiber.collect_errors` and report the failure as an
error `LogMessage`, degrading to unformatted hover types instead of failing
the request. Fixes the reproduction in #2132.


rgrinberg:
Combined the regression test and fix in #2132.

